PR #67, opened 2020-10-14
Rewrite cstETH math (stETH provides current ratio)

lxzrv:


ongrid (changes requested):
Add tests with 'slashing' (decreasing totalControlledETH)
Add allowance tests to both dynamic branches

ongrid (approved):

PR #183, opened 2020-11-25
E2E: Change oracle invocation and vars. Live tag.

ongrid:
* Rename oracle.sh to oracle_build.sh (more informative name)
* Adapt .env and docker-compose.yml to oracle env names
* Use "e2e" tag for oracles instead of "latest" in docker-compose.yml
* Remove argvs from docker compose commands


e2e/.env
```diff
@@ -15,13 +15,13 @@ VALIDATOR_MNEMONIC2="web scheme virus economy adult word frozen erupt adjust beh
 OWNER="0xb4124cEB3451635DAcedd11767f004d8a28c6eE7"
 HOLDERS="0xb4124cEB3451635DAcedd11767f004d8a28c6eE7,0x8401Eb5ff34cc943f096A32EF3d5113FEbE8D4Eb,0x306469457266CBBe7c0505e8Aad358622235e768"
 
-LIDO_CONTRACT=0x
+POOL_CONTRACT=0xc12e8e7adcEaF31c1Ca5F8aFD99AB88439628183
```
krogla, line 18: no need to set value due to it automatically loaded from `deployed.json`

PR #189, opened 2020-11-25
Emit Approval() event on transferFrom() since allowance was changed

k06a:
See this discussion for both arguments to have and not have this event emission: https://github.com/makerdao/dss/issues/76


skozin:
Thank you for your contribution! This has been implemented in the process of merging `StETH.sol` and `Lido.sol` (#225), so I'm going to close this PR.

PR #190, opened 2020-11-25
SLOADs and requires optimizations

k06a:
- each SLOAD rn cost you 800 gas
- some require() statements were followed by SafeMath.sub() checking of the same underflow behavior


skozin:
Caching storage vars in memory vars is a good idea, was actually going to propose this improvement. But I don't think that removing the explicit preconditions is also a good idea — I'd prefer to get `BALANCE_INSUFFICIENT` revert reason instead of `MATH_SUB_UNDERFLOW`. That said, currently, the `require` calls lack any revert reasons. I would probably add the specific revert reasons instead of removing the calls—gas savings from removing these excess checks are negligible but readability and debuggability will suffer, imho.

k06a:
@skozin I agree on revert reasons, but revert reasons were not provided in double checks so I removed them. Also, check out this cool version of `SafeMath.sub()` which also accepts revert reason as an argument: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v3.3/contracts/math/SafeMath.sol#L60

ujenjt:
We've adopted this approach and remove unnecessary reads in the current version

PR #193, opened 2020-11-26
Fix return statements #156

### LINKED ISSUE #156: QuantStamp Audit: QSP-7 - Missing return values
[QuantStamp security assessment](https://github.com/lidofinance/lido-dao/files/5561204/QuantStampReport.pdf) of [v0.1.0-rc.1](https://github.com/lidofinance/lido-dao/releases/tag/v0.1.0-rc.1) raised the following security issue:

# Issue: QSP-7

| Short Descr. | Missing return values |
|-------------|--------|
| Severity | Informational |
| Files affected | `DePool.sol`, `StakingProvidersRegistry.sol` |

## Description

The following functions are declared to return values but have no explicit return statement:

1. StakingProvidersRegistry.addStakingProvider does not return uint256 id .

2. DePool.submit is declared to return uint256 StETH but has no return statement.

## Recommendations

Add return statements to these functions.

lxzrv:
Closes #156
- Lido.submit() now returns shares amount generated instead of token amount because this value is fixed and carries more information than dynamically changing balance. if we return anything at all, that would be better.

### DIFF
--- a/contracts/0.4.24/Lido.sol
+++ b/contracts/0.4.24/Lido.sol
@@ -123,9 +123,9 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
     /**
     * @notice Send funds to the pool with optional _referral parameter
     * @dev This function is alternative way to submit funds. Supports optional referral address.
-    * @return StETH Amount of StETH tokens generated
+    * @return Amount of StETH shares generated
     */
-    function submit(address _referral) external payable returns (uint256 StETH) {
+    function submit(address _referral) external payable returns (uint256) {
         return _submit(_referral);
     }
 
@@ -424,9 +424,9 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
     /**
     * @dev Process user deposit, mints liquid tokens and increase the pool buffer
     * @param _referral address of referral.
-    * @return StETH amount of tokens generated
+    * @return amount of StETH shares generated
     */
-    function _submit(address _referral) internal whenNotStopped returns (uint256 StETH) {
+    function _submit(address _referral) internal whenNotStopped returns (uint256) {
         address sender = msg.sender;
         uint256 deposit = msg.value;
         require(deposit != 0, "ZERO_DEPOSIT");
@@ -443,6 +443,8 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
         }
 
         _submitted(sender, deposit, _referral);
+
+        return sharesAmount;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


contracts/0.4.24/Lido.sol
```diff
@@ -123,9 +123,9 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
     /**
     * @notice Send funds to the pool with optional _referral parameter
     * @dev This function is alternative way to submit funds. Supports optional referral address.
-    * @return StETH Amount of StETH tokens generated
+    * @return Amount of StETH shares generated
```
ongrid, line 126: Explain why shares instead of tokens. We discussed this verbally, but this should be stated explicitly in commit message.

contracts/0.4.24/Lido.sol
```diff
@@ -123,9 +123,9 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
...
     */
-    function submit(address _referral) external payable returns (uint256 StETH) {
+    function submit(address _referral) external payable returns (uint256) {
```
ongrid, line 128: this behavior should be covered by test and returned amount should be asserted.

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -111,6 +111,8 @@ contract NodeOperatorsRegistry is INodeOperatorsRegistry, IsContract, AragonApp
         operator.stakingLimit = _stakingLimit;
 
         emit NodeOperatorAdded(id, _name, _rewardAddress, _stakingLimit);
+
+        return id;
```
ongrid, line 115: This is out of scope. Should be done in a separate issue and in separate commit.
I'd propose to drop this change.

ongrid (changes requested):
* The new behavior must be tested, numbers must be asserted.
* Changing returnable value should be justified in commit message.

contracts/0.4.24/Lido.sol
lxzrv: Done

contracts/0.4.24/Lido.sol
lxzrv: Discussed verbally, tests will be later.

ongrid: Reason: return value of transact is not visible in the returned object of promise. To test this we need to wrap into `.call()`able mock but we are under time pressure. We'll put this in backlog.

ongrid (approved):

krogla (approved):
looks good

PR #201, opened 2020-11-28
[WIP] Add Transfer event indicating mint on submit

vshvsh:
There's not Transfer event on submit indicating token minted. This leads to bad UX for stakers - e.g. they can't see their tx on etherscan etc.  Added it to `_submitted` function, though might not be the best place - token's `mintShare` might be better, but more expensive gas-wise.

@skozin @ongrid WDYT?

### DIFF
--- a/contracts/0.4.24/Lido.sol
+++ b/contracts/0.4.24/Lido.sol
@@ -447,7 +447,6 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
         } else {
             token.mintShares(sender, sharesAmount);
         }
-
         _submitted(sender, deposit, _referral);
 
         return sharesAmount;
@@ -636,7 +635,7 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
     */
     function _submitted(address _sender, uint256 _value, address _referral) internal {
         BUFFERED_ETHER_VALUE_POSITION.setStorageUint256(_getBufferedEther().add(_value));
-
+        emit Transfer(address(0), _sender, _value);
         emit Submitted(_sender, _value, _referral);
     }
 


skozin:
I think that a good solution would be to move all share-calculation logic from `StETH` to `Lido` (they're too coupled) and make `StETH` a mere non-Aragon ERC20 interface to `Lido`. But that would require significant changes.

As for `Transfer`s from `Lido`, it's clearly an improvement but I don't know how Etherscan would reflect `Transfers` emitted from a non-ERC20 contract different form the token being transferred.

skozin (commented):
I think that a good solution would be to move all share-related logic from `StETH` to `Lido` (they're too coupled) and make `StETH` a mere non-Aragon ERC20 interface to `Lido`. But that would require significant changes.

As for `Transfer`s from `Lido`, it's clearly an improvement but I don't know how Etherscan would reflect `Transfers` emitted from a non-ERC20 contract different from the token being transferred.

skozin:
Another solution which doesn't require spending a lot of gas: pass `totalSupply` to `mintStares` and emit `Transfer` from the latter.

vshvsh:
Non-Aragon erc20 interface for Aragon would still use a load of gas. Also, you're right, my fix doesn't work bc Lido contract doesn't know what Transfer is.

skozin:
> Non-Aragon erc20 interface for Aragon would still use a load of gas

When sending TXes to the token, yes. But minting would go without any overhead since it would be done without any calls, it would come down to simply modifying the Lido’s storage. And the main issue we have right now is the minting cost.

ongrid:
To see the token minted, the event should be emitted by the token contract, not Lido itself. I believe, this event will only confuse the user and won't be treated as StETH-related.

`Lido` doesn't have `Transfer` event declared to emit it. 

```
contracts/0.4.24/Lido.sol:638:14: DeclarationError: Undeclared identifier.
        emit Transfer(address(0), _sender, _value);
             ^------^
```

To make this technically possible we need either to declare it or inherit from ERC-20.

> I think that a good solution would be to move all share-related logic from StETH to Lido (they're too coupled) and make StETH a mere non-Aragon ERC20 interface to Lido. But that would require significant changes.

Good point, this will definitely will save gas on external calls and make the logic more straightforward, with less wrapping layers. But merging the code back to the monolith will need a lot of changes. The simplification of the code on one side will have the opposite effect on tests - they will become more complicated (will need separate mocks to expose internals for asserting).

PR #215, opened 2020-12-01
Unstructured storage

### LINKED ISSUE #181: Use unstructured storage for conventional state vars in upgradable contracts
The variables that currently don't use Unstructured Storage:

**Lido.sol:** `withdrawalCredentials` (will be added in #204: `shares`, `totalShares`).
**LidoOracle.sol:** `members`, `quorum`, `pool`, `beaconSpec`, `earliestReportableEpochId`, `lastReportedEpochId`, `gatheredEpochData`.
**NodeOperatorsRegistry.sol:** `operators`, `totalOperatorsCount`, `activeOperatorsCount`, `pool`.

lxzrv:
Closes #181

### DIFF
--- a/contracts/0.4.24/Lido.sol
+++ b/contracts/0.4.24/Lido.sol
@@ -534,7 +534,8 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
     * @param _signature Signature of the deposit call
     */
     function _stake(bytes memory _pubkey, bytes memory _signature) internal {
-        require(getWithdrawalCredentials() != 0, "EMPTY_WITHDRAWAL_CREDENTIALS");
+        bytes32 withdrawalCredentials = getWithdrawalCredentials();
+        require(withdrawalCredentials != 0, "EMPTY_WITHDRAWAL_CREDENTIALS");
 
         uint256 value = DEPOSIT_SIZE;
 
@@ -553,7 +554,7 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
 
         bytes32 depositDataRoot = sha256(
             abi.encodePacked(
-                sha256(abi.encodePacked(pubkeyRoot, getWithdrawalCredentials())),
+                sha256(abi.encodePacked(pubkeyRoot, withdrawalCredentials)),
                 sha256(abi.encodePacked(_toLittleEndian64(depositAmount), signatureRoot))
             )
         );
@@ -561,7 +562,7 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
         uint256 targetBalance = address(this).balance.sub(value);
 
         getValidatorRegistrationContract().deposit.value(value)(
-            _pubkey, abi.encodePacked(getWithdrawalCredentials()), _signature, depositDataRoot);
+            _pubkey, abi.encodePacked(withdrawalCredentials), _signature, depositDataRoot);
         require(address(this).balance == targetBalance, "EXPECTING_DEPOSIT_TO_HAPPEN");
     }
 
--- a/contracts/0.4.24/Lido.sol
+++ b/contracts/0.4.24/Lido.sol
@@ -72,7 +72,7 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
 
 
     /// @dev Credentials which allows the DAO to withdraw Ether on the 2.0 side
-    bytes private withdrawalCredentials;
+    bytes32 internal constant WITHDRAWAL_CREDENTIALS_POSITION = keccak256("lido.Lido.withdrawalCredentials");
 
 
     // Memory cache entry used in the _ETH2Deposit function
@@ -210,7 +210,7 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
     function setWithdrawalCredentials(bytes _withdrawalCredentials) external auth(MANAGE_WITHDRAWAL_KEY) {
         require(_withdrawalCredentials.length == WITHDRAWAL_CREDENTIALS_LENGTH, "INVALID_LENGTH");
 
-        withdrawalCredentials = _withdrawalCredentials;
+        WITHDRAWAL_CREDENTIALS_POSITION.setStorageBytes32(BytesLib.toBytes32(_withdrawalCredentials, 0));
         getOperators().trimUnusedKeys();
 
         emit WithdrawalCredentialsSet(_withdrawalCredentials);
@@ -240,7 +240,7 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
         require(_beaconValidators <= depositedValidators, "REPORTED_MORE_DEPOSITED");
 
         uint256 beaconValidators = BEACON_VALIDATORS_VALUE_POSITION.getStorageUint256();
-        // Since the calculation of funds in the ingress queue is based on the number of validators 
+        // Since the calculation of funds in the ingress queue is based on the number of validators
         // that are in a transient state (deposited but not seen on beacon yet), we can't decrease the previously
         // reported number (we'll be unable to figure out who is in the queue and count them).
         // See LIP-1 for details https://github.com/lidofinance/lido-improvement-proposals/blob/develop/LIPS/lip-1.md
@@ -307,13 +307,6 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
     }
 
     /**
-      * @notice Returns current credentials to withdraw ETH on ETH 2.0 side after the phase 2 is launched
-      */
-    function getWithdrawalCredentials() external view returns (bytes) {
-        return withdrawalCredentials;
-    }
-
-    /**
     * @notice Get the amount of Ether temporary buffered on this contract balance
     * @dev Buffered balance is kept on the contract from the moment the funds are received from user
     * until the moment they are actually sent to the official Deposit contract.
@@ -330,6 +323,13 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
     */
     function getTotalPooledEther() external view returns (uint256) {
         return _getTotalPooledEther();
+    }
+
+    /**
+      * @notice Returns current credentials to withdraw ETH on ETH 2.0 side after the phase 2 is launched
+      */
+    function getWithdrawalCredentials() public view returns (bytes) {
+        return abi.encodePacked(WITHDRAWAL_CREDENTIALS_POSITION.getStorageBytes32());
     }
 
     /**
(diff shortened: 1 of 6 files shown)


contracts/0.4.24/Lido.sol
```diff
@@ -332,6 +325,13 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
         return _getTotalPooledEther();
     }
 
+    /**
+      * @notice Returns current credentials to withdraw ETH on ETH 2.0 side after the phase 2 is launched
+      */
+    function getWithdrawalCredentials() public view returns (bytes) {
```
skozin, line 331: The same here for the return type. This way we can do without the `bytes <-> bytes32` conversion.

contracts/0.4.24/Lido.sol
```diff
@@ -210,7 +210,7 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
     function setWithdrawalCredentials(bytes _withdrawalCredentials) external auth(MANAGE_WITHDRAWAL_KEY) {
```
skozin, line 210: Since we're already using `bytes32` storage for withdrawal credentials, I'd just change the argument type to `bytes32`. We can upgrade the contract if the spec changes.

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -101,10 +101,16 @@ contract NodeOperatorsRegistry is INodeOperatorsRegistry, IsContract, AragonApp
         validAddress(_rewardAddress)
         returns (uint256 id)
     {
-        id = totalOperatorsCount++;
+        bytes32 totalOperatorsCountPosition = TOTAL_OPERATORS_COUNT_VALUE_POSITION;
```
skozin, line 104: No need to cache `TOTAL_OPERATORS_COUNT_VALUE_POSITION` here: since it's a constant, reading it doesn't produce and `SLOAD`s.

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -101,10 +101,16 @@ contract NodeOperatorsRegistry is INodeOperatorsRegistry, IsContract, AragonApp
...
+        id = totalOperatorsCountPosition.getStorageUint256();
+        totalOperatorsCountPosition.setStorageUint256(id.add(1));
+
         NodeOperator storage operator = operators[id];
 
-        activeOperatorsCount++;
+        bytes32 activeOperatorsCountPosition = ACTIVE_OPERATORS_COUNT_VALUE_POSITION;
```
skozin, line 110: The same here for `ACTIVE_OPERATORS_COUNT_VALUE_POSITION`.

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -191,7 +199,7 @@ contract NodeOperatorsRegistry is INodeOperatorsRegistry, IsContract, AragonApp
     function updateUsedKeys(uint256[] _ids, uint64[] _usedSigningKeys) external onlyPool {
         require(_ids.length == _usedSigningKeys.length, "BAD_LENGTH");
         for (uint256 i = 0; i < _ids.length; ++i) {
-            require(_ids[i] < totalOperatorsCount, "NODE_OPERATOR_NOT_FOUND");
+            require(_ids[i] < getNodeOperatorsCount(), "NODE_OPERATOR_NOT_FOUND");
```
skozin, line 202: I'd cache the result of `getNodeOperatorsCount()` here to avoid doing multiple `SLOAD`s per loop. Anyway, this function was removed in #226.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -317,14 +332,25 @@ contract LidoOracle is ILidoOracle, IsContract, AragonApp {
...
+            return (false, Report({beaconBalance: 0, beaconValidators: 0}));
 
         // unpack Report struct from uint256
         Report memory modeReport = uint256ToReport(mode);
 
+        return (true, modeReport);
+    }
+
+    /**
+     * @dev Pushed the current data point if quorum is reached
```
skozin, line 344: A mistype here, you've probably meant Pushe**s**.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -288,17 +303,17 @@ contract LidoOracle is ILidoOracle, IsContract, AragonApp {
...
+            EARLIEST_REPORTABLE_EPOCH_ID_VALUE_POSITION.getStorageUint256()
+        );
         return (earliestReportableEpochId, getCurrentEpochId());
     }
 
-    /**
-     * @dev Pushed the current data point if quorum is reached
-     */
-    function _tryPush(uint256 _epochId) internal {
+    function _isQuorum(uint256 _epochId) internal returns (bool, Report) {
```
skozin, line 312: This function should be marked as `view`.

skozin: Also, since it returns the report I'd rename it to smth like `_getQuorumReport`. Naming it `_isQuorum` misleadingly implies that it returns a bool.

skozin: One more thing: I'd explicitly specify the `memory` storage location for the report, i.e. `returns (bool, Report memory)`.

contracts/0.4.24/Lido.sol
lxzrv: OK, Done

contracts/0.4.24/Lido.sol
lxzrv: Done

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
lxzrv: Agree, fixed.

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
lxzrv: Done

contracts/0.4.24/oracle/LidoOracle.sol
lxzrv: Fixed

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
lxzrv: Understood, if fun was removed, no need to fix that. I just will leave it as is.

contracts/0.4.24/oracle/LidoOracle.sol
lxzrv: Fixed and added description

ongrid (changes requested):
1. 
Since earliest and last are good adjectives for the events that already happened, for potential events such naming is confusing. New naming:
```
minReportableEpochId
maxReportableEpochId
```
BTW
We should keep `lastReportedEpochId` since it has been reported

2. 
Since beaconSpec is static-length of 32 bytes
```
struct BeaconSpec {
        uint64 epochsPerFrame;
        uint64 slotsPerEpoch;
        uint64 secondsPerSlot;
        uint64 genesisTime;
    }
```
it's possible to store it with unstructured pattern

ongrid:
Sorry, some previous notes were not submitted to the PR and got cached in the browser.

Actual ones are:

1. 
Since earliest and last are good adjectives for the events that already happened, for potential events such naming is confusing. New naming:
```
minReportableEpochId
maxReportableEpochId
```
BTW
We should keep `lastReportedEpochId` since it has been reported

2. 
Since beaconSpec is static-length of 32 bytes
```
struct BeaconSpec {
        uint64 epochsPerFrame;
        uint64 slotsPerEpoch;
        uint64 secondsPerSlot;
        uint64 genesisTime;
    }
```
it's possible to store it with unstructured pattern

contracts/0.4.24/Lido.sol
```diff
@@ -72,7 +72,7 @@ contract Lido is ILido, IsContract, Pausable, AragonApp {
 
 
     /// @dev Credentials which allows the DAO to withdraw Ether on the 2.0 side
-    bytes private withdrawalCredentials;
```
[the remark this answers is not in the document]
skozin: Currently, withdrawal credentials are 32 bytes long, so we can store them in a bytes32 array. This can change, but then we can upgrade our contracts.

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -78,14 +78,14 @@ contract NodeOperatorsRegistry is INodeOperatorsRegistry, IsContract, AragonApp
...
-        require(_id < totalOperatorsCount, "NODE_OPERATOR_NOT_FOUND");
+        require(_id < getNodeOperatorsCount(), "NODE_OPERATOR_NOT_FOUND");
         _;
     }
 
     function initialize(address _pool) public onlyInit {
-        totalOperatorsCount = 0;
-        activeOperatorsCount = 0;
-        pool = _pool;
+        TOTAL_OPERATORS_COUNT_VALUE_POSITION.setStorageUint256(0);
```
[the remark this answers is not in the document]
skozin: I think it's unnecessary. It improves readability, though, so I'd leave it in place.

contracts/0.4.24/Lido.sol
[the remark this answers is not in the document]
ongrid: Sorry, this was an outdated comment from my cache

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -284,21 +299,26 @@ contract LidoOracle is ILidoOracle, IsContract, AragonApp {
     function getCurrentReportableEpochs()
         public view
         returns (
-            uint256 firstReportableEpochId,
-            uint256 lastReportableEpochId
+            uint256 minReportableEpochId,
```
skozin, line 302: This will require oracle daemon changes — need not to forget about this.

ongrid (approved):

skozin:
> 2. 
> Since beaconSpec is static-length of 32 bytes
> ```
> struct BeaconSpec {
>        uint64 epochsPerFrame;
>        uint64 slotsPerEpoch;
>        uint64 secondsPerSlot;
>        uint64 genesisTime;
>    }
> ```
> it's possible to store it with unstructured pattern

@ongrid good idea! We can indeed pack it into a single `uint256`. If you or @lxzrv have time to do that in the next 3-4 hours, let's do that.

skozin (approved):

skozin:
Otherwise, I think its's good to go.

contracts/0.4.24/oracle/LidoOracle.sol
ongrid: This getter especially is not used by oracle daemon. And other call results are returned as the pythonic list [] (with the default web3.py's ABI decoder) so the names don't really matter here, just position does. And it's unchanged.

ongrid (approved):

skozin (approved):

ongrid:
Agreed to assign @skozin for merging this

contracts/0.4.24/oracle/LidoOracle.sol
skozin: Ah, got it. `web3.js` returns named vars, too, if the return args are named, so I thought that maybe it's like this on the Python side, too.

PR #221, opened 2020-12-03
impose a limit on max signing keys per submission

azat-s:
This PR features an additional validation rule for the signing keys submission. It disallows users to submit more than 20 keys in one go. The limit can be specified as an environment variable in package.json, otherwise defaults to 20.


azat-s:
Closed because these changes are present in #222

PR #222, opened 2020-12-03
limit signing keys to 20 per submission & advanced validation

azat-s:
This PR features an additional validation rule for the signing keys submission. It disallows users to submit more than 20 keys in one go. The limit can be specified as an environment variable in package.json, otherwise defaults to 20.

Also, fixes for some cases in which addition of floats returns unexpected results.


azat-s:
@skozin 

Implemented advanced validation using an external api. Checks public keys for duplicates among already submitted ones, and also verifies signatures.
Added a checkbox and a warning so that the user can opt out of advanced validation.

*Quick demo*
image


*Loader*
image

PR #223, opened 2020-12-03
Add report bounds logic, #39

### LINKED ISSUE #39: Bound oracles' possible reports from both sides by a configurable relative change value
To make oracles less dangerous, we can bound rewards report by 0.1% increase in stake and 15% decrease in stake, with both values configurable by DAO voting in case of extremely unusual circumstances.

lxzrv:
Closes #39


vshvsh:
Upper reward bound should be per day, not per report (so if there weren't a report. for two weeks we should allow 14x more reward than usual)
Also, this probably really should be in the oracle, not in Lido.

lxzrv:
> 
> 
> Upper reward bound should be per day, not per report (so if there weren't a report. for two weeks we should allow 14x more reward than usual)
> Also, this probably really should be in the oracle, not in Lido.

Okay, my misunderstanding, I will fix that.

ongrid (changes requested):
Since the spec in-progress, this work is irrelevant. The entire branch can be discarded and PR to be closed without merging.

PR #225, opened 2020-12-04
Merge StETH and Lido into the one contract

ujenjt:
The changes are aimed to remove a lot of cross calls between the Lido and the StETH smart contracts and simplify the protocol's code.

The protocol code in the Lido contract is closely coupled with StETH token. Basic operations like deposits and reward distribution require multiple reads and writes between the Lido and the StETH contracts. This makes then costly, and doubly so because of the Aragon tax.

The pull request makes StETH a base contract for Lido, eliminating the need to the numerous cross calls. StETH is an abstract contract implementing ERC20 spec. It encapsulates the logic of share-based balances and methods like `_transferShares`, `_mintShares` and `_burnShares`. Note that the only virtual function is `_getTotalPooledEther`, and the Lido as a subclass contract implements it.

`burnShares` is moved to the Lido code and is wrapped with ACL role checker. Public `mintShares` method was deleted since only the Lido protocol is authorized to mint tokens. It does it through the internal `_mintShares` method.

While the current result is bigger than what we have started from, it's still bellow the maximum ethereum contract size (24576 bytes) and the total amount of the code we're deploying to the blockchain is 7644 bytes less.

Lido contract before merge: 18973 bytes
StETH contract before merge: 9993 bytes
Lido contract after merge: 21322 bytes

The updated `submit` method is almost twice cheaper than the current one. As every protocol user bears the gas costs of calling it, optimizing the method is quite important for the Lido.

First `submit` call from the address: current 158494 gas, pull version 86335 gas
Next `submit` calls from the address: current 146494 gas, pull version 74335 gas


ujenjt:
This pull is still WIP since tests related to the `burnShares` logic was commented out. Will update it soon.

skozin (approved):

krogla (approved):

PR #236, opened 2020-12-12
Fix reward distribution algorithm

### LINKED ISSUE #237: LID-01 Node Operator Rewards Unevenly Leaked
Node operators get part of the fees as reward. This reward is proportional to the number of validators the node operator has. The calculation happens in `NodeOperatorsRegistry.distributeRewards` method. Current ([`v0.2.1-rc.0`](https://github.com/lidofinance/lido-dao/tree/v0.2.1-rc.0)) implementation uses integer division for every node operator, and invariably results in some small remainder that stays with the NodeOperatorRegistry contract balance. As the `distributeRewards` method distributes only current reward amount and not the whole balance, those remainders accumulate.

The audit proposes to design a more sophisticated process for reward distribution. The points are:
* It’s better to distribute the whole NodeOperatorsRegistry token balance, preventing remainders from accumulating.
* Calculate the operators’ rewards using single division — calculate the amount for one validator and distribute fees accordingly.

### LINKED ISSUE #240: Values in events generated by `submit` and fallback function are incorrect
Due to the order of operations in `_submit` method in `Lido.sol` the `value` field in generated `Transfer` event is calculated incorrectly

### LINKED ISSUE #241: LID-10 Precision loss in node operators fee calculation
The node operator’s fee is specified in current Lido implementation (`v0.2.0-rc.0`) in a variable number of 1/10000 points. The implementation has multiple integer divisions involving the fee variable and `10000` number, which causes precision loss and makes for unnecessary truncation.
The audit suggests changing the implementation to avoid `div(10000)` operations until later. Namely, calculate the `sharesToMint` value as `(feeBasis * totalRewards * prevTotalShares) / ((10000 * newTotalPooledEther) * (feeBasis * totalRewards) )`

ujenjt:
This PR refactors node operators’ reward distribution mechanics. Previously, `NodeOperatorsRegistry` distributed token rewards between operators on its own. Now, `Lido` gets a list of reward amounts and the corresponding operator addresses from `NodeOperatorRegistry` and performs transfers by itself. In the case of registering a profit without any active node operators, the new algorithm transfers all operators' fee tokens to the treasury. This approach is more gas efficient and prevents accumulating `StETH` dust on any contract involved.

Fix #237 
Fix #241 
Fix #240

### DIFF
--- a/contracts/0.4.24/Lido.sol
+++ b/contracts/0.4.24/Lido.sol
@@ -609,11 +609,12 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
         _emitTransferAfterMintingShares(treasury, toTreasury);
     }
 
-    function _distributeNodeOperatorsReward(uint256 _sharesToDistribute) internal returns (uint256) {
+    function _distributeNodeOperatorsReward(uint256 _sharesToDistribute) internal returns (uint256 distributed) {
         (address[] memory recipients, uint256[] memory shares) = getOperators().getRewardsDistribution(_sharesToDistribute);
 
         assert(recipients.length == shares.length);
 
+        distributed = 0;
         for (uint256 idx = 0; idx < recipients.length; ++idx) {
             _transferShares(
                 address(this),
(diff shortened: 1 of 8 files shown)


kadmil:
Fixes #237 #240 #241

skozin (approved):

PR #238, opened 2020-12-13
Prevent accidental submission of Ether by calling non-existing functions

ujenjt:
Current `Lido` smart contract fallback function treats any call as funds submission. This may result in users accidentally locking their funds if they mistook the Lido contract for something else, or accessed it through an inaccurate interface. This PR fixes the issue by adding `require(msg.data.length == 0);` to the fallback function.


kadmil:
Fixes #244

PR #249, opened 2020-12-14
e2e-update

itaven:


ongrid:
I'd suggest to target this PR to the `develop` branch instead of `master`. @skozin ?

ongrid:
Oracles are broken since ABI has been changed. To be fixed in [lido-oracle#76](https://github.com/lidofinance/lido-oracle/issues/76) and [lido-oracle#75](https://github.com/lidofinance/lido-oracle/issues/75).

PR #253, opened 2020-12-14
Add additional protection for implementations from accidental Ether submission

### LINKED ISSUE #243: LID-07 Lido smart contract isn’t ready to receive deposits right after deployment
Current (`v0.2.0-rc.0`) implementation doesn't follow some of the Aragon app guides. Namely, it's possible to call state changing methods of `Lido` contract implementation directly on the implementation, and not the proxy wired as an Aragon app. This creates the possibility for malicious actors: if they guide users to, say, call `submit` directly on implementation, their funds would be locked forever. While the audit assesses the risk to be low, it's better to protect the implementation.

ujenjt:
Fixes #243. This changes the `Pausable` contract’s logic to be paused by default and switches it on in `Lido.initialize()` using the `_resume()` method. Also, this adds the `whenNotStopped` modifier to the  `pushBeacon` method so it isn't possible to report oracle data until the protocol is initialized.

### DIFF
--- a/contracts/0.4.24/Lido.sol
+++ b/contracts/0.4.24/Lido.sol
@@ -256,7 +256,7 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
     * @param _beaconValidators number of Lido's keys in the beacon state
     * @param _beaconBalance simmarized balance of Lido-controlled keys in wei
     */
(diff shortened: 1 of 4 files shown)


kadmil (approved):
LGTM

kadmil:
Fixing #243

Issue #250, opened 2020-12-14
pre-mainnet review: NodeOperatorsRegistry


Eenae:
- [x] warning https://github.com/lidofinance/lido-dao/blob/ad4b2f634850e46375bf5c8717abf0e49cdce3be/contracts/0.4.24/nos/NodeOperatorsRegistry.sol#L320

It's dangerous to put business logic (in this case - the increment) into assert statements since they control the self-consistency of the code and can be removed any time (e.g. in a release build).

- [x] comment https://github.com/lidofinance/lido-dao/blob/ad4b2f634850e46375bf5c8717abf0e49cdce3be/contracts/0.4.24/nos/NodeOperatorsRegistry.sol#L75-L81

The implied notation of the slot naming is `keccak256("project.contract.variable")`, so `Lido` should be replaced with `NodeOperatorsRegistry`. This kind of slip also appears in some other contracts.

ujenjt:
Fixed in #255 and #232

PR #255, opened 2020-12-14
No side effects in an assertion

skozin:
Avoid modifying state in an assertion since the latter may be removed by automated tools.

Highlighted in: #250.


ujenjt (approved):

ujenjt:
LGTM

Issue #251, opened 2020-12-14
pre-mainnet review: Lido


Eenae:
- [x] warning https://github.com/lidofinance/lido-dao/blob/ad4b2f634850e46375bf5c8717abf0e49cdce3be/contracts/0.4.24/Lido.sol#L447

The emitted event will have a zero value since `getPooledEthByShares(deposit)` is still zero at this moment.

- [ ] warning https://github.com/lidofinance/lido-dao/blob/ad4b2f634850e46375bf5c8717abf0e49cdce3be/contracts/0.4.24/Lido.sol#L442-L443

Precision loss during `sharesAmount` computation may be mistaken as the zero pooled ether situation.
Consider an exagerated situation when 1 unit of share equals 1e9 units of ether. Sending a 0.999 gwei deposit will result in `sharesAmount == 0` yielding `true`, `_mintShares(sender, 0.999 gwei)` will be executed. This will result in net profit `0.999e9 * 1e9 - 1e9 - tx_fee` which is much more than `1e9 + tx_fee` ether paid.

I suggest replacing the `sharesAmount` check with an explicit one checking `0 == _getTotalPooledEther()`.

- [ ] comment https://github.com/lidofinance/lido-dao/blob/ad4b2f634850e46375bf5c8717abf0e49cdce3be/contracts/0.4.24/Lido.sol#L423
https://github.com/lidofinance/lido-dao/blob/ad4b2f634850e46375bf5c8717abf0e49cdce3be/contracts/0.4.24/Lido.sol#L428

I suggest adding `isContract` checks.

ujenjt:
1) is fixed in #236 pull request, also this issue is mentioned in #240.
2) this is good to have, but not a critical one, the fact that that addresses is contracts should be validate off-chain while DAO voting.

PR #266, opened 2020-12-17
Fix possible precision loss issue while shares calculation

ujenjt:
This PR is fixing the issue described in #251 related to possible precision loss while sharesAmount computation in the _submit method. Indeed, for some extreme shares/ether ratios (e.g 1 wei of shares corresponds to 1e9 wei of ether) it's possible to submit a little deposit that is less than ether per share ratio (in our example it is 1e9 - 1 wei) and gets an unreasonably large number of tokens in exchange. The described situation is possible only if the protocol will profit during a significant amount of time (like many years) or in case of significant token burning.

As @Eenae suggests, this PR replace condition `(sharesAmount == 0)` to `(_getTotalPooledEther() != 0)` which prevents such situation. The best from the correctness perspective is to check both `(_getTotalPooledEther() != 0 && totalShares != 0)`, but this is not necessary because complete zeroing of the total controlled ether is impossible right now.

While the pull request removes the inconsistency, this change increases the amount of gas required per submission, which we want to keep as low as possible.

### DIFF
--- a/contracts/0.4.24/Lido.sol
+++ b/contracts/0.4.24/Lido.sol
@@ -442,10 +442,12 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
         uint256 deposit = msg.value;
         require(deposit != 0, "ZERO_DEPOSIT");
 
-        uint256 sharesAmount = getSharesByPooledEth(deposit);
-        if (sharesAmount == 0) {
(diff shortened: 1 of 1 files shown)


Eenae (approved):

TheDZhon:
Superseded by #482

PR #277, opened 2020-12-24
Fix reporting of unimodal garbage data

lxzrv:
Introduces alternative quorum calculation mechanics to oracle contract.

### Old mechanics
1. Oracles report their data;
2. when gathered data values quantity is equal to `quorum`, oracle tries to find mode of these values;
3. when gathered data is unimodal, oracle pushes mode value to the Lido contract.

### New mechanics
1. Oracles report their data;
2. when gathered data values quantity is equal to `quorum`, oracle checks that we have at least as much identical values as `quorum` value is;
3. if that check passes, oracle pushes this most-frequent value to the Lido contract.

### For what this change?
Answer is: to prevent garbage data pushing to the Lido. We shouldn't pass mode value if all data is non-valid (it's unlikely, of course, but it may be). In the new solution we we accept the data that was first received a certain number of times (quorum). I think this approach is more correct conceptually and safe.


ongrid (changes requested):
* Describe the problem it solves in the commit message(s) or create the issue for the bug. Now it's unclear.
* Avoid `for` loop in test. Just random garbage "1234", "2345" and so on. To be trivial and readable.
* Add user story what is going on like "Within `members > 4` and `quorum > 2` configuration oracles gone crazy by any reason and started reporting garbage data. But two oracles reported the same value so two conditions get fulfilled as a result - array become unimodal and quorum got reached. So the garbage report gets pushed to the Lido."
* Add trivial failing test that gets fixed with the updated implementation.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -376,14 +375,108 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+
+        return (true, dataValues[mostFrequentValueIndex]);
+    }
+
+    /**
+     * @dev Returns quorum-frequent value when it exists in `data`
+     * @return isQuorum - true, when value frequency equals to quorum
+     * @return quorumValue - quorum-frequent value
+     */
+    function _findQuorumValue(uint256 quorum, uint256[] data) internal pure returns (bool isQuorum, uint256 quorumValue) {
```
ongrid, line 435: _findQuorumValue is non-trivial and need to be unit-tested. Expose it using mock and cover separately against different inputs.

ongrid (changes requested):

test/0.4.24/lidooracle.test.js
```diff
@@ -423,4 +443,45 @@ contract('LidoOracle', ([appManager, voting, user1, user2, user3, user4, nobody]
...
+      for (i = 0; i < 8; i++) {
+        await app.reportBeacon(0, i, i, { from: oracles[i] })
+      }
+
+      await assertReportableEpochs(0, 0)
+
+      // But the 8th and the 9th oracles randomly report same values. We should not use
+      // that random matched value as quorum!
+      await app.reportBeacon(0, 7, 7, { from: oracles[8] })
+      await assertReportableEpochs(0, 0)
```
ongrid, line 475: I'd suggest to use more direct approach for better readability - assert 'Completed' not in events.
+ keep assertReportableEpochs

test/0.4.24/lidooracle.test.js
```diff
@@ -423,4 +443,45 @@ contract('LidoOracle', ([appManager, voting, user1, user2, user3, user4, nobody]
...
+
+      // The 10th oracle reports same value as the 8th and the 9th.
+      // But we should accept value that was received certain number of times,
+      // because obviously probabilty of this case much less than probability
+      // of 2 random matches. Such quorum mechanics more safe than calculating
+      // mode when certain number of reports recieved.
```
ongrid, line 481: Too overcomplicated. The explanation should be simpler:

Previous implementation:
if N members submitted some random garbage and the array is unimodal (contains duplicate reports), it gets pushed (once N >= quorum).

New implementation:
Report get pushed to lido only if N+ members submitted strictly same numbers.

Test checks that previous behaviour got fixed (unimodal data doesn'r get submitted until the quantity of equal reports equals quorum)

ongrid (changes requested):

ongrid (changes requested):
Looks good. But solidity coverage got silently broken on CI

Screenshot 2020-12-29 at 15 18 09

ongrid:
It's broken for 2 weeks already. Raised issue #279

ongrid (approved):

dechjo:
This merge request is superseded with https://github.com/lidofinance/lido-dao/pull/282.

vshvsh:
Replaced by #282

PR #282, opened 2021-01-23
Update oracle quorum mode

dechjo:
TODO:
1. ✔️ For data [1,1,2,2,3,3] and quorum=2, we return none.
2. ✔️ assert vs require: Do we need `assert`s in contracts?
3. ✔️ emit `QuorumFailed` event if no quorum found, test it - not feasible here,
4. ✔️ fix "member removal removes their data" test.
5. ✔️ automated stress test for the algorithm.
6. Compare gas usage with old implementation?..


vshvsh:
1. yes
2. @skozin

contracts/0.4.24/oracle/Algorithm.sol
```diff
@@ -11,55 +11,62 @@ library Algorithm {
...
     /**
-      * Computes mode of a non-empty array, if array is unimodal.
-      * Low gas cost.
+      * @dev Finds the first element in the given array |data| that occures
+      * more that |quorum| times. Low gas cost for majority case. Returns the
+      * discovered element or 0 if no element is frequent enough.
       */
-    function mode(uint256[] data) internal pure returns (bool, uint256) {
-
+    function frequent(uint256[] data, uint256 quorum) internal pure returns (uint256) {
```
vshvsh, line 18: Maybe return (bool, uint256)? And return (false, 0) if there's no quorum. In practice 0 is unused and there will be no overlap, but I think that clarity outweighs minor gas economy in this case.

dechjo: Yes, I thought the same indeed, epochId cannot be 0. But OK, I'll switch to (bool, val) pair.

contracts/0.4.24/oracle/Algorithm.sol
```diff
@@ -11,55 +11,62 @@ library Algorithm {
...
+        if (quorum * 2 > data.length) {
+            
+            // If the expected quorum is more then half of data length, apply
+            // Boyer–Moore majority vote algorithm
+            for (i = 0; i < data.length; ++i) {
+                if (ctr == 0) {
+                    acc = data[i];
+                    ++ctr;
+                } else if (acc == data[i]) {
+                    if (++ctr == quorum) return acc;
```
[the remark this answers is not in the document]
dechjo: Added. Please tell me if you think of a better phrase

contracts/0.4.24/oracle/Algorithm.sol
vshvsh: oracles don't report epochId though, they report balances

contracts/0.4.24/oracle/Algorithm.sol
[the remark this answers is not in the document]
vshvsh: doesn't really matter, just control flow break needs to stand out

contracts/0.4.24/oracle/Algorithm.sol
```diff
@@ -11,55 +11,81 @@ library Algorithm {
...
-        dataValuesCounts[0] = 1;
-        uint256 dataValuesLength = 1;
+            // And make sure the resulted element is frequent enough
+            ctr = 0;
+            for (i = 0; i < data.length; ++i) {
+                if (data[i] == acc && ++ctr == quorum) return (true, acc);
+            }
+        } else {
+            
+            // Otherwise, apply optimized insertion sort
```
Eenae, line 47: Btw, there was a quick sort impl here https://github.com/lidofinance/lido-dao/blob/8b1203b248b638275f441fc16de0f0e2e1893c55/apps/depooloracle/contracts/Algorithm.sol#L25-L65

dechjo: Cool! Maybe It would be good to put our algorithms in a separate library (like this one)?..
In this case, I chose insertion sort because `data.length` should be low. But as we discussed with @vshvsh, this branch (line 45 and onwards) should work extremely rarely so the gas price is not important here at all.

Eenae: I agree. It's completely up to you.

Eenae:
1. For data [1,1,2,2,3,3,3] and quorum=2, should we still return 1?
2. Asserts are used for the same purpose as in java or c/c++. Although, solidity compiler can't omit them in a release build.

contracts/0.4.24/oracle/Algorithm.sol
dechjo: Ok, will have in mind for now, thank you for pointing this out!

test/0.4.24/algorithm.generator.py
```diff
@@ -0,0 +1,51 @@
...
+    template = inp.read()
+
+cases = ''
+for arr_len in range(2, 10):
+    positive = 0
+    for i in range(NUM_TESTS):
+        nums = []
+        for j in range(arr_len):
+            nums.append(random.randint(1, MAX_NUM))
+        quorum = random.randint(1, arr_len)
```
Eenae, line 38: I suggest making the test reproducible by seeding the RNG and logging the seed.

dechjo: Good point, implemented. Although this is mostly a manual test (I mean, not automatically run in any CI/CD) at least now.

Eenae (approved):

vshvsh:
> * For data [1,1,2,2,3,3,3] and quorum=2, should we still return 1?

It's not very important in the desired use case, so we can adopt the spec where 1, 2 and 3 are all valid returns. I think the spec where 3 is the only valid return is slightly better but it doesn't matter very much.

dechjo:
> > * For data [1,1,2,2,3,3,3] and quorum=2, should we still return 1?
> 
> It's not very important in the desired use case, so we can adopt the spec where 1, 2 and 3 are all valid returns. I think the spec where 3 is the only valid return is slightly better but it doesn't matter very much.

No need, I already fixed that: [1,1,2,2,3,3,3] will return 3 with quorum=2. Tests cover such case as well.

dechjo:
Replaced by https://github.com/lidofinance/lido-dao/pull/296.

PR #292, opened 2021-02-05
Oracle v2

dechjo:
The changes are described here: https://github.com/lidofinance/lido-improvement-proposals/blob/lip-002/LIPS/lip-2.md.


contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -198,26 +214,65 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
-
-        gatheredEpochData[_epochId].reportsBitMask = bitMask.setBit(index, true);
-
-        Report memory currentReport = Report(_beaconBalance, _beaconValidators);
-        gatheredEpochData[_epochId].reports[index] = currentReport;
+        // Push this report to the matching kind
+        uint256 i = 0;
+        Report memory report = Report(_beaconBalance, _beaconValidators);
+        uint256 reportRaw = reportToUint256(report);
+        while (i < gatheredReportKinds.length && reportToUint256(gatheredReportKinds[i].report) != reportRaw) ++i;
```
vsmelov, line 237: What is the max length of gatheredReportKinds? Could we go out-of-gas here?

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -198,26 +214,65 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+        if (_epochId > currentEpoch) _clearVotingAndAdvanceTo(_epochId);
 
-        // check & set contribution flag
-        uint256 bitMask = gatheredEpochData[_epochId].reportsBitMask;
+        // Make sure the oracle is from members list and has not yet voted
+        uint256 index = _getMemberId(msg.sender);
+        require(index != MEMBER_NOT_FOUND, "MEMBER_NOT_FOUND");
+        uint256 bitMask = REPORTS_BITMASK_POSITION.getStorageUint256();
         require(!bitMask.getBit(index), "ALREADY_SUBMITTED");
+        REPORTS_BITMASK_POSITION.setStorageUint256(bitMask.setBit(index, true));
```
vsmelov, line 231: Is it possible that we'll not have enough bits in uint256?

vshvsh: no, max number of oracles is 
```
 uint256 public constant MAX_MEMBERS = 256;
```

contracts/0.4.24/oracle/LidoOracle.sol
vshvsh: At most number of oracles, 256, but a good catch.

vshvsh:
```suggestion
        //gatheredReportKinds.length is at most MAX_MEMBERS (256) in length, so we'll not run out of gas
        while (i < gatheredReportKinds.length && reportToUint256(gatheredReportKinds[i].report) != reportRaw) ++i;
```

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -380,60 +452,73 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+        PREV_COMPLETED_EPOCH_ID_POSITION.setStorageUint256(LAST_COMPLETED_EPOCH_ID_POSITION.getStorageUint256());
+        LAST_COMPLETED_EPOCH_ID_POSITION.setStorageUint256(epochId);
+        reportSanityChecks();  // rollback on boundaries violation (logical consistency)
+
+        // call the quorum delegate
+        IQuorumCallback quorumCallbackAddr = IQuorumCallback(QUORUM_CALLBACK_POSITION.getStorageUint256());
+        if (address(quorumCallbackAddr) != address(0)) {
+            (uint256 lastTotalPooledEther,
+             uint256 prevTotalPooledEther,
+             uint256 timeElapsed) = getLastCompletedReports();
```
dechjo, line 510: Looks we are abusing public `getLastCompletedReports`: calling it here and inside `reportSanityChecks` just above.  Need to optimise it.

dechjo: Fix in https://github.com/lidofinance/lido-dao/pull/306...

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -73,153 +76,211 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
-
         uint256 index = _getMemberId(_member);
         require(index != MEMBER_NOT_FOUND, "MEMBER_NOT_FOUND");
-
-        uint256 lastReportedEpochId = (
-            LAST_REPORTED_EPOCH_ID_POSITION.getStorageUint256()
-        );
-
-        MIN_REPORTABLE_EPOCH_ID_POSITION.setStorageUint256(lastReportedEpochId);
         uint256 last = members.length.sub(1);
```
ujenjt, line 178: Is it required to use SafeMath here?

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -73,153 +76,211 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
-        uint256 bitMask = gatheredEpochData[lastReportedEpochId].reportsBitMask;
-        if (index != last) {
-            members[index] = members[last];
-            bitMask = bitMask.setBit(index, bitMask.getBit(last));
-        }
-        bitMask = bitMask.setBit(last, false);
-        gatheredEpochData[lastReportedEpochId].reportsBitMask = bitMask;
-
+        if (index != last) members[index] = members[last];
         members.length--;
```
ujenjt, line 180: And not using SafeMath here? I suggest to stay consistent here.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -73,153 +76,211 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+     */
+    function getQuorumCallback() public view returns(address) {
+        return address(QUORUM_CALLBACK_POSITION.getStorageUint256());
+    }
 
-        _assertInvariants();
+    /**
+     * @notice Set the callback contract address to be called upon quorum
+     */
+    function setQuorumCallback(address _addr) external auth(SET_QUORUM_CALLBACK) {
```
ujenjt, line 208: Better to check that address is not 0x0

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -73,153 +76,211 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+     * @param _epochId Beacon Chain epoch id
      * @param _beaconBalance Balance in wei on the ETH 2.0 side
      * @param _beaconValidators Number of validators visible on this epoch
      */
     function reportBeacon(uint256 _epochId, uint128 _beaconBalance, uint128 _beaconValidators) external {
-        (uint256 startEpoch, uint256 endEpoch) = getCurrentReportableEpochs();
-        require(_epochId >= startEpoch, "EPOCH_IS_TOO_OLD");
+        BeaconSpec memory beaconSpec = _getBeaconSpec();
+        (uint256 currentEpoch, uint256 endEpoch) = getCurrentReportableEpochs();
+        require(_epochId >= currentEpoch, "EPOCH_IS_TOO_OLD");
```
ujenjt, line 222: I propose to prohibit accepting any epochs other than the current frame. Why are we allowing the range here?

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -73,153 +76,211 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+
+        // so iterate on all report kinds we alredy have, not more then oracle members maximum
+        while (i < gatheredReportKinds.length && reportToUint256(gatheredReportKinds[i].report) != reportRaw) ++i;
+        if (i < gatheredReportKinds.length) {
+            ++gatheredReportKinds[i].count;
+        } else {
+            gatheredReportKinds.push(ReportKind(report, 1));
+        }
+        emit BeaconReported(_epochId, _beaconBalance, _beaconValidators, msg.sender);
         _tryPush(_epochId);
```
ujenjt, line 249: Every time the oracle reports the value, the `_tryPush` method is called. I suggest saving gas and not calling the method if there are fewer reports than quorum in the current epoch. What do you think?

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -73,153 +76,211 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
-
-        Report memory currentReport = Report(_beaconBalance, _beaconValidators);
-        gatheredEpochData[_epochId].reports[index] = currentReport;
-
+        // Push this report to the matching kind
+        uint256 i = 0;
+        Report memory report = Report(_beaconBalance, _beaconValidators);
+        uint256 reportRaw = reportToUint256(report);
+
+        // so iterate on all report kinds we alredy have, not more then oracle members maximum
```
ujenjt, line 241: If the current report results in a quorum, then the contract does not have to store the passed value in the contract storage. This improvement could save us minimum 5k gas for this tx.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -73,153 +76,211 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+        // If reported epoch has advanced, clear the last unsuccessful reporting
+        if (_epochId > currentEpoch) _clearReportingAndAdvanceTo(_epochId);
 
-        // check & set contribution flag
-        uint256 bitMask = gatheredEpochData[_epochId].reportsBitMask;
+        // Make sure the oracle is from members list and has not yet voted
+        uint256 index = _getMemberId(msg.sender);
+        require(index != MEMBER_NOT_FOUND, "MEMBER_NOT_FOUND");
+        uint256 bitMask = REPORTS_BITMASK_POSITION.getStorageUint256();
         require(!bitMask.getBit(index), "ALREADY_SUBMITTED");
```
ujenjt, line 233: It's better to merge `getBit` and `setBit` to one `testAndSet` method

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -380,60 +459,76 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
-        uint256 membersLength = members.length;
-        for (uint256 index = 0; index < membersLength; ++index) {
-            if (mask.getBit(index)) {
-                data[i++] = reportToUint256(gatheredEpochData[_epochId].reports[index]);
+        // if more than 2 kind of reports exist, choose the most frequent
+        uint256 maxi = 0;
+        for (uint256 i = 1; i < gatheredReportKinds.length; ++i) {
+            if (gatheredReportKinds[i].count > gatheredReportKinds[maxi].count) maxi = i;
+        }
+        for (i = 0; i < gatheredReportKinds.length; ++i) {
```
ujenjt, line 477: I think we can improve this algorithm and merge that two loops in one. What do you think?

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -380,60 +459,76 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
-        if (!isQuorum)
-            return;
+    function _tryPush(uint256 epochId) internal {
+        (bool isQuorum, Report memory modeReport) = _getQuorumReport();
+        if (!isQuorum) return;
 
         // data for this frame is collected, now this frame is completed, so
-        // minReportableEpochId should be changed to first epoch from next frame
+        // reportableEpochId should be changed to first epoch from the next frame
         BeaconSpec memory beaconSpec = _getBeaconSpec();
```
ujenjt, line 494: It's already called _getBeaconSpec in the beginning of reportBeacon. Maybe we can reorganize code to cache this?

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -380,60 +459,76 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+        // report to the Lido and collect stats
         ILido lido = getLido();
-        if (address(0) != address(lido))
-            lido.pushBeacon(modeReport.beaconValidators, modeReport.beaconBalance);
+        uint256 prevTotalPooledEther = ISTETH(lido).totalSupply();
+        lido.pushBeacon(modeReport.beaconValidators, modeReport.beaconBalance);
+        uint256 postTotalPooledEther = ISTETH(lido).totalSupply();
+
+        PRE_COMPLETED_TOTAL_POOLED_ETHER_POSITION.setStorageUint256(prevTotalPooledEther);
+        POST_COMPLETED_TOTAL_POOLED_ETHER_POSITION.setStorageUint256(postTotalPooledEther);
```
ujenjt, line 509: Do we need to store this value in Oracle smart contract? Because we've already have this value in Lido. What is the motivation behind this?

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -380,60 +459,76 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
-
-        assert(i == data.length);
-
-        // find mode value of this array
-        (bool isUnimodal, uint256 mode) = Algorithm.mode(data);
-        if (!isUnimodal)
-            return (false, Report({beaconBalance: 0, beaconValidators: 0}));
-
-        // unpack Report struct from uint256
-        modeReport = uint256ToReport(mode);
```
ujenjt, line 409: Please delete unused method `uint256ToReport `

test/0.4.24/lidooracle.test.js
```diff
@@ -84,11 +76,12 @@ contract('LidoOracle', ([appManager, voting, user1, user2, user3, user4, nobody]
 
   describe('Test utility functions:', function () {
     it('addOracleMember works', async () => {
+      await app.setTime(1606824000)
```
ujenjt, line 79: Why 1606824000?

test/0.4.24/lidooracle.test.js
```diff
@@ -114,11 +109,30 @@ contract('LidoOracle', ([appManager, voting, user1, user2, user3, user4, nobody]
       await app.removeOracleMember(user2, { from: voting })
 
       await assertRevert(app.removeOracleMember(user2, { from: user1 }), 'APP_AUTH_FAILED')
-      await assertRevert(app.removeOracleMember(user3, { from: voting }), 'QUORUM_WONT_BE_MADE')
 
       assert.deepStrictEqual(await app.getOracleMembers(), [user3])
     })
 
+    it('removeOracleMember updates minReportableEpochId', async () => {
```
ujenjt, line 116: It's critical to check that removeOracleMember clears all previous votes. Did't find this test.

test/0.4.24/lidooracle.test.js
```diff
@@ -338,86 +648,96 @@ contract('LidoOracle', ([appManager, voting, user1, user2, user3, user4, nobody]
...
+            await app.reportBeacon(epoch, START_BALANCE + 32, 1, { from: user1 })
+            await app.reportBeacon(epoch, START_BALANCE + 64, 2, { from: user2 })
+          }
+          await assertReportableEpochs(5, 5)
 
           await app.removeOracleMember(user1, { from: voting })
           await assertReportableEpochs(5, 5)
 
-          await app.reportBeacon(5, 64, 2, { from: user3 })
+          await app.reportBeacon(5, START_BALANCE + 33, 2, { from: user2 }) // user2 reports again
```
ujenjt, line 706: Worth to add explicit check that data has been deleted

test/0.4.24/lidooracle.test.js
```diff
@@ -338,86 +648,96 @@ contract('LidoOracle', ([appManager, voting, user1, user2, user3, user4, nobody]
...
+          await app.reportBeacon(5, START_BALANCE + 32, 1, { from: user1 }) // user1 reports again
+          await app.reportBeacon(5, START_BALANCE + 32, 1, { from: user2 })
           await assertReportableEpochs(5, 5)
 
-          await app.reportBeacon(5, 32, 1, { from: user3 })
+          await app.reportBeacon(5, START_BALANCE + 32, 1, { from: user3 })
           await assertReportableEpochs(6, 5)
         })
 
         it('quorum change triggers finalization of last reported epoch, all other reportable epochs will be staled', async () => {
```
ujenjt, line 732: Do we have a test for split voting with following configuration?

3 oracles vote for A
3 another oracles vote for B 
quorum is 3
and epoch shoudn't be finalized in that case

ujenjt (changes requested):

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -9,27 +9,28 @@ import "@aragon/os/contracts/apps/AragonApp.sol";
 import "@aragon/os/contracts/lib/math/SafeMath.sol";
 import "@aragon/os/contracts/lib/math/SafeMath64.sol";
 
-import "../interfaces/ILidoOracle.sol";
 import "../interfaces/ILido.sol";
+import "../interfaces/ILidoOracle.sol";
+import "../interfaces/IQuorumCallback.sol";
+import "../interfaces/ISTETH.sol";
```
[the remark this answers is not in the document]
dechjo: Looks the idea was to separate token-line code (ISTETH.sol) from the Lido.
Anyway, can we edit it now, when the Lido is already deployed?..

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: Yes, but this is the snapshot of the Lido's `totalPooledEther` for the time of `lido.pushBeacon()` tx. In any other time it may change (e.g. because of stETH purchase).

contracts/0.4.24/test_helpers/LidoMockForOracle.sol
```diff
@@ -0,0 +1,25 @@
...
+  * @dev Only for testing purposes! Lido version with some functions exposed.
+  */
+contract LidoMockForOracle {
+    uint256 private totalPooledEther;
+
+    function totalSupply() external view returns (uint256) {
+        return totalPooledEther;
+    }
+
+    function pushBeacon(uint256 /*_beaconValidators*/, uint256 _beaconBalance) external {
```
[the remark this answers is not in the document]
dechjo: because it is unused in this mock implementation. If uncommented, solc will show a warning.

test/0.4.24/bitops.test.js
```diff
@@ -0,0 +1,33 @@
+const { assert } = require('chai')
+const { assertBn } = require('@aragon/contract-helpers-test/src/asserts')
+const { bn } = require('@aragon/contract-helpers-test')
+
+const BitOps = artifacts.require('BitOpsMock.sol')
+
+contract('BitOps', ([testUser]) => {
```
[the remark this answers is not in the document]
dechjo: Agree.

contracts/0.4.24/test_helpers/LidoMockForOracle.sol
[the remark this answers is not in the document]
ujenjt: make sence, thanks

contracts/0.4.24/oracle/LidoOracle.sol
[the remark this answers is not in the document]
ujenjt: Sure, it's good idea to have one interface `ILido` with all required methods there

test/0.4.24/lidooracle.test.js
vshvsh: won't it finalize at the third vote for A?

contracts/0.4.24/oracle/LidoOracle.sol
vshvsh: I don't think we need to use safemath at all (if length is 0, `require(MEMBER_NOT_FOUND == _getMemberId(_member), "MEMBER_EXISTS");` will fail) but I must note that using safemath for the first check that substraction is possible and then using regular math where we know it's safe is a consistent pattern.

contracts/0.4.24/oracle/LidoOracle.sol
vshvsh: 0x0 is "unset". A valid value. Should be commented though.

contracts/0.4.24/oracle/LidoOracle.sol
vshvsh: There's the downside that it won't be queryable for monitoring purposes though

contracts/0.4.24/oracle/LidoOracle.sol
vshvsh: how do you you suppose to check that? use popcnt?

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: Fix in https://github.com/lidofinance/lido-dao/pull/311/files for clarity.

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: Fix in https://github.com/lidofinance/lido-dao/pull/311.

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: Fix in https://github.com/lidofinance/lido-dao/pull/311/files for clarity.
The last report would not have been seen in monitoring anyway, because it would be cleared immediately after the push to Lido. Good catch. Fixed that first.

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: Fix in https://github.com/lidofinance/lido-dao/pull/311. After the fix, no need to `_tryPush` every time.

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: Agree. Fix in https://github.com/lidofinance/lido-dao/pull/311.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -47,15 +48,17 @@ contract LidoOracle is ILidoOracle, AragonApp {
         uint128 beaconValidators;
     }
 
-    struct EpochData {
-        uint256 reportsBitMask;
-        mapping (uint256 => Report) reports;
+    struct ReportKind {
```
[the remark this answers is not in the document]
dechjo: I created `ReportUtils` library here: https://github.com/lidofinance/lido-dao/pull/312/.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -73,153 +76,211 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
-        uint64 _secondsPerSlot,
-        uint64 _genesisTime
-    )
-        public onlyInit
-    {
-        assert(1 == ((1 << (MAX_MEMBERS - 1)) >> (MAX_MEMBERS - 1)));  // static assert
+    /// @dev storage for all gathered reports for the last epoch
+    bytes32 internal constant REPORTABLE_EPOCH_ID_POSITION = keccak256("lido.LidoOracle.reportableEpochId");
+    bytes32 internal constant REPORTS_BITMASK_POSITION = keccak256("lido.LidoOracle.reportsBitMask");
+    ReportKind[] private gatheredReportKinds;
```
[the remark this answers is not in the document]
dechjo: Not yet, but there are only 2 structured variables in both versions. I put them separately here: https://github.com/lidofinance/lido-dao/pull/312/files#diff-6fe7caa502f4e476aef5b7ea5f09cbc22a0cec1b97bf7ff12a816c7f913edeafR85.

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: Good point, I agree, but this actually will lead us to remove `SafeMath` at all.

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: Deleted, actually moved them to `ReportUtils` library here: https://github.com/lidofinance/lido-dao/pull/312.

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: Yes, implemented: https://github.com/lidofinance/lido-dao/pull/313/commits/0c3110f48f59361caefafddf0401c18a30dbf60d. Although this code is run extremally rarely so Vasiliy initially asked me to make it "just works" :)

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -73,153 +76,211 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+     * one-time events that decrease the balance a fair amount - a few percent at a time in a
+     * realistic scenario. Thus, instead of sanity check for an APR, we check if the plain relative
+     * decrease is within bounds.  Note that it's not annual value, its just one-jump value.  May
+     * be changed by the governance.
+     */
+    bytes32 internal constant ALLOWED_BEACON_BALANCE_RELATIVE_DECREASE_POSITION =
+        keccak256("lido.LidoOracle.allowedBeaconBalanceDecrease");
+    uint256 public constant DEFAULT_ALLOWED_BEACON_BALANCE_RELATIVE_DECREASE = 50000;  // 5% ~ 50000 PPM
+
+    function getAllowedBeaconBalanceAnnualRelativeIncrease() public view returns(uint256) {
```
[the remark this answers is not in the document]
dechjo: I updated this (and all other) getters to only return data: no complex logic inside, no access from other parts of the contracts. Do you want to make them external? Please let me know.

contracts/0.4.24/template/LidoTemplate.sol
```diff
@@ -626,8 +626,9 @@ contract LidoTemplate is IsContract {
         perms[0] = _state.oracle.MANAGE_MEMBERS();
         perms[1] = _state.oracle.MANAGE_QUORUM();
         perms[2] = _state.oracle.SET_BEACON_SPEC();
+        perms[3] = _state.oracle.SET_REPORT_BOUNDARIES();
```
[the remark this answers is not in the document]
dechjo: Will do.

contracts/0.4.24/oracle/LidoOracle.sol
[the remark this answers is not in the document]
vshvsh: external with args are cheaper to call than public, but I think argless have no difference

Issue #289, opened 2021-02-05
Add APR calculation to Oracle contract.


dechjo:
We need getters that would show a correct APR for stETH stakers.

dechjo:
Briefly discussed here: https://p2p-validator.slack.com/archives/C017VG67EPL/p1612267340045700

dechjo:
> @vshvsh:
> yeah, or in that case, return not the APR but the raw data from the getter: totalPooledEtherBeforeReport, totalPooledEtherAfterReport, reportTimeDelta (edited) 
As I understand here, only the following is needed to report:
- `totalPooledEtherAfterReport` = last commented `beaconBalance` and 
- `reportTimeDelta` = time elapsed since then.
Because `totalPooledEtherBeforeReport` is not yet known to the contract. Am I right?

vshvsh:
> Because `totalPooledEtherBeforeReport` is not yet known to the contract. Am I right?

You are not. `totalPooledEtherBeforeReport` is trivially queried from Lido contract. `totalPooledEtherAfterReport` does not equal `beaconBalance`

PR #297, opened 2021-02-08
Add getLastCompletedReport() to estimate APR #289

dechjo:


vshvsh:
https://github.com/lidofinance/lido-dao/issues/289#issuecomment-775083588

dechjo:
Should `getLastCompletedReports` be `public` or `external`?

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -328,6 +338,25 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+            uint256 lastTotalPooledEther,
+            uint256 prevTotalPooledEther,
+            uint256 timeElapsed
+        )
+    {
+        BeaconSpec memory beaconSpec = _getBeaconSpec();
+        CompletedReport storage last = lastCompletedReports[COMPLETED_REPORT_ID_LAST];
+        CompletedReport storage prev = lastCompletedReports[COMPLETED_REPORT_ID_PREV];
+        lastTotalPooledEther = last.totalPooledEther;
+        prevTotalPooledEther = prev.totalPooledEther;
```
vshvsh, line 356: is the difference between `lastTotalPooledEther` and `prevTotalPooledEther` profit?

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -83,6 +89,10 @@ contract LidoOracle is ILidoOracle, AragonApp {
     bytes32 internal constant MAX_REPORTED_EPOCH_ID_POSITION = keccak256("lido.LidoOracle.maxReportedEpochId");
     /// @dev storage for all gathered from reports data
     mapping(uint256 => EpochData) private gatheredEpochData;
+    /// @dev historic data about 2 last completed reports
+    bytes32 internal constant COMPLETED_REPORT_ID_LAST = bytes32(0);
+    bytes32 internal constant COMPLETED_REPORT_ID_PREV = bytes32(1);
+    mapping(bytes32 => CompletedReport) private lastCompletedReports;
```
vshvsh, line 95: mapping are inefficient and not work with unstructured storage. better to make it just two constant address structures

vshvsh: or get rid of it alltogether

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -328,6 +338,25 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+            uint256 prevTotalPooledEther,
+            uint256 timeElapsed
+        )
+    {
+        BeaconSpec memory beaconSpec = _getBeaconSpec();
+        CompletedReport storage last = lastCompletedReports[COMPLETED_REPORT_ID_LAST];
+        CompletedReport storage prev = lastCompletedReports[COMPLETED_REPORT_ID_PREV];
+        lastTotalPooledEther = last.totalPooledEther;
+        prevTotalPooledEther = prev.totalPooledEther;
+        timeElapsed = (last.epochId - prev.epochId).mul(beaconSpec.slotsPerEpoch).mul(beaconSpec.secondsPerSlot);
```
vshvsh, line 357: I'm pretty sure we'd be better off caching this and balance before report instead of having two reports stored

vshvsh (changes requested):

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: I think yes, as we discussed. From the steth holder point of view.

dechjo: As discussed, updated the code to avoid accounting sold stETH during the epoch. TODO: try to avoid 2 calls to `Lido.sol` per frame.

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: Fixed by using 4 plain `uint256` variables. Not so elegant, but I agree, it is the best we can do here.

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: Right, implemented this.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -90,9 +85,12 @@ contract LidoOracle is ILidoOracle, AragonApp {
     /// @dev storage for all gathered from reports data
     mapping(uint256 => EpochData) private gatheredEpochData;
     /// @dev historic data about 2 last completed reports
-    bytes32 internal constant COMPLETED_REPORT_ID_LAST = bytes32(0);
-    bytes32 internal constant COMPLETED_REPORT_ID_PREV = bytes32(1);
-    mapping(bytes32 => CompletedReport) private lastCompletedReports;
+    bytes32 internal constant LAST_COMPLETED_TOTAL_POOLED_ETHER_POSITION =
+        keccak256("lido.LidoOracle.lastCompletedTotalPooledEther");
+    bytes32 internal constant PREV_COMPLETED_TOTAL_POOLED_ETHER_POSITION =
```
vshvsh, line 90: It's not prev/last completed total pooled ether - it's pre-report and post-report, I feel it's better to make that disticntion explicit.

dechjo: Renamed to `pre/postTotalPooledEther`. Please feel free to tell me if you meant anything different.

dechjo:
One small idea here:
Combine `PREV_COMPLETED_EPOCH_ID` and `LAST_COMPLETED_EPOCH_ID` into a single slot. Objections?

vshvsh:
> One small idea here:
> Combine `PREV_COMPLETED_EPOCH_ID` and `LAST_COMPLETED_EPOCH_ID` into a single slot. Objections?

No objections

vshvsh (approved):

dechjo:
> > One small idea here:
> > Combine `PREV_COMPLETED_EPOCH_ID` and `LAST_COMPLETED_EPOCH_ID` into a single slot. Objections?
> 
> No objections

Will do this in the final clean-ups.

PR #298, opened 2021-02-09
Sanity checks

vsmelov:


vshvsh:
Beacon balance increase does not imply APR. Lamost all of it is to account for deposited validators, not staking rewards.

vshvsh:
There's no limit to allowed beacon validators amount increase, I think we had ~1000eth in one day.

vshvsh:
https://github.com/lidofinance/lido-dao/issues/39 - see comments here

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -81,9 +81,37 @@ contract LidoOracle is ILidoOracle, AragonApp {
     bytes32 internal constant MIN_REPORTABLE_EPOCH_ID_POSITION = keccak256("lido.LidoOracle.minReportableEpochId");
     /// @dev the max id of reported epochs
     bytes32 internal constant MAX_REPORTED_EPOCH_ID_POSITION = keccak256("lido.LidoOracle.maxReportedEpochId");
+    /// @dev storage for the last completed report and its time
+    bytes32 internal constant LAST_COMPLETED_REPORT = keccak256("lido.LidoOracle.lastCompletedReport");  // why do we need storage? why not just store `Report public lastReport;` attribute?
```
dechjo, line 85: > why do we need storage? why not just store `Report public lastReport;` attribute?
This is [Aragon's unstructured storage](https://github.com/aragon/aragonOS/blob/master/contracts/common/UnstructuredStorage.sol). It protects variables from accidentally changing their slot in the new version of the contract (e.g. the developer added some new variable not to the end of existing variables).

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -416,6 +481,8 @@ contract LidoOracle is ILidoOracle, AragonApp {
             .mul(beaconSpec.epochsPerFrame);
         MIN_REPORTABLE_EPOCH_ID_POSITION.setStorageUint256(minReportableEpochId);
         emit MinReportableEpochIdUpdated(minReportableEpochId);
+
+        LAST_COMPLETED_REPORT.setStorageUint256(_getTime() << 128 | modeReport.beaconBalance);  // fixme: I need modeReport.beaconValidators also!
```
dechjo, line 485: I am thinking if that would be cheaper to go to `lido.getBeaconStat()` and grab it from there (then to store this on our side)?..

vshvsh: I think it won't. Cross apps aragon calls are very expensive

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -81,9 +83,55 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+    So the granularity step for yearly APR is 1%
+    daily_granularity = 0.01 / 365 = 2.74e-05
+    Currently used basisPoints (1e-4) is unable to express such microscopic ratio.
+    So we should take another popular metric to express this - PPM or 1e-6
+    daily_granularity_in_ppm = 2.74e-05 / 1e-6 = 27 (pretty good)
+    with 100% buffer for future spikes it will be 50
+
+    may be changed by Gov
+    */
+    uint128 public ALLOWED_BEACON_BALANCE_INCREASE = 50;  // PPM
```
dechjo, line 104: We need to keep variables in unstructured storage (keccak256 shift, as above) to comply with upgradeability.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -214,6 +263,34 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+     *  we can bound rewards report by 0.1% increase in stake and 15% decrease in stake,
+     *  with both values configurable by DAO voting in case of extremely unusual circumstances.
+     *  daily_reward_rate_PPM = 1e6 * reward / totalPooledEther / days
+     **/
+    function reportSanityChecks(Report report) private {
+        (uint256 lastTotalPooledEther,
+         uint256 curTotalPooledEther,
+         uint256 lastValidators,
+         uint256 timeElapsed) = getLastCompletedReport();
+        // todo: use safeMath everywhere
```
dechjo, line 277: good idea :)

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -81,9 +83,55 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+
+    /*
+    When slashing happens, the balance may decrease at a much faster pace.
+    On the start the system will have about 100 validators and we can guess one of them can be slashed for any reason.
+    1/100 = 10000 PPM
+
+    may be changed by Gov
+    */
+    uint128 public ALLOWED_BEACON_BALANCE_DECREASE = 10000;  // PPM
+    uint256 public constant DAY = 60 * 60 * 24;
```
dechjo, line 114: Looks you can use `1 days` Solidity built-in constant.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -214,6 +263,34 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+     *  with both values configurable by DAO voting in case of extremely unusual circumstances.
+     *  daily_reward_rate_PPM = 1e6 * reward / totalPooledEther / days
+     **/
+    function reportSanityChecks(Report report) private {
+        (uint256 lastTotalPooledEther,
+         uint256 curTotalPooledEther,
+         uint256 lastValidators,
+         uint256 timeElapsed) = getLastCompletedReport();
+        // todo: use safeMath everywhere
+        if (report.curTotalPooledEther >= lastTotalPooledEther) {  // check profit constraint
```
dechjo, line 278: No `report.` here?

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -214,6 +263,34 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+            uint256 reward = curTotalPooledEther - lastTotalPooledEther;
+            uint256 rewardPPM = 1e6 * reward / curTotalPooledEther * DAY / timeElapsed;
+            require(rewardPPM <= ALLOWED_BEACON_BALANCE_INCREASE, "ALLOWED_BEACON_BALANCE_INCREASE");
```
dechjo, line 281: Seems to me, it should be `curTotalPooledEther <= lastTotalPooledEther + timeElapsed * ALLOWED_BEACON_BALANCE_INCREASE / DAY / 1e6`. Here, I try to avoid divisions to save precision.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -214,6 +263,34 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+     * @notice To make oracles less dangerous,
+     *  we can bound rewards report by 0.1% increase in stake and 15% decrease in stake,
+     *  with both values configurable by DAO voting in case of extremely unusual circumstances.
+     *  daily_reward_rate_PPM = 1e6 * reward / totalPooledEther / days
+     **/
+    function reportSanityChecks(Report report) private {
+        (uint256 lastTotalPooledEther,
+         uint256 curTotalPooledEther,
+         uint256 lastValidators,
+         uint256 timeElapsed) = getLastCompletedReport();
```
dechjo, line 276: Well, it's not so good to call `getLastCompletedReport` here, because this function invokes `Lido` contract and we end up with 2 Lido calls per frame (frame=day here): one here and one in line#521. And Lido call is gas-expensive because of Aragon proxy (`AppProxyUpgradeable`). I believe we can save 1 call by doing this: 
- call `uint256 getTotalPooledEther = ISTETH(lido).totalSupply());` as in line#521,
- compare this `getTotalPooledEther` with its previous saved copy `LAST_COMPLETED_REPORT_TOTAL_SUPPLY.getStorageUint256()`,
- if the check passed, save the new value to the copy, otherwise revert.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -202,6 +250,7 @@ contract LidoOracle is ILidoOracle, AragonApp {
 
         uint256 i = 0;
         Report memory report = Report(_beaconBalance, _beaconValidators);
+        reportSanityChecks(report);  // rollback on boundaries violation
```
dechjo, line 253: Looks that we want to make this check 
- not to make it every `reportBeacon`, 
- but after the quorum is reached and we are ready to notify `Lido` contract.

vshvsh (changes requested):

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -214,6 +263,34 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+            uint256 reward = curTotalPooledEther - lastTotalPooledEther;
+            uint256 rewardPPM = 1e6 * reward / curTotalPooledEther * DAY / timeElapsed;
+            require(rewardPPM <= ALLOWED_BEACON_BALANCE_INCREASE, "ALLOWED_BEACON_BALANCE_INCREASE");
+        } else {  // check loss constraint (???)
+            uint256 loss = lastTotalPooledEther - curTotalPooledEther;
+            uint256 lossPPM = 1e6 * loss / curTotalPooledEther * DAY / timeElapsed;
+            require(lossPPM <= ALLOWED_BEACON_BALANCE_DECREASE, "ALLOWED_BEACON_BALANCE_DECREASE");
+        }
+
+        if (report.beaconValidators >= lastValidators) {
```
dechjo, line 288: I think we do not want to limit the growth of validators. This is from our latest discussion in chat, not from the issue.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -328,6 +405,26 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+            uint256 lastValidators,
+            uint256 timeElapsed
+        )
+    {
+        ILido lido = getLido();
+        require(address(lido) != address(0), "LIDO_IS_ZERO");
+        lastTotalPooledEther = LAST_COMPLETED_REPORT_TOTAL_SUPPLY.getStorageUint256();
+        curTotalPooledEther = ISTETH(lido).totalSupply();
+        lastValidators = LAST_COMPLETED_REPORT_VALIDATORS.getStorageUint256();
+        timeElapsed = _getTime() - LAST_COMPLETED_REPORT_TIME.getStorageUint256();
```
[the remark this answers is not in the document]
dechjo: yes, that's my bad, will fix it in #297 today

contracts/0.4.24/oracle/LidoOracle.sol
dechjo, line 281: Could you please move divisions to the right part of inequality for better precision? Smth like `ALLOWED_BEACON_BALANCE_INCREASE * timeElapsed * curTotalPooledEther`.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -81,9 +83,55 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+    uint256 public constant DAY = 60 * 60 * 24;
+    /*
+    Based on this analytics https://etherscan.io/address/0xae7ab96520de3a18e5e111b5eaab095312d7fe84#analytics
+    The max number of daily deposited validators is about 2000 ETH (19 Dec)
+    Divide it by DEPOSIT_SIZE and get the number of daily activated validators 2000 / 32 = 62.5
+    And we leave 50-percent buffer for possible short-time bursts ~= 100
+    */
+    uint128 public ALLOWED_BEACON_VALIDATORS_DAILY_INCREASE = 100;  // validators
+
+    function set_ALLOWED_BEACON_BALANCE_INCREASE(uint128 _ALLOWED_BEACON_BALANCE_INCREASE) external auth(SET_REPORT_LIMITS) {
```
dechjo, line 123: We need to consider the argument as annual. Currently, we treat it as 'daily'  (divide by DAY).

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -81,9 +83,55 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+    uint128 public ALLOWED_BEACON_BALANCE_INCREASE = 50;  // PPM
+
+    /*
+    When slashing happens, the balance may decrease at a much faster pace.
+    On the start the system will have about 100 validators and we can guess one of them can be slashed for any reason.
+    1/100 = 10000 PPM
+
+    may be changed by Gov
+    */
+    uint128 public ALLOWED_BEACON_BALANCE_DECREASE = 10000;  // PPM
```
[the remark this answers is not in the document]
vsmelov: so what do you think it should be?

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -81,9 +83,55 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+    uint256 public constant DAY = 60 * 60 * 24;
+    /*
+    Based on this analytics https://etherscan.io/address/0xae7ab96520de3a18e5e111b5eaab095312d7fe84#analytics
+    The max number of daily deposited validators is about 2000 ETH (19 Dec)
+    Divide it by DEPOSIT_SIZE and get the number of daily activated validators 2000 / 32 = 62.5
+    And we leave 50-percent buffer for possible short-time bursts ~= 100
+    */
+    uint128 public ALLOWED_BEACON_VALIDATORS_DAILY_INCREASE = 100;  // validators
```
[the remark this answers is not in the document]
vsmelov: maybe add some crazy number like 1000000? or its absolutely not needed?

contracts/0.4.24/oracle/LidoOracle.sol
vsmelov: OK I can rename the const to ALLOWED_BEACON_BALANCE_ANNUAL_INCREASE

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -81,9 +83,55 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+
+    function set_ALLOWED_BEACON_BALANCE_INCREASE(uint128 _ALLOWED_BEACON_BALANCE_INCREASE) external auth(SET_REPORT_LIMITS) {
+        ALLOWED_BEACON_BALANCE_INCREASE = _ALLOWED_BEACON_BALANCE_INCREASE;  // to discuss: emit event?
```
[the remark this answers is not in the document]
vsmelov: `event Deposit(address indexed author, uint128 newValue)`
are these parameters enough? (or maybe I can even remove `author`)

contracts/0.4.24/oracle/LidoOracle.sol
vsmelov: Hmmm, I thought the purpose of these boundaries is to deny "crazy" oracles to report and participate in quorum.
So they should be denied on the early stage. Am I right, @vshvsh?

contracts/0.4.24/oracle/LidoOracle.sol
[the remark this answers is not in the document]
dechjo: Why not to create the new event for this? Just add it to ../interfaces/ILido.sol.

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: just a style nit: I suggest using CAPITAL_NAMES for constants. This is just a variable (for this contract). It should be `allowedBeaconBalanceIncrease` or something like that. 
But when you use unstructured storage (keccak256 offsets), they are constants: `ALLOWED_BEACON_BALANCE_INCREASE_POSITION` (note the POSITION in the end).

contracts/0.4.24/oracle/LidoOracle.sol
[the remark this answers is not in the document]
vshvsh: New event, author not needed

contracts/0.4.24/oracle/LidoOracle.sol
vshvsh: They are needed to protect lido from the wrong info, so only once on report beacon is good enough. Those checks are really expensive.

contracts/0.4.24/oracle/LidoOracle.sol
vshvsh: Strictly speaking it's not increase but a relative increase

contracts/0.4.24/oracle/LidoOracle.sol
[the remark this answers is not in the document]
vshvsh: it's checked [in `pushBeacon`](https://github.com/lidofinance/lido-dao/blob/081af6bc7de490b2c57836c2afbad9d53a25aa23/contracts/0.4.24/Lido.sol#L263) anyway. Can skip this check here completely.

contracts/0.4.24/oracle/LidoOracle.sol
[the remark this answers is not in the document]
vshvsh: Off the top of my head it should be between 5% and 15% in one report (not in terms of APR). I don't have a good reason to choose one over other though. 5% is safer in terms of harm a malicious oracle can do, and 15% is safer in form of allowing more real-world situations to be reported without DAO intervention.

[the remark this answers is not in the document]
vshvsh: I'd go with 5% but don't hold that opinion strongly.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -419,8 +516,12 @@ contract LidoOracle is ILidoOracle, AragonApp {
         emit Completed(_epochId, modeReport.beaconBalance, modeReport.beaconValidators);
 
         ILido lido = getLido();
-        if (address(0) != address(lido))
+        if (address(0) != address(lido)) {  // to discuss: use require?
             lido.pushBeacon(modeReport.beaconValidators, modeReport.beaconBalance);
+            LAST_COMPLETED_REPORT_TOTAL_SUPPLY.setStorageUint256(ISTETH(lido).totalSupply()); // = _getTotalPooledEther
+            LAST_COMPLETED_REPORT_TIME.setStorageUint256(_getTime());
```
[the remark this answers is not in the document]
dechjo: > Need to cache the epoch of report (I think it's done already)

Did it in #297.

> Also need to cache time delta between the last and second to last reported epochs.

why do we need it, sorry?

contracts/0.4.24/interfaces/ILido.sol
```diff
@@ -131,4 +131,7 @@ interface ILido {
       * @return beaconBalance - total amount of Beacon-side Ether (sum of all the balances of Lido validators)
       */
     function getBeaconStat() external view returns (uint256 depositedValidators, uint256 beaconValidators, uint256 beaconBalance);
+
+    event SetAllowedBeaconBalanceAnnualRelativeIncrease(uint256 value);
+    event SetAllowedBeaconBalanceAnnualRelativeDecrease(uint256 value);
```
dechjo, line 136: Wrong interface. We need ILidoOracle.sol :)

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -92,6 +93,48 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+        return ALLOWED_BEACON_BALANCE_ANNUAL_RELATIVE_INCREASE_POSITION.getStorageUint256(value);
+    }
+
+    function getAllowedBeaconBalanceAnnualRelativeDecrease() public pure returns(uint256) {
+        return ALLOWED_BEACON_BALANCE_ANNUAL_RELATIVE_DECREASE_POSITION.getStorageUint256(value);
+    }
+
+    function setAllowedBeaconBalanceAnnualRelativeIncrease(uint256 value) external auth(SET_REPORT_BOUNDARIES) {
+        ALLOWED_BEACON_BALANCE_ANNUAL_RELATIVE_INCREASE_POSITION.setStorageUint256(value);
+        emit SetAllowedBeaconBalanceAnnualRelativeIncrease(value);
```
dechjo, line 130: I think it is better to use imperative style: `AllowedBeaconBalanceAnnualRelativeIncreaseSet`.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -92,6 +93,48 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+    daily_granularity = 0.01 / 365 = 2.74e-05
+    Currently used basisPoints (1e-4) is unable to express such microscopic ratio.
+    So we should take another popular metric to express this - PPM or 1e-6
+    daily_granularity_in_ppm = 2.74e-05 / 1e-6 = 27 (pretty good)
+    with 100% buffer for future spikes it will be 50.
+    May be changed by Gov.
+    */
+    bytes32 internal constant ALLOWED_BEACON_BALANCE_ANNUAL_RELATIVE_INCREASE_POSITION =
+        keccak256("lido.LidoOracle.allowedBeaconBalanceAnnualRelativeIncrease");
+    uint256 public constant DEFAULT_ALLOWED_BEACON_BALANCE_ANNUAL_RELATIVE_INCREASE = 50;  // PPM
```
dechjo, line 108: I think there is no need to make it public: this will create a new getter for it. Let's make them all `internal`?

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -222,6 +267,31 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+    function reportSanityChecks(Report report) private {
+        (uint256 postTotalPooledEther,
+         uint256 preTotalPooledEther,
+         uint256 timeElapsed) = getLastCompletedReports();
+        // todo: use safeMath everywhere
+        if (postTotalPooledEther == 0 || timeElapsed == 0) return;  // it's possible at the beginning of the work with the contract or in tests
+        if (postTotalPooledEther >= preTotalPooledEther) {  // check profit constraint
+            uint256 reward = postTotalPooledEther - preTotalPooledEther;
+            uint256 allowedBeaconBalanceAnnualIncreasePPM = getAllowedBeaconBalanceAnnualRelativeIncrease() * postTotalPooledEther;
+            uint256 rewardAnnualizedPPM = 1e6 * reward * 365 days / timeElapsed;
```
dechjo, line 285: `356 days` is `1 years`, if I remember cirrectly.

vsmelov: `1 years` is deprecated

vshvsh (changes requested):

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -92,6 +93,53 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+        keccak256("lido.LidoOracle.allowedBeaconBalanceAnnualRelativeIncrease");
+    uint256 internal constant DEFAULT_ALLOWED_BEACON_BALANCE_ANNUAL_RELATIVE_INCREASE = 50;  // PPM
+
+    /*
+    When slashing happens, the balance may decrease at a much faster pace.
+    On the start the system will have about 100 validators and we can guess one of them can be slashed for any reason.
+    1/100 = 10000 PPM.
+    May be changed by Gov.
+    */
+    bytes32 internal constant ALLOWED_BEACON_BALANCE_ANNUAL_RELATIVE_DECREASE_POSITION =
```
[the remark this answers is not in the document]
vsmelov: it's handled per report but the jump is annualized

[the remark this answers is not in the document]
dechjo: @vshvsh, do you mean that this threshold should be daily instead of annually? 
("Daily" = "per-report" normally, because our reports happen once a frame and the frame is one day long).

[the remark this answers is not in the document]
vshvsh: Not daily. A maximum decrease since the last report, whenever it was, is limited to a fixed relative decrease (e.g. 15%). You had it right in the previous implementation.

contracts/0.4.24/oracle/LidoOracle.sol
vshvsh: it's the other way round, better to default to public data unless we have a reason to hide

contracts/0.4.24/oracle/LidoOracle.sol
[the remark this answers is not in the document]
vshvsh: To explain better:
There can be a very rare, but a sane one-time decrease of total balance by ~5%. 
If we annualize that, we get an unreasonable annualized 1500%-is boundary, which is not sane. Even 10% one-time decrease in two days (e.g. if the oracle is down for two days) is not sane. We need to allow a single slashing to be ~5-10%, not all slasings collectively to be 1500% per year.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -92,6 +93,53 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+        ALLOWED_BEACON_BALANCE_ANNUAL_RELATIVE_INCREASE_POSITION.setStorageUint256(value);
+        emit AllowedBeaconBalanceAnnualRelativeIncreaseSet(value);
+    }
+
+    function setAllowedBeaconBalanceAnnualRelativeDecrease(uint256 value) public auth(SET_REPORT_BOUNDARIES) {
+        ALLOWED_BEACON_BALANCE_ANNUAL_RELATIVE_DECREASE_POSITION.setStorageUint256(value);
+        emit AllowedBeaconBalanceAnnualRelativeDecreaseSet(value);
+    }
+
+    function initialize_v2() public {
```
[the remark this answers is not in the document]
vsmelov: removed the method

contracts/0.4.24/interfaces/ILido.sol
vsmelov: done

contracts/0.4.24/oracle/LidoOracle.sol
vsmelov: thnx, done

contracts/0.4.24/oracle/LidoOracle.sol
vsmelov: reverted to public

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -92,6 +93,53 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+    daily_granularity = 0.01 / 365 = 2.74e-05
+    Currently used basisPoints (1e-4) is unable to express such microscopic ratio.
+    So we should take another popular metric to express this - PPM or 1e-6
+    daily_granularity_in_ppm = 2.74e-05 / 1e-6 = 27 (pretty good)
+    with 100% buffer for future spikes it will be 50.
+    May be changed by Gov.
+    */
+    bytes32 internal constant ALLOWED_BEACON_BALANCE_ANNUAL_RELATIVE_INCREASE_POSITION =
+        keccak256("lido.LidoOracle.allowedBeaconBalanceAnnualRelativeIncrease");
+    uint256 internal constant DEFAULT_ALLOWED_BEACON_BALANCE_ANNUAL_RELATIVE_INCREASE = 50;  // PPM
```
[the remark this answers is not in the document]
vsmelov: done

contracts/0.4.24/oracle/LidoOracle.sol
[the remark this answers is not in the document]
vsmelov: It looks like I should change the mechanism.
What I can do is that I can
1) store one report per week/month and calculate relative jump for a significant amount of time not just between two reports.
2) calculate something like Smoothed EWMA(Balance) and check the jump `currentBalance - EWMA(Balance)`, this way if there will be several consecutive big jumps it's sane (EWMA track the balance smoothly), but if it's rare big jumps then it's OK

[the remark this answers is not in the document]
vsmelov: discussed with @vshvsh 
decided just to leave one jump check (not annualized)
done

contracts/0.4.24/oracle/LidoOracle.sol
vsmelov: check now pls

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -101,31 +101,32 @@ contract LidoOracle is ILidoOracle, AragonApp {
     So we should take another popular metric to express this - PPM or 1e-6
     daily_granularity_in_ppm = 2.74e-05 / 1e-6 = 27 (pretty good)
     with 100% buffer for future spikes it will be 50.
```
vshvsh, line 103: Need to update comments, they are outdated.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -101,31 +101,32 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
     May be changed by Gov.
     */
     bytes32 internal constant ALLOWED_BEACON_BALANCE_ANNUAL_RELATIVE_INCREASE_POSITION =
         keccak256("lido.LidoOracle.allowedBeaconBalanceAnnualRelativeIncrease");
-    uint256 internal constant DEFAULT_ALLOWED_BEACON_BALANCE_ANNUAL_RELATIVE_INCREASE = 50;  // PPM
+    uint256 public constant DEFAULT_ALLOWED_BEACON_BALANCE_ANNUAL_RELATIVE_INCREASE = 100000;  // PPM ~ 10%
 
     /*
     When slashing happens, the balance may decrease at a much faster pace.
     On the start the system will have about 100 validators and we can guess one of them can be slashed for any reason.
```
vshvsh, line 113:
```suggestion
    Slashing are one-time events that decrease the balance a fair amount - a few percent at a time in a realistic scenario. Thus, instead of sanity check for an APR, we check if the plain relative decrease is within bounds.
```

vshvsh (changes requested):

contracts/0.4.24/oracle/LidoOracle.sol
vsmelov: done
just left `If we use APR as a basic reference for increase, and expected range is between 1 and 10% APR.`

contracts/0.4.24/oracle/LidoOracle.sol
vsmelov: done

contracts/0.4.24/oracle/LidoOracle.sol
vshvsh: `expected range is under 10% APR`

vshvsh (approved):

PR #303, opened 2021-02-15
Use one last epoch for voting by oracles

dechjo:


contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -190,36 +159,32 @@ contract LidoOracle is ILidoOracle, AragonApp {
      * @param _beaconValidators Number of validators visible on this epoch
      */
     function reportBeacon(uint256 _epochId, uint128 _beaconBalance, uint128 _beaconValidators) external {
-        (uint256 startEpoch, uint256 endEpoch) = getCurrentReportableEpochs();
-        require(_epochId >= startEpoch, "EPOCH_IS_TOO_OLD");
+        (uint256 currentEpoch, uint256 endEpoch) = getCurrentReportableEpochs();
+        require(_epochId >= currentEpoch, "EPOCH_IS_TOO_OLD");
```
vshvsh, line 163: need to require that `_epochId.mod(beaconSpec.epochsPerFrame) == 0`, otherwise there's a fairly simple DoS attack by a malicious oracle.

vshvsh (changes requested):

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: done and tested

vshvsh (approved):
utACK

Issue #286, opened 2021-02-05
Ability to call arbitrary functions after the `reportBeacon` call.


dechjo:

dechjo:
Should the function be called right after every `reortBeacon` is executed OR after the quorum happened (maybe as a result of `setQuorum` call)? Do we maybe have any specifics for the function to be called (i.e. specific arguments, gas usage, ethers transfer)?
@vshvsh  @ujenjt, what do you think?

vshvsh:
After a successful report to Lido DAO (so after there a quorum, not after every report). 

It should provide the same data as `getLastCompletedReport()` as parameters.

dechjo:
@vshvsh, please advise, who and how often will set this function?
As a draft, I implemented a simple public setter: https://github.com/lidofinance/lido-dao/pull/305/files#diff-6fe7caa502f4e476aef5b7ea5f09cbc22a0cec1b97bf7ff12a816c7f913edeafR158.

PR #305, opened 2021-02-18
Added delegate function call after quorum #286

dechjo:


vshvsh:
"Delegate call" has a specific meaning in eth, let's not use this word

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -152,6 +155,10 @@ contract LidoOracle is ILidoOracle, AragonApp {
         _tryPush();
     }
 
+    function setQuorumDelegate(address _addr, bytes4 _sig) external /* TODO: auth? */ {
```
vshvsh, line 158: only by voting

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -414,6 +421,16 @@ contract LidoOracle is ILidoOracle, AragonApp {
         POST_COMPLETED_TOTAL_POOLED_ETHER_POSITION.setStorageUint256(ISTETH(lido).totalSupply());
         PREV_COMPLETED_EPOCH_ID_POSITION.setStorageUint256(LAST_COMPLETED_EPOCH_ID_POSITION.getStorageUint256());
         LAST_COMPLETED_EPOCH_ID_POSITION.setStorageUint256(epochId);
+
+        // call the quorum delegate
+        uint256 quorumDelegateRaw = QUORUM_DELEGATE_POSITION.getStorageUint256();
+        (uint256 lastTotalPooledEther,
+         uint256 prevTotalPooledEther,
+         uint256 timeElapsed) = getLastCompletedReports();
+        address(quorumDelegateRaw).call(bytes4(quorumDelegateRaw >> 160),
```
vshvsh, line 430: if delegate is not 0

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -86,6 +86,9 @@ contract LidoOracle is ILidoOracle, AragonApp {
     bytes32 internal constant LAST_COMPLETED_EPOCH_ID_POSITION = keccak256("lido.LidoOracle.lastCompletedEpochId");
     bytes32 internal constant PREV_COMPLETED_EPOCH_ID_POSITION = keccak256("lido.LidoOracle.prevCompletedEpochId");
 
+    /// @dev function credentials to be called when the quorum is reached
+    bytes32 internal constant QUORUM_DELEGATE_POSITION = keccak256("lido.LidoOracle.quorumDelegate");
```
vshvsh, line 90: "Delegate call" has a specific meaning in eth, let's not use this word

contracts/0.4.24/oracle/LidoOracle.sol
vshvsh: it can be 0

contracts/0.4.24/oracle/LidoOracle.sol
vshvsh: we can require function sig to be constant (e.g. "processLidoOracleReport") to simplify the code and interfaces

vshvsh (changes requested):

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: it is OK: in this case the call simply returns `false` (which we ignore anyway)

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: yes, sounds good, will do

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: Renaming to `QUORUM_CALLBACK`...

contracts/0.4.24/oracle/LidoOracle.sol
dechjo: moved to using interface `IQuorumCallback` and thus added the check.

vshvsh: I mean that callback can be simply not set. If you call `0x0.processLidoOracleReport(...)` we'll revert.

dechjo: Yes, now did so. And tested.
I meant that previously when I did `address(callback).call(bytes4(keccak256("...")), ...)`, this construction would not revert even if the `callback` was 0.

vshvsh: huh, right. I assumed calling an EOA would revert.

PR #306, opened 2021-02-23
Add more events about state changes, small refactoring

dechjo:
Also a small refactoring:
- removed internal usage of public function `getLastCompletedReports`, use values directly instead.


contracts/0.4.24/interfaces/ILidoOracle.sol
```diff
@@ -55,6 +55,11 @@ interface ILidoOracle {
         uint128 beaconBalance,
         uint128 beaconValidators
     );
+    event ReportPushed(
+         uint256 postTotalPooledEther,
+         uint256 preTotalPooledEther,
+         uint256 timeElapsed,
+         uint256 totalShares);
```
vshvsh, line 62: how about `postTotalShares` for clarity?

vshvsh:
utACK, have a rename suggestion, take it or leave it

contracts/0.4.24/interfaces/ILidoOracle.sol
dechjo: Sure, will rename it. Just a question: should the first three parameters still be reported (`postTotalPooledEther, preTotalPooledEther, timeElapsed`)?

PR #322, opened 2021-03-24
Oracle manual up to date with explicit Docker versions

kolyasapphire:
To make sure node operators would not accidentally update to a new version with breaking changes, let's suggest setting an appropriate image version explicitly.


vshvsh:
@kolyasapphire @ujenjt something to remember: `master` branch is primarily for the deployed, production code on mainnet or working testnets.  Having non-production pre-release version oracle recommended in the `master` docs without a disclaimer that it's for a not-yet-live testnet is wrong. Also, if we don't have relevant docs for production/mainnet and don't want to write them now (which looks to be the case), we should have a disclaimer.

vshvsh (changes requested):
Add a disclaimer there that it's for the testnet; add a placeholder for mainnet instructions.

vshvsh:
Ah, it's recommending the production (outdated for the testnet really) version of an oracle, so all fine on that front, but has instructions for deploying in the testnet.

vshvsh (approved):

Issue #15, opened 2020-09-21
depool.withdraw should clearly signal it's not implemented yet


vshvsh:
As of now [withdraw](https://github.com/depools/depool-dao/blob/bc2364584c5b72b98d9a6531d3632ad03c7e4f2a/apps/depool/contracts/DePool.sol#L226) lets you get a bit of ether from the buffer and burns your tokens + signals that you want to withdraw. At the start there'll be no withdrawal logic implemented yet bc there's no state transitions on eth2 and working temporary solutions around that is too complex for us to possibly fit into the deadline. 

Thus, withdraw function should be there with a correct interface (just like now) but it should be always throwing something along the lines of "NOT_IMPLEMENTED"

PR #329, opened 2021-04-05
Pre testnet updates

dechjo:
1. Move sanity check values to 1e4 precision, same as in `Lido.sol`,
2. Upgrade script #15 add frontend argument.


ujenjt:
Looks good, except I can't see the check script

PR #345, opened 2021-05-16
refactor: create separate functions for padding sigs/keys/hashes

TomAFrench:
I'm currently updating the tests to handle offchain validator keys and am ending up making some changes to the tests which are unrelated to that project. In order to keep that PR as small as possible I'd like to merge some of these changes separately:

I find using `pad("0x1", 48)` and `pad("0x1", 96)` to pad public keys and signatures to the correct length isn't particularly readable, particularly with magic values for the lengths. I've created `padKey` and `padSig` which make it much clearer about what's happening.

This gets even worse with longer amount of keys, we've got a couple of instances of something like the below

```
hexConcat(
    pad('0x010203', 48),
    pad('0x010204', 48),
    pad('0x010205', 48),
    pad('0x010206', 48),
    pad('0x010207', 48),
    pad('0x010208', 48)
)

```

I've refactored these into something of the form of

```
sanitiseKeyArray(['0x010203', '0x010204', '0x010205', '0x010206', '0x010207', '0x010208'])
````

Changelog:
- Created `padKey` and `padSig` functions to handle these special cases of `pad` functionality.
- Added `sanitiseKeyArray` to pad and concatenate an array of keys concisely.
- Consolidated some helper functions which were defined in multiple places.

### CODE REVIEW

File: test/0.4.24/helpers/publicKeyArrays.js [hunk 1]
```diff
@@ -0,0 +1,14 @@
+const { hexConcat, padKey, padSig } = require('../../helpers/utils')
+
+const createKeys = (numKeys) => Array.from({ length: numKeys }, (_value, i) => padKey(`0x${i + 1}`))
```
skozin, line 3: seems to be unused?

test/0.4.24/helpers/publicKeyArrays.js
```diff
@@ -0,0 +1,14 @@
...
+const createSigs = (numSigs) => Array.from({ length: numSigs }, (_value, i) => padSig(`0x${i + 1}`))
```
skozin, line 4: seems to be unused?

test/0.4.24/helpers/publicKeyArrays.js
```diff
@@ -0,0 +1,14 @@
...
+
+const sanitiseKeyArray = (keys) => hexConcat(...keys.map((key) => padKey(key)))
```
skozin, line 6: I think `sanitiseKeyArray` is not a good name for this fn, I'd call it like `packKeys` and `packSigs`.

skozin: Also, I'd not extract these utilities into a separate file: this makes it harder to digest what actually happens in tests.

kadmil:
Hey-hey, @TomAFrench! Thank you for the pull! Most new utils are spot on =) We've left couple codestyle-specific (well, "test-style" would be a bit more appropriate word) comments — please, take a look at those when you'll have the time/

test/0.4.24/helpers/publicKeyArrays.js
TomAFrench: Fair enough, this can wait until the full PR as we're going to be generating much more keys in there.

test/0.4.24/helpers/publicKeyArrays.js
TomAFrench: Good point on the name. I was focusing on the padding but this function does more than that.

This does need to be pushed out to a utils file as when dealing with offchain keys we're going to be manipulating much longer arrays of keys than current so this function would otherwise be duplicated.

PR #361, opened 2021-10-13
add sanity check for pauseIntentValidityPeriodBlocks > 0 and minDepos…

bulbozaur:
https://www.notion.so/c62f21886ac444c69dea97d79f19d53e?v=03783a607cdd41da902fb16023ba6646&p=cb98cc49d5624c549b3b11ff41295190

### DIFF
--- a/contracts/0.8.9/DepositSecurityModule.sol
+++ b/contracts/0.8.9/DepositSecurityModule.sol
@@ -160,6 +160,7 @@ contract DepositSecurityModule {
     }
 
     function _setPauseIntentValidityPeriodBlocks(uint256 newValue) internal {
+        require(newValue > 0, "invalid value for pauseIntentValidityPeriodBlocks: must be greater then 0");
         pauseIntentValidityPeriodBlocks = newValue;
         emit PauseIntentValidityPeriodBlocksChanged(newValue);
     }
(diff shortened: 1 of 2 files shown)


contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -200,6 +201,7 @@ contract DepositSecurityModule {
     }
 
     function _setMinDepositBlockDistance(uint256 newValue) internal {
+        require(newValue > 0, "invalid value");
```
ujenjt, line 204: Maybe change to something readable? Like "minDepositBlockDistance should be greater than zero"?
WDYT?

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -160,6 +160,7 @@ contract DepositSecurityModule {
     }
 
     function _setPauseIntentValidityPeriodBlocks(uint256 newValue) internal {
+        require(newValue > 0, "invalid value");
```
ujenjt, line 163: Maybe change to something readable? Like "pauseIntentValidityPeriodBlocks should be greater than zero"?
WDYT?

ujenjt (approved):
LGTM, excluding wording of error messages

PR #374, opened 2021-12-22
Initial version of Oracle initialization according to new LIP

arwer13:
- restore initialize() function which performs up-to-date initialization from scratch
- keep the latest increment initialize function: namely dummy upgrade from v1 to v3 to correct version number
- still a few TODOs left

### DIFF
--- a/contracts/0.4.24/oracle/LidoOracle.sol
+++ b/contracts/0.4.24/oracle/LidoOracle.sol
@@ -330,6 +330,7 @@ contract LidoOracle is ILidoOracle, AragonApp {
         preTotalPooledEther = PRE_COMPLETED_TOTAL_POOLED_ETHER_POSITION.getStorageUint256();
         timeElapsed = TIME_ELAPSED_POSITION.getStorageUint256();
     }
+
     /**
      * @notice Initialize the contract (version 3 for now) from scratch
      * @dev TODO: Add link to the related LIP
@@ -340,7 +341,6 @@ contract LidoOracle is ILidoOracle, AragonApp {
      * @param _genesisTime Genesis time
      * @param _allowedBeaconBalanceAnnualRelativeIncrease Allowed beacon balance annual relative increase (e.g. 1000 means 10% yearly increase)
      * @param _allowedBeaconBalanceRelativeDecrease Allowed beacon balance moment descreat (e.g. 500 means 5% moment decrease)
-     * @param _lastCompletedEpochId Id of the last completed epoch TODO can it always be zero?
      */
     function initialize(
         address _lido,
@@ -349,8 +349,7 @@ contract LidoOracle is ILidoOracle, AragonApp {
         uint64 _secondsPerSlot,
         uint64 _genesisTime,
         uint256 _allowedBeaconBalanceAnnualRelativeIncrease,
-        uint256 _allowedBeaconBalanceRelativeDecrease,
-        uint256 _lastCompletedEpochId
+        uint256 _allowedBeaconBalanceRelativeDecrease
     )
         external onlyInit
     {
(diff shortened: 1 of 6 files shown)


krogla (approved):

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -73,7 +73,8 @@ contract LidoOracle is ILidoOracle, AragonApp {
     bytes32 internal constant BEACON_SPEC_POSITION =
         0x805e82d53a51be3dfde7cfed901f1f96f5dad18e874708b082adb8841e8ca909; // keccak256("lido.LidoOracle.beaconSpec")
 
-    /// Version of the initialized contract data, v1 is 0
+    /// Version of the initialized contract data
+    /// Verstion at state right after deployment when no initializer is invoked yet is 0
```
TheDZhon, line 77: typo

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -352,16 +380,39 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+        // Initializations for v2 --> v3
+        _initialize_v3();
+
+        // Need this despite contract version check because Aragon requires it to handle auth() modificators properly
+        initialized();
+    }
+
+
+    /**
+     * @notice A function to finalize upgrade to v3 (from v1). Can be called only once
```
TheDZhon, line 398: It's not clear (from the comment context) why we allow upgrading from v1 to v3 and not from v2 to v3.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -352,16 +380,39 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+     * @dev For more details see _initialize_v3()
+     */
+    function finalizeUpgrade_v3() external {
+        require(CONTRACT_VERSION_POSITION.getStorageUint256() == 1, "WRONG_BASE_VERSION");
```
TheDZhon, line 402: Why the previous (pre-upgrade) version is set to `1`?

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -352,16 +380,39 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
-
-        // set expected epoch to the first epoch for the next frame
+        // // set expected epoch to the first epoch for the next frame
         BeaconSpec memory beaconSpec = _getBeaconSpec();
-        uint256 expectedEpoch = _getFrameFirstEpochId(lastReportedEpoch, beaconSpec) + beaconSpec.epochsPerFrame;
+        uint256 expectedEpoch = _getFrameFirstEpochId(0, beaconSpec) + beaconSpec.epochsPerFrame;
         EXPECTED_EPOCH_ID_POSITION.setStorageUint256(expectedEpoch);
         emit ExpectedEpochIdUpdated(expectedEpoch);
+
+        // Initializations for v2 --> v3
```
TheDZhon, line 389: Here we state that v2 → v3 transition is legit for `_inirialize_v3`

TheDZhon (changes requested):
That's a great effort to make things consistent :+1: 

I fave only one significant suggestion in addition to the provided comments. 
Let's add testing coverage for the `finalizeUpgrade_v3`

contracts/0.4.24/oracle/LidoOracle.sol
arwer13: Will add a comment here. Will also consider adding test via Mocking intermediate upgrade functions (from repo history)

TheDZhon: > Will add a comment here. We will also consider adding test via Mocking intermediate upgrade functions (from repo history)
Yeah, cause we are using a cumulative init func only for the testnets while mainnet requires to call `finalizeUpgrade_v3` on top of the current live version.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -352,16 +380,39 @@ contract LidoOracle is ILidoOracle, AragonApp {
...
+        // Need this despite contract version check because Aragon requires it to handle auth() modificators properly
+        initialized();
+    }
+
+
+    /**
+     * @notice A function to finalize upgrade to v3 (from v1). Can be called only once
+     * @dev For more details see _initialize_v3()
+     */
+    function finalizeUpgrade_v3() external {
```
arwer13, line 401: As @TheDZhon noted test coverage is needed for the function. Implementing is postponed till we settle down if we want to keep intermediate initializers (see [example](https://github.com/krogla/contract_versions/blob/master/contracts/ContractVersions.sol) and LIP)

arwer13: We've decided to write no test on it because it' relatively too difficult and not much useful.

TheDZhon:
The code adopter for another PR: #401

PR #377, opened 2021-12-30
Add `transferShares` function for the `stETH` token

TheDZhon:
Use it when rounding matters.


contracts/0.4.24/StETH.sol
```diff
@@ -291,6 +300,29 @@ contract StETH is IERC20, Pausable {
...
+     *
+     * Requirements:
+     *
+     * - `_recipient` cannot be the zero address.
+     * - the caller must have at least `_sharesAmount` shares.
+     * - the contract must not be paused.
+     *
+     * @dev The `_sharesAmount` argument is the amount of shares, not tokens.
+     */
+    function transferShares(address _recipient, uint256 _sharesAmount) public returns (bool) {
```
skozin, line 318: Imo, calling `transferShares(recipient, sharesAmt)` should lead to exactly the same outcome as calling `transfer(recipient, sharesAmt * sharePrice)`, meaning that both `transferShares` and `transfer` should emit `TransferShares` and `Transfer` events, the only difference being the measurement unit of the input amount.

skozin: Also, I'd get rid of the return value: we use it in `transfer` since it's required by ERC20, but we (as well as most other token contracts out there) never actually return `false`, reverting the transaction instead. Keeping this return value here doesn't bring any advantages (since it's not a standard function) and at the same time either complicates the calling code or forces it to rely on the assumption that might possibly change.

skozin: Instead, I'd return the number of stETH tokens transferred. We're already calculating this value to emit the `Transfer` event so it won't consume any additional gas.

skozin (changes requested):

contracts/0.4.24/StETH.sol
TheDZhon: Makes sense; I'll update the code.
Thank you for coming here :+1:

TheDZhon: I think that we have a full resolution with the latest commits bunch :)

contracts/0.4.24/StETH.sol
```diff
@@ -81,6 +81,15 @@ contract StETH is IERC20, Pausable {
      */
     bytes32 internal constant TOTAL_SHARES_POSITION = keccak256("lido.StETH.totalShares");
 
+    /**
+      * @notice An executed shares transfer from `sender` to `recipient`
+      */
+    event TransferShares(
```
skozin, line 87: We define other events in interfaces, e.g. `Stopped` and `Resumed` are defined in `ISTETH.sol`. I think this event should also be moved there for consistency.

TheDZhon: Just did it properly, thx!

skozin (approved):

TheDZhon:
Need to sync on the `IStETH` interface problem. Should we fix and update it or drop altogether?

krogla (approved):

TheDZhon:
The code is adopted for another PR: #401

PR #401, opened 2022-01-30
Merge-ready protocol service pack :package: 

arwer13:
# :package:  Merge-ready protocol service pack

## There is an implementation for the following Lido improvement proposals:
- [LIP-6: In-protocol coverage application mechanism](https://github.com/lidofinance/lido-improvement-proposals/blob/develop/LIPS/lip-6.md) :heart_on_fire: 
- [LIP-7: Composite oracle beacon report receiver](https://github.com/lidofinance/lido-improvement-proposals/blob/develop/LIPS/lip-7.md) :1234: 
- [LIP-8: Increase keysOpIndex in assignNextSigningKeys](https://github.com/lidofinance/lido-improvement-proposals/blob/develop/LIPS/lip-8.md) :key: 
- [LIP-9: Add an explicit log for the stETH burn events](https://github.com/lidofinance/lido-improvement-proposals/blob/develop/LIPS/lip-9.md) :receipt: 
- [LIP-10: Proxy initializations and LidoOracle upgrade](https://github.com/lidofinance/lido-improvement-proposals/blob/develop/LIPS/lip-10.md) :arrow_up: 
- [LIP-11: Add a transfer shares function for stETH](https://github.com/lidofinance/lido-improvement-proposals/blob/develop/LIPS/lip-11.md) :safety_pin: 
- [LIP-12: On-chain part of the rewards distribution after the Merge](https://github.com/lidofinance/lido-improvement-proposals/blob/develop/LIPS/lip-12.md) :moneybag: 
- [LIP-14: Protocol safeguards. Staking rate limiting](https://github.com/lidofinance/lido-improvement-proposals/blob/develop/LIPS/lip-14.md) :safety_vest: 
- [LIP-15: Protocol safeguards. Resume role](https://github.com/lidofinance/lido-improvement-proposals/blob/develop/LIPS/lip-15.md) :runner: 

#### Context
- [Announcement on the Lido research forum](https://research.lido.fi/t/announcement-merge-ready-protocol-service-pack/2184)
- [Snapshot vote](https://snapshot.org/#/lido-snapshot.eth/proposal/0x37f68adc95bf8adb19dc14c0f0734bc4f6330274d3e63fbafae0a12373e8e997)
- [Twitter announcements](https://twitter.com/LidoFinance/status/1526121726639316994) :bird:


TheDZhon (approved):
I am ready to highlight an explicit 'ACK' from my side :+1:

contracts/0.4.24/Lido.sol
```diff
@@ -78,28 +82,38 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
     bytes32 internal constant BEACON_BALANCE_POSITION = keccak256("lido.Lido.beaconBalance");
     /// @dev number of Lido's validators available in the Beacon state
     bytes32 internal constant BEACON_VALIDATORS_POSITION = keccak256("lido.Lido.beaconValidators");
+    /// @dev total amount of Ether MEV and transaction rewards received by Lido contract
+    bytes32 internal constant MEV_TX_FEE_ETHER_POSITION = keccak256("lido.Lido.mevTxFeeEther");
```
ujenjt, line 86: Looks like this one is not used?

ujenjt (changes requested):

contracts/0.4.24/Lido.sol
```diff
@@ -240,6 +281,15 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
         emit WithdrawalCredentialsSet(_withdrawalCredentials);
     }
 
+    /**
+    * @dev Sets given address as the address of LidoMevTxFeeVault contract
+    * @param _mevTxFeeVault MEV and Tx Fees Vault contract address
+    */
+    function setMevTxFeeVault(address _mevTxFeeVault) external auth(SET_MEV_TX_FEE_VAULT_ROLE) {
+        require(isContract(_mevTxFeeVault), "NOT_A_CONTRACT");
+        MEV_TX_FEE_VAULT_POSITION.setStorageAddress(_mevTxFeeVault);
```
ujenjt, line 290: Could we add an event here, like in `setInsuranceFund ` for example?

ujenjt (changes requested):

contracts/0.4.24/Lido.sol
arwer13: yes, thanks, fixed

contracts/0.4.24/Lido.sol
arwer13: added

TheDZhon:
I have closed all previous PRs which were effectively already included into this one.

contracts/0.4.24/template/LidoTemplate.sol
```diff
@@ -653,12 +652,10 @@ contract LidoTemplate is IsContract {
         perms[0] = _state.lido.PAUSE_ROLE();
         perms[1] = _state.lido.MANAGE_FEE();
         perms[2] = _state.lido.MANAGE_WITHDRAWAL_KEY();
-        perms[3] = _state.lido.SET_ORACLE();
+        perms[3] = _state.lido.MANAGE_PROTOCOL_CONTRACTS_ROLE();
```
arwer13, line 655: need to cover the new roles in LidoTemplate.sol: `RESUME_ROLE`, `STAKING_PAUSE_ROLE`, etc
and probably in `protocol-levers.md`, `12-check-dao.js`, `deploy.js` (search by `PAUSE_ROLE` occurrences)

contracts/0.4.24/Lido.sol
```diff
@@ -79,33 +97,145 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
-        _setOracle(_oracle);
-        _setOperators(_operators);
-        _setTreasury(_treasury);
-        _setInsuranceFund(_insuranceFund);
+        NODE_OPERATORS_REGISTRY_POSITION.setStorageAddress(address(_operators));
+        DEPOSIT_CONTRACT_POSITION.setStorageAddress(address(_depositContract));
+
+        _setProtocolContracts(_oracle, _treasury, _insuranceFund);
+
+        STAKE_LIMIT_POSITION.setStorageUint256(
```
arwer13, line 132: According to [lip-10](https://github.com/lidofinance/lido-improvement-proposals/blob/develop/LIPS/lip-10.md#initializer-vs-setter-functions) this kind of initialization should be done by introducing `initialize_vX()` and contract versioning

I also think that if we want to set the staking limit right after the upgrading, we also might want to set it explicitly on initialization after the deployment from scratch.

contracts/0.8.9/LidoMevTxFeeVault.sol
```diff
@@ -0,0 +1,128 @@
...
+        LIDO = _lido;
+        TREASURY = _treasury;
+    }
+
+    /**
+    * @notice Allows the contract to receive ETH
+    * @dev MEV rewards may be sent as plain ETH transfers
+    */
+    receive() external payable {
+        totalRewardsReceivedViaTransactions = totalRewardsReceivedViaTransactions + msg.value;
```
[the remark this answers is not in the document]
TheDZhon: The counter has been removed with the latest commits.

contracts/0.4.24/Lido.sol
```diff
@@ -548,13 +787,13 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
 
     /**
     * @dev Distributes rewards by minting and distributing corresponding amount of liquid tokens.
-    * @param _totalRewards Total rewards accrued on the Ethereum 2.0 side in wei
+    * @param _totalRewards Total rewards occurred on the Ethereum 2.0 side in wei
```
skozin, line 783: Why this change? Rewards accrue, not occur

contracts/0.4.24/lib/StakeLimitUtils.sol
```diff
@@ -0,0 +1,110 @@
...
+//
+// We need to pack four variables into the same 256bit-wide storage slot
+// to lower the costs per each staking request.
+//
+// As a result, slot's memory aligned as follows:
+//
+// LSB ------------------------------------------------------------------------------> MSB
+// 0______________________32______________128_________________________160______________256
+// |______________________|________________|___________________________|________________|
+// | prevStakeBlockNumber | prevStakeLimit | maxStakeLimitGrowthBlocks | maxStakeLimit  |
```
skozin, line 17: From the code it seems that fields are actually stored in the inverse order?

contracts/0.4.24/Lido.sol
```diff
@@ -79,33 +97,145 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
+    function resumeStaking(
+        uint96 _maxStakeLimit,
+        uint96 _stakeLimitIncreasePerBlock
+    ) external {
+        _auth(STAKING_RESUME_ROLE);
+
+        require(_maxStakeLimit >= _stakeLimitIncreasePerBlock, "TOO_LARGE_INCREASE");
+        require(
+            (_stakeLimitIncreasePerBlock == 0)
+            || (_maxStakeLimit / _stakeLimitIncreasePerBlock <= uint32(-1)),
```
skozin: this check belongs to the `StakeLimitUtils` library and not here

contracts/0.4.24/Lido.sol
```diff
@@ -79,33 +97,145 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
+    * To disable rate-limit pass zero arg values.
+    * @dev Reverts if:
+    * - `_maxStakeLimit` < `_stakeLimitIncreasePerBlock`
+    * - `_maxStakeLimit` / `_stakeLimitIncreasePerBlock` >= 2^32 (only if `_stakeLimitIncreasePerBlock` != 0)
+    * Emits `StakeResumed` event
+    * @param _maxStakeLimit max stake limit value
+    * @param _stakeLimitIncreasePerBlock stake limit increase per single block
+    */
+    function resumeStaking(
+        uint96 _maxStakeLimit,
```
skozin, line 178: it's better to have `uint256` argument type for external functions and explicitly check numeric bounds. this will procude more explicit errors on overflow and cheaper gas-wise

skozin (changes requested):

skozin:
^ this is first part of the review, will continue tomorrow

contracts/0.8.9/LidoMevTxFeeVault.sol
```diff
@@ -0,0 +1,121 @@
...
+      *
+      * @param _token an ERC20-compatible token
+      * @param _amount token amount
+      */
+    function recoverERC20(address _token, uint256 _amount) external {
+        require(_amount > 0, "ZERO_RECOVERY_AMOUNT");
+
+        emit ERC20Recovered(msg.sender, _token, _amount);
+
+        require(IERC20(_token).transfer(TREASURY, _amount));
```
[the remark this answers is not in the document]
skozin: (not critical, can leave as is)

contracts/0.4.24/Lido.sol
skozin: (not critical, can leave as is)

contracts/0.4.24/Lido.sol
```diff
@@ -79,33 +97,145 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
         initialized();
     }
 
+    /**
+    * @notice Cut-off new stake (every new staking transaction submitting user-provided ETH
+    * would revert if `pauseStake` was called previously).
+    * @dev A way to pause stake without pushing PAUSE for the whole proto.
+    * The main goal is to prevent huge APR losses for existing stakers due to high demands
+    * on post-Merge entry queue.
+    * Emits `StakingPaused` event.
```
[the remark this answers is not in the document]
skozin: (not critical, can leave as is)

contracts/0.4.24/Lido.sol
```diff
@@ -79,33 +97,145 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
+        );
+
+        emit StakingResumed(_maxStakeLimit, _stakeLimitIncreasePerBlock);
+    }
+
+    /**
+    * @notice Check staking state: whether it's paused or not
+    */
+    function isStakingPaused() external view returns (bool) {
+        return STAKE_LIMIT_POSITION.getStorageUint256().isStakingPaused();
```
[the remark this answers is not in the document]
skozin: (not critical, can leave as is)

contracts/0.4.24/Lido.sol
TheDZhon: yep, moved this one

contracts/0.8.9/LidoMevTxFeeVault.sol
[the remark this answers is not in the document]
kadmil: We had audit comment about the same issue in Jumpgates contract from Oxorio (look for the first finding):
https://github.com/lidofinance/audits/blob/main/Oxorio%20Jumpgate%20Smart%20Contracts%20Security%20Audit%20Report%2005-2022.pdf

contracts/0.4.24/lib/StakeLimitUtils.sol
```diff
@@ -0,0 +1,117 @@
...
+// SPDX-License-Identifier: GPL-3.0
+
+/* See contracts/COMPILERS.md */
+pragma solidity 0.4.24;
+
+//
+// We need to pack four variables into the same 256bit-wide storage slot
+// to lower the costs per each staking request.
+//
+// As a result, slot's memory aligned as follows:
```
skozin, line 12: Need to mention that there's a "staking paused" state encoded by all fields being zero, and "staking unlimited" state encoded by `maxStakeLimit` being zero and `prevStakeBlockNumber` being non-zero.

contracts/0.4.24/Lido.sol
TheDZhon: Generally, I agree, but I propose to leave as is cause we have a contract's size issues :thinking:

contracts/0.4.24/template/LidoTemplate.sol
TheDZhon: Maybe is not an issue cause we decided to minimize the assigned roles, wdyt?

contracts/0.8.9/LidoMevTxFeeVault.sol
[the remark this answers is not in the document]
TheDZhon: Yeah, that was reported by Oxorio recently for the Jumpgates, as @kadmil said. Nice finding, ty :+1:

contracts/0.4.24/Lido.sol
```diff
@@ -33,20 +44,23 @@ import "./StETH.sol";
...
+    bytes32 constant public RESUME_ROLE = keccak256("RESUME_ROLE");
+    bytes32 constant public STAKING_PAUSE_ROLE = keccak256("STAKING_PAUSE_ROLE");
+    bytes32 constant public STAKING_RESUME_ROLE = keccak256("STAKING_RESUME_ROLE");
     bytes32 constant public MANAGE_FEE = keccak256("MANAGE_FEE");
     bytes32 constant public MANAGE_WITHDRAWAL_KEY = keccak256("MANAGE_WITHDRAWAL_KEY");
-    bytes32 constant public SET_ORACLE = keccak256("SET_ORACLE");
+    bytes32 constant public MANAGE_PROTOCOL_CONTRACTS_ROLE = keccak256("MANAGE_PROTOCOL_CONTRACTS_ROLE");
     bytes32 constant public BURN_ROLE = keccak256("BURN_ROLE");
-    bytes32 constant public SET_TREASURY = keccak256("SET_TREASURY");
-    bytes32 constant public SET_INSURANCE_FUND = keccak256("SET_INSURANCE_FUND");
```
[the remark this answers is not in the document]
TheDZhon: Yep, it's planned to happen during the second deployment round. It's hard to afford during the first deploy round cause too many omnibus vote items are planned already.

contracts/0.4.24/Lido.sol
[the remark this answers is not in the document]
TheDZhon: Well, `stETH minting` also maybe not an ideal, cause the reward distribution run will mint new stETH if oracle report is profitable even if new staking requests are on pause. :thinking:

contracts/0.4.24/Lido.sol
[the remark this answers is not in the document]
TheDZhon: Initially, I was enthusiastic about the structures. But then it turned out that `pragma experimental ABIEncoderV2` needed for the `Lido.sol` contract to freely pass and return struct args since we stick to the solidity 0.4 here.

contracts/0.4.24/lib/StakeLimitUtils.sol
TheDZhon: Well, I believe it's not. 
LSB and MSB markers are here to disambiguate since EVM is big-endian under the hood.

TheDZhon:
Following the review changes here: https://github.com/lidofinance/lido-dao/pull/414

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -20,7 +20,7 @@ import "../lib/MemUtils.sol";
   *
   * See the comment of `INodeOperatorsRegistry`.
   *
-  * NOTE: the code below assumes moderate amount of node operators, e.g. up to 50.
+  * NOTE: the code below assumes moderate amount of node operators, i.e. up to 50 (MAX_NODE_OPERATORS_COUNT).
```
loga4, line 23: change to `up to 200`

contracts/0.4.24/Lido.sol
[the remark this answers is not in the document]
folkyatina: It's probably more about stop accepting new stakes or something. If I get it right.

contracts/0.4.24/Lido.sol
```diff
@@ -233,31 +415,56 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
+        _auth(MANAGE_WITHDRAWAL_KEY);
+
         WITHDRAWAL_CREDENTIALS_POSITION.setStorageBytes32(_withdrawalCredentials);
         getOperators().trimUnusedKeys();
 
         emit WithdrawalCredentialsSet(_withdrawalCredentials);
     }
 
+    /**
+    * @dev Sets given address as the address of LidoMevTxFeeVault contract
```
skozin, line 428: "Sets the address of LidoMevTxFeeVault contract"

contracts/0.4.24/StETH.sol
```diff
@@ -413,5 +471,7 @@ contract StETH is IERC20, Pausable {
         // all other token holders. The total supply of the token doesn't change as the result.
         // This is equivalent to performing a send from `address` to each other token holder address,
         // but we cannot reflect this as it would require sending an unbounded number of events.
+
+        // We're emitting StETHBurnt event to provide an explicit rebase log record nonetheless.
```
skozin, line 475: there's no `StETHBurnt` event in the code

contracts/0.4.24/interfaces/ILido.sol
```diff
@@ -79,16 +153,22 @@ interface ILido {
...
+    /**
+    * @dev Sets given address as the address of LidoMevTxFeeVault contract
+    * @param _mevTxFeeVault MEV and Tx Fees Vault contract address
+    */
+    function setMevTxFeeVault(address _mevTxFeeVault) external;
+
+    event LidoMevTxFeeVaultSet(address mevTxFeeVault);
 
     /**
       * @notice Ether on the ETH 2.0 side reported by the oracle
```
skozin, line 167: need to update the docs here (should match the ones in `Lido.sol`)

contracts/0.4.24/lib/StakeLimitUtils.sol
skozin: ah, I see

skozin: maybe, let's still invert the direction in the doc, keeping LSB and MSB markers? it's much easier to read the code that follows in this way as shift operator symbols don't consider endianness

contracts/0.8.9/LidoMevTxFeeVault.sol
```diff
@@ -0,0 +1,120 @@
...
+/* See contracts/COMPILERS.md */
+pragma solidity 0.8.9;
+
+import "@openzeppelin/contracts-v4.4/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts-v4.4/token/ERC721/IERC721.sol";
+import "@openzeppelin/contracts-v4.4/token/ERC20/utils/SafeERC20.sol";
+
+interface ILido {
+    /**
+    * @notice A payable function supposed to be funded only by LidoMevTxFeeVault contract
```
skozin, line 14: funded => called

contracts/0.4.24/Lido.sol
```diff
@@ -128,19 +284,37 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
         return _submit(_referral);
     }
 
+    /**
+    * @notice A payable function for Mev Tx Fee rewards. Can be funded only by LidoMevTxFeeVault contract
```
skozin, line 288: funded => called

contracts/0.4.24/interfaces/ILido.sol
```diff
@@ -58,13 +117,28 @@ interface ILido {
...
+        uint16 insuranceFeeBasisPoints,
+        uint16 operatorsFeeBasisPoints
+    );
 
     event FeeSet(uint16 feeBasisPoints);
 
     event FeeDistributionSet(uint16 treasuryFeeBasisPoints, uint16 insuranceFeeBasisPoints, uint16 operatorsFeeBasisPoints);
 
+    /**
+      * @notice A payable function supposed to be funded only by LidoMevTxFeeVault contract
```
skozin, line 131: funded => called

skozin (changes requested):

contracts/0.4.24/Lido.sol
```diff
@@ -69,7 +84,10 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
     bytes32 internal constant NODE_OPERATORS_REGISTRY_POSITION = keccak256("lido.Lido.nodeOperatorsRegistry");
     bytes32 internal constant TREASURY_POSITION = keccak256("lido.Lido.treasury");
     bytes32 internal constant INSURANCE_FUND_POSITION = keccak256("lido.Lido.insuranceFund");
+    bytes32 internal constant MEV_TX_FEE_VAULT_POSITION = keccak256("lido.Lido.mevTxFeeVault");
```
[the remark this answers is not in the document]
skozin: (not critical, can be left as is)

contracts/0.4.24/Lido.sol
```diff
@@ -79,33 +97,171 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
+    * would revert if `pauseStake` was called previously).
+    * @dev A way to pause stake without pushing PAUSE for the whole proto.
+    * The main goal is to prevent huge APR losses for existing stakers due to high demands
+    * on post-Merge entry queue.
+    * Emits `StakingPaused` event.
+    */
+    function pauseStaking() external {
+        _auth(STAKING_PAUSE_ROLE);
+
+        STAKE_LIMIT_POSITION.setStorageUint256(0);
```
ujenjt, line 150: Maybe let's move to the library. The abstractionMaybe let's move to the library. The abstraction leaked here imho.

contracts/0.4.24/Lido.sol
```diff
@@ -79,33 +97,171 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
+        uint256 currentStakeLimit,
+        uint256 maxStakeLimit,
+        uint256 stakeLimitIncPerBlock,
+        uint256 prevStakeLimit,
+        uint256 prevStakeBlockNumber
+    ) {
+        uint256 slotValue = STAKE_LIMIT_POSITION.getStorageUint256();
+        isStakingPaused = slotValue.isStakingPaused();
+        isStakingLimitApplied = slotValue.isStakingRateLimited();
+        currentStakeLimit = getCurrentStakeLimit();
```
ujenjt, line 256: Maybe slotValue.calculateCurrentStakeLimit() to save some gas?

ujenjt (changes requested):

contracts/0.4.24/Lido.sol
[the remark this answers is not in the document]
ujenjt: I like this approach

contracts/0.4.24/Lido.sol
```diff
@@ -79,33 +97,171 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
+        STAKE_LIMIT_POSITION.setStorageUint256(
+            StakeLimitUtils.encodeStakeLimitSlot(0, 0, 0, uint32(block.number))
+        );
 
         initialized();
     }
 
+    /**
+    * @notice Cut-off new stake (every new staking transaction submitting user-provided ETH
+    * would revert if `pauseStake` was called previously).
```
skozin, line 141: I'd write this instead:

```
@notice Stops accepting new Ether to the protocol.

@dev While accepting new Ether is stopped, calls to the `submit` function,
as well as to the default payable function, will revert.

Emits `StakingPaused` event.
```

I'd skip describing the main intention for this function since this can be misinterpreted as a commitment.

contracts/0.4.24/Lido.sol
```diff
@@ -79,33 +97,171 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
+    * @notice Check staking state: whether it's paused or not
+    */
+    function isStakingPaused() external view returns (bool) {
+        return STAKE_LIMIT_POSITION.getStorageUint256().isStakingPaused();
+    }
+
+    /**
+      * @notice Returns how much Ether can be staked in the current block
+      * @dev Special return values:
+      * - max uint256 if staking is unlimited;
```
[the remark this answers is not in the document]
ujenjt: Maybe uint256(-1)?

contracts/0.4.24/Lido.sol
ujenjt: Oh, I see for consistency let's do:
isStakingLimitApplied ? slotValue.calculateCurrentStakeLimit() : uint256(-1)

contracts/0.4.24/lib/StakeLimitUtils.sol
```diff
@@ -0,0 +1,127 @@
...
+    * @notice check if staking is on pause (i.e. slot contains zero value)
+    */
+    function isStakingPaused(uint256 _slotValue) internal pure returns(bool) {
+        return (_slotValue == 0);
+    }
+
+    /**
+    * @notice check if rate limit is set (otherwise staking is unlimited)
+    */
+    function isStakingRateLimited(uint256 _slotValue) internal pure returns(bool) {
```
ujenjt, line 124: Maybe it's better to rename `isStakingLimitApplied` for consistency? Or rename similar variables in Lido.sol to `isStakingRateLimited `. Wdyt?

contracts/0.4.24/Lido.sol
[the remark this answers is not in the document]
ujenjt: `@notice Stopping accepting new ether and minting stETH in return`
What about that?

contracts/0.4.24/Lido.sol
```diff
@@ -79,33 +97,171 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
+        _auth(STAKING_RESUME_ROLE);
+
+        (
+            uint256 maxStakeLimit,,
+            uint256 prevStakeLimit,
+        ) = STAKE_LIMIT_POSITION.getStorageUint256().decodeStakeLimitSlot();
+
+        // if staking was paused or unlimited previously,
+        // or new limit is lower than previous, then
+        // reset prev stake limit to max
```
ujenjt, line 191:
```suggestion
        // reset prev stake limit to new max stake limit
```

contracts/0.4.24/Lido.sol
```diff
@@ -79,33 +97,171 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
+        if ((maxStakeLimit == 0) || (_maxStakeLimit < prevStakeLimit)) {
```
ujenjt, line 192: Also could be bring to the StakeLimitUtils

ujenjt: resumeStakingWithNewLimit(_maxStakeLimit, _stakeLimitIncreasePerBlock)

contracts/0.4.24/Lido.sol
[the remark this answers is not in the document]
TheDZhon: I am for 2^256 - 1 since it's not so low-level

contracts/0.4.24/Lido.sol
```diff
@@ -440,28 +690,43 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
     * @return amount of StETH shares generated
     */
     function _submit(address _referral) internal whenNotStopped returns (uint256) {
```
ujenjt, line 692: Idea for future may be postponed: save us 2100 gas - remove the modifier and add the code that actually set staking limit to the stop() method.

TheDZhon: You mean that `stop()` also effectively should call `pauseStaking()`? 
Interesting, but maybe non-expected outside :thinking:

contracts/0.4.24/lib/StakeLimitUtils.sol
skozin: btw, big-endian systems store LSB at the largest memory address, so this layout is incorrect anyway

ujenjt: I also think that MSB and LSB needs to be changed here.

TheDZhon: I will update the sketch to make it more readable :+1:

contracts/0.4.24/Lido.sol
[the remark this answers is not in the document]
TheDZhon: It's a cool concept, I am on it. :+1:

contracts/0.4.24/StETH.sol
```diff
@@ -403,6 +458,9 @@ contract StETH is IERC20, Pausable {
         uint256 accountShares = shares[_account];
         require(_sharesAmount <= accountShares, "BURN_AMOUNT_EXCEEDS_BALANCE");
 
+        uint256 amount = getPooledEthByShares(_sharesAmount);
```
Psirex, line 461: The calculation of the amount of burned stETH isn't correct in this line. Cause the burning also decreases the total amount of minted shares, the balance of the account will be decreased by a different amount, so we can't just calculate it as `sharesAmount * pooledEth / totalAmountOfShares`. The simplest correct way to count the burned amount of stETH is `balanceBefore - balanceAfter`.

TheDZhon: I like your idea :+1: . Though, the `SharesBurnt` event has the following definition
```
     * @param amount amount of burnt shares (expressed in stETH just before burning invocation)
```
So it doesn't violate the definition, but your approach sounds more practical.

contracts/0.4.24/Lido.sol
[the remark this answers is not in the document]
arwer13: How has it influenced the gas costs of `_submit()`?

contracts/0.4.24/interfaces/ILido.sol
skozin: you've marked the prev conversation as resolved but the doc still doesn't reflect what we have in `Lido.sol`

contracts/0.8.9/LidoExecutionLayerRewardsVault.sol
```diff
@@ -0,0 +1,120 @@
...
+
+import "@openzeppelin/contracts-v4.4/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts-v4.4/token/ERC721/IERC721.sol";
+import "@openzeppelin/contracts-v4.4/token/ERC20/utils/SafeERC20.sol";
+
+interface ILido {
+    /**
+    * @notice A payable function supposed to be called only by LidoExecLayerRewardsVault contract
+    * @dev We need a separate function because funds received by default payable function
+    * will go through entire deposit algorithm
```
skozin, line 16:
```
* @dev We need a dedicated function because funds received by the default payable function
* are treated as a user deposit
```

contracts/0.8.9/LidoExecutionLayerRewardsVault.sol
```diff
@@ -0,0 +1,120 @@
...
+* happen much less often, only on LidoOracle beacon balance reports
+*/
+contract LidoExecutionLayerRewardsVault {
+    using SafeERC20 for IERC20;
+
+    address public immutable LIDO;
+    address public immutable TREASURY;
+
+    /**
+      * Emitted when the ERC20 `token` recovered (e.g. transferred)
```
skozin, line 35: e.g. => i.e.

contracts/0.8.9/LidoExecutionLayerRewardsVault.sol
```diff
@@ -0,0 +1,120 @@
...
+      * to the Lido treasury address by `requestedBy` sender.
+      */
+    event ERC20Recovered(
+        address indexed requestedBy,
+        address indexed token,
+        uint256 amount
+    );
+
+    /**
+      * Emitted when the ERC721-compatible `token` (NFT) recovered (e.g. transferred)
```
skozin, line 45: e.g. => i.e.

contracts/0.4.24/Lido.sol
```diff
@@ -79,33 +100,149 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
+    /**
+    * @notice Returns internal info about stake limit
+    * @dev Might be used for the advanced integration requests.
+    * @return
+    * `maxStakeLimit` max stake limit
+    * `maxStakeLimitGrowthBlocks` blocks needed to restore max stake limit from the fully exhausted state
+    * `prevStakeLimit` previously reached stake limit
+    * `prevStakeBlockNumber` previously seen block number
+    */
+    function getStakeLimitInternalInfo() external view returns (
```
ujenjt, line 232: There are two field missed here:
```
bool isStakingPaused
bool isStakingLimitApplied
 ```

File: contracts/0.4.24/Lido.sol [hunk 36]
```diff
@@ -350,8 +581,26 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
     }
 
     /**
-      * @notice Gets deposit contract handle
-      */
+    * @notice Get total amount of execution level rewards collected to Lido contract
+    * @dev Ether got through LidoExecutionLayerRewardsVault is kept on this contract's balance the same way
+    * as other buffered Ether is kept (until it gets deposited)
+    * @return uint256 of funds received as execution layer rewards (in wei)
```
skozin, line 587: `@return amount of funds ...` (uint256 => amount)

contracts/0.4.24/interfaces/ILidoExecutionLayerRewardsVault.sol
```diff
@@ -0,0 +1,16 @@
...
+
+pragma solidity 0.4.24;
+
+
+interface ILidoExecutionLayerRewardsVault {
+
+    /**
+    * @notice Withdraw all accumulated execution layer rewards to Lido contract
+    * @param _maxAmount Max amount of ETH to withdraw
+    * @return amount uint256 of funds received as execution layer rewards (in wei)
```
skozin, line 13: `@return amount of funds ...`

contracts/0.8.9/LidoExecutionLayerRewardsVault.sol
```diff
@@ -0,0 +1,120 @@
...
+    */
+    receive() external payable {
+        // no-op
+    }
+
+    /**
+    * @notice Withdraw all accumulated rewards to Lido contract
+    * @dev Can be called only by the Lido contract
+    * @param _maxAmount Max amount of ETH to withdraw
+    * @return amount uint256 of funds received as execution layer rewards (in wei)
```
skozin, line 80: `@return amount of funds ...`

contracts/0.4.24/StETH.sol
ujenjt: Maybe rename `amount` to `tokenAmountBeforeBurn`? To make it more semantically clear?

contracts/0.4.24/StETH.sol
```diff
@@ -81,6 +81,34 @@ contract StETH is IERC20, Pausable {
...
+     * and corresponding stETH amount.
+     * The stETH amount is calculated just before the burning incurred rebase.
+     *
+     * @param account holder of the burnt shares
+     * @param amount amount of burnt shares (expressed in stETH just before burning invocation)
+     * @param sharesAmount amount of burnt shares (expressed in shares themselves)
+     */
+    event SharesBurnt(
+        address indexed account,
+        uint256 amount,
```
ujenjt, line 108: `tokenAmountBeforeBurn`

contracts/0.4.24/StETH.sol
skozin: I'd go with what Eugene suggested

the amount after the burn is easily recoverable via `getPooledEthByShares(event.sharesAmount)`, but the amount before burn cannot be easily obtained

contracts/0.4.24/StETH.sol
```diff
@@ -81,6 +81,34 @@ contract StETH is IERC20, Pausable {
...
+
+    /**
+     * @notice An executed `burnShares` request
+     *
+     * @dev Reports simultaneously burnt shares amount
+     * and corresponding stETH amount.
+     * The stETH amount is calculated just before the burning incurred rebase.
+     *
+     * @param account holder of the burnt shares
+     * @param amount amount of burnt shares (expressed in stETH just before burning invocation)
```
skozin, line 103:
```
@param amount amount of stETH the burnt shares corresponded to before the burn
```

contracts/0.4.24/StETH.sol
```diff
@@ -81,6 +81,34 @@ contract StETH is IERC20, Pausable {
...
+     * @param sharesAmount amount of burnt shares (expressed in shares themselves)
```
skozin, line 104:
```
@param sharesAmount amount of burnt shares
```

contracts/0.4.24/StETH.sol
TheDZhon: @Psirex [suggested](https://github.com/lidofinance/lido-dao/pull/401#discussion_r874516829) to use here the actual balance difference instead, wdyt?


cc: @ujenjt

contracts/0.4.24/StETH.sol
skozin: we can add `tokenAmountAfterBurn`, though

I'd even call these `preRebaseTokenAmount` and `postRebaseTokenAmount`; this will also make it more clear that rebase is happening here

ujenjt: +1 on `preRebaseTokenAmount` and `postRebaseTokenAmount`! I like it!

ujenjt: The main advantage here is that is more usable from subgraph perspective

contracts/0.4.24/Lido.sol
```diff
@@ -79,33 +100,149 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
+        }
+
+        return stakeLimitData.calculateCurrentStakeLimit();
+    }
+
+    /**
+    * @notice Returns internal info about stake limit
+    * @dev Might be used for the advanced integration requests.
+    * @return
+    * `maxStakeLimit` max stake limit
```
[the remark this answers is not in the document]
TheDZhon: it's neat, thx!

contracts/0.8.9/LidoExecutionLayerRewardsVault.sol
```diff
@@ -0,0 +1,120 @@
...
+}
+
+
+/**
+* @title A vault for temporary storage of execution layer rewards (MEV and tx priority fee)
+*
+* These vault replenishments happen continuously through a day, while withdrawals
+* happen much less often, only on LidoOracle beacon balance reports
+*/
+contract LidoExecutionLayerRewardsVault {
```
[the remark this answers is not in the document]
TheDZhon: Yeah, there is a little tradeoff here.

contracts/0.4.24/StETH.sol
```diff
@@ -291,21 +321,46 @@ contract StETH is IERC20, Pausable {
...
+     *
+     * - `_recipient` cannot be the zero address.
+     * - the caller must have at least `_sharesAmount` shares.
+     * - the contract must not be paused.
+     *
+     * @dev The `_sharesAmount` argument is the amount of shares, not tokens.
+     */
+    function transferShares(address _recipient, uint256 _sharesAmount) public returns (uint256) {
+        _transferShares(msg.sender, _recipient, _sharesAmount);
+        emit TransferShares(msg.sender, _recipient, _sharesAmount);
```
Psirex, line 341: Maybe it's better to put the emitting of the `TransferShares` event in the `_transferShares()` method? It might help reduce the bytecode of the contract and remove redundant code.

contracts/0.4.24/Lido.sol
```diff
@@ -153,40 +308,57 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
     }
 
     /**
-      * @notice Stop pool routine operations
```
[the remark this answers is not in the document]
arwer13: Added an [issue](https://github.com/lidofinance/lido-dao/issues/420)

TheDZhon:
Folks, thank you all for the reviews, really appreciate your help and suggestions! 
:+1:

contracts/0.4.24/Lido.sol
```diff
@@ -276,65 +468,104 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
+            }
+        }
+
+        // Don’t mint/distribute any protocol fee on the non-profitable Lido oracle report
+        // (when beacon chain balance delta is zero or negative).
+        // See ADR #3 for details: https://research.lido.fi/t/rewards-distribution-after-the-merge-architecture-decision-record/1535
         if (_beaconBalance > rewardBase) {
             uint256 rewards = _beaconBalance.sub(rewardBase);
-            distributeRewards(rewards);
+            distributeRewards(rewards.add(executionLayerRewards));
```
arwer13, line 497: I see now `distributeRewards()` as not an accurate name. The rewards distribution also happens above in line

    BEACON_BALANCE_POSITION.setStorageUint256(_beaconBalance)
   
when the total supply is increased and all the stETH balances go up.

What happens in `distributeRewards()` is taking part of the rewards as fees and distributing them to the protocol and the node operators.

Not sure about the better naming. As a hint, a naive coarse naming could be like: `takeAndDistributePartOfTheRewardsAsFees()`, `distributeProtocolAndNodeOperatorsRewards`

contracts/0.4.24/StETH.sol
arwer13: +1 This change forces a number of other changes.. Preparing a separate PR to address this.

arwer13: Here is a rough sketch which mostly aims to highlight the scope of the changes required https://github.com/lidofinance/lido-dao/pull/421

contracts/0.4.24/Lido.sol
skozin: I'd either leave it as is or go with smth like `distributeFees` since what it actually does is taking fees (by minting tokens). Smth like `mintFees` is even closer to the truth but might mislead a casual reader.

contracts/0.4.24/StETH.sol
skozin: I suggest leaving this improvement for a later upgrade since we need to finally freeze the code for audit

arwer13: Created issue #422

contracts/0.4.24/Lido.sol
TheDZhon: Let it be `distributeFee`

contracts/0.4.24/StETH.sol
TheDZhon: It's provisionally scheduled for the next protocol upgrade :+1:

TheDZhon:
Here we are, just have merged all the changes from the previous review round :checkered_flag:

TheDZhon:
Let's move on :rocket:

PR #409, opened 2022-04-13
Feat: add staking pause

TheDZhon:
Add submits break feature to address massive demands on the entry queue.
It would allow preventing APR drop for existing Lido stakers if the entry queue became too large.

Context:
* https://blog.lido.fi/modelling-the-entry-queue-post-merge-an-analysis-of-impacts-on-lidos-socialized-model/


contracts/0.4.24/Lido.sol
```diff
@@ -133,6 +136,36 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
+    */
+    function resumeSubmits() external auth(SUBMITS_BREAK_ROLE)  {
+        require(isSubmitsStopped(), "SUBMITS_ALREADY_RESUMED");
+        SUBMITS_BREAK_POSITION.setStorageBool(false);
+        emit SubmitsResumed();
+    }
+
+    /**
+    * @notice check staking break state
+    * Returns true if new staking is cutted-off
```
arwer13, line 162: cutted-off --> cut-off

contracts/0.4.24/Lido.sol
```diff
@@ -50,6 +50,7 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
 
     /// ACL
     bytes32 constant public PAUSE_ROLE = keccak256("PAUSE_ROLE");
+    bytes32 constant public SUBMITS_BREAK_ROLE = keccak256("SUBMITS_BREAK_ROLE");
```
ujenjt, line 53: I prefer solid naming convention is stopSumbits -> SUBMITS_STOPPING_ROLE
Maybe its better to call it pause instead.

ujenjt (changes requested):
Looks good, but I would like to consider better naming and maybe two separate roles.

@skozin @kadmil what do you think?

arwer13:
That's a nice and useful PR, thank you!

I'd favour using "staking" instead of "submits" everywhere. Or some other word but "submit".

More naming options brainstorming: stopAcceptingStakings, pauseFundsAccepting, stopReceivingFunds, stopTakingFunds, stopAcceptingFunds. 

"submit" does clearly refer to `submit(...)` function, but it's quite a technical detail. I think the entire feature is better be expressed in more high-level protocol terms. 

And also because there is no noun "submit".

I'd also prefer a less "loud" word instead of "break": pause or stop. E. g. `STAKING_PAUSE_ROLE` or `STAKING_STOP_ROLE`

arwer13:
We've also got a problem...
New contract size is 25,641 bytes, but the limit is 24,576

arwer13 (changes requested):

contracts/0.4.24/Lido.sol
TheDZhon: Yep, fixed this one

TheDZhon:
I have pushed the update containing refactorings to reduce the contact's size.
Also I have added the separate `RESUME_ROLE` for the whole protocol.

contracts/0.4.24/Lido.sol
```diff
@@ -241,34 +276,24 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
...
+    *
+    * @param _oracle oracle contract
+    * @param _treasury treasury contract which accumulates treasury fee
+    * @param _insuranceFund insurance fund contract which accumulates insurance fee
+    */
+    function setDAOContracts(
+        address _oracle,
+        address _treasury,
+        address _insuranceFund
+    ) external auth(MANAGE_DAO_CONTRACTS_ROLE) {
```
arwer13, line 295: Actually I like how `MANAGE_DAO_CONTRACTS_ROLE` sounds. 
Just want to share a thought. 
That "DAO contracts" may be comprehended as "contracts related to DAO governance"
Maybe "protocol contracts" would be a bit closer

contracts/0.4.24/test_helpers/LidoPushableMock.sol
```diff
@@ -52,7 +52,7 @@ contract LidoPushableMock is Lido {
     }
 
     function initialize(address _oracle) public onlyInit {
-        _setOracle(_oracle);
+        _setDAOContracts(_oracle, _oracle, _oracle);
```
arwer13, line 55: Why not `_setDAOContracts(_oracle, address(0), address(0))`?

contracts/0.4.24/Lido.sol
```diff
@@ -397,23 +418,6 @@ contract Lido is ILido, IsContract, StETH, AragonApp {
         emit RecoverToVault(vault, _token, balance);
     }
 
-    /**
-      * @notice Send NTFs to recovery Vault
-      * @param _token Token to be sent to recovery vault
-      * @param _tokenId Token Id
-      */
-    function transferERC721ToVault(address _token, uint256 _tokenId) external {
```
arwer13, line 405: I'm not sure, but think it's fine to live some time without capability to recover NFT.
Anyway it seems the next Lido.sol iteration gonna come quite soon and it will require refactoring lots of the code to a separate contract. And then there will be enough room to restore NTF recovering.

arwer13 (commented):
Do we also need to add one more deploy script to add new roles and discard the obsolete ones?

contracts/0.4.24/Lido.sol
TheDZhon: Yep, sounds great, ty for the input!

contracts/0.4.24/test_helpers/LidoPushableMock.sol
TheDZhon: There is a check that reverts tx if address is zero. Thus, let's keep it as is.

contracts/0.4.24/Lido.sol
TheDZhon: Yep, I have talked with @ujenjt and we converge to the decision: if needed we would upgrade to recover some highly-priced-NFT stuff.

TheDZhon:
Closed here on behalf of #410

PR #413, opened 2022-05-14
fix: enforce stake limit invariants by the lib

TheDZhon:
Move the 'require' statements about max values inside the library.
Update tests, and give up the absolute block numbering.


loga4 (approved):
LGTM

PR #416, opened 2022-05-17
feat: add exec layer rewards withdrawal limit controls

mymphe:
This PR includes changes to the Lido frontend app,
- display MEV tx fee withdrawal limit (pic 1) as returned from [`getMevTxFeeWithdrawalLimitPoints()` on `Lido.sol`](https://github.com/lidofinance/lido-dao/blob/19d6e425c685ab105d477113211936340969ad44/contracts/0.4.24/Lido.sol#L612), formatted as a percentage (i.e. divided by 100 assuming, e.g. if the return value from `getMevTxFeeWithdrawalLimitPoints()` is 1500, the display value will be 15%);
- allow the user to set a new MEV tx fee withdrawal limit (pic 2) using [`setMevTxFeeWithdrawalLimit(uint256 _limitPoints)` on `Lido.sol`](https://github.com/lidofinance/lido-dao/blob/19d6e425c685ab105d477113211936340969ad44/contracts/0.4.24/Lido.sol#L443) . The input value must be a number from 0 (inclusive) to 100 (inclusive) that can have 1 or 2 decimal places (pic 3);
- update MEV tx fee withdrawal limit displayed whenever `MevTxFeeWithdrawalLimitSet` event is emitted.
---
pic 1.
image

pic 2.
image

pic 3.
image

image

image


TheDZhon:
Nice one :+1: 

Please update the code according to the new naming approach: we are moving from 'mev/tx fee' stuff to 'execution layer rewards'. It's better in terms of abstraction leakage and overall meaning.

mymphe:
@TheDZhon Updated to match the new naming.

apps/lido/arapp.json
```diff
@@ -42,6 +39,11 @@
       "name": "Deposit role",
       "id": "DEPOSIT_ROLE",
       "params": []
+    },
+    {
+      "name": "Set MEV tx fee withdrawal limit role",
+      "id": "SET_MEV_TX_FEE_WITHDRAWAL_LIMIT_ROLE",
```
arwer13, line 45: We also need to update the role names, ie
`SET_MEV_TX_FEE_WITHDRAWAL_LIMIT_ROLE ` --> `SET_EL_REWARDS_WITHDRAWAL_LIMIT_ROLE`

(and update the "name" field above)

arwer13 (changes requested):

apps/lido/arapp.json
mymphe: Done

arwer13 (approved):

apps/lido/app/src/components/ChangeELRewardsWithdrawalLimitSidePanel.js
```diff
@@ -0,0 +1,94 @@
...
+const validationSchema = yup.object().shape({
+  limit: yup
+    .number()
+    .positive()
+    .required()
+    .min(0)
+    .max(100)
+    .test(
+      'limit',
+      `Limit must be an integer or have 1 or 2 decimal places.`,
```
TheDZhon, line 20: The message may sound a bit confusing. 
I propose to rewrite smth like:
`The limit must be a number with up to 2 optional decimal places`

mymphe: Yeah, thought so too, I like yours better :)

TheDZhon (approved):
That's great, thank you! :+1:

mymphe:
Covered in #433

PR #431, opened 2022-05-25
feat: Add ABI linter on CI

dgusakov:
Add ABI actuality linting on each commit + add ad-hoc action to create PR with ABI fix (tested on fork repo see https://github.com/dgusakov/lido-dao/pull/3)

### DIFF
--- a/.github/assert-git-changes.py
+++ b/.github/assert-git-changes.py
@@ -1,10 +1,16 @@
 #!/usr/bin/python3
 import subprocess
+import os
+
+target_dir = os.environ.get("TARGET_DIR")
+
+if not target_dir:
+    print("No TARGET_DIR env variable provided. Exiting")
+    exit(1)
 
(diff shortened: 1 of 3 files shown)


.github/assert-git-changes.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/python3
+import subprocess
+
+git_changes = subprocess.getoutput("git status --porcelain")
+
+print(git_changes)
+
+if git_changes != "":
+    print("Git changes detected! Failing")
```
TheDZhon, line 9: Maybe it's better to check explicitly for the updated files only inside the `lib/abi` dir?
Otherwise, it could be misunderstood in the future and may have compatibility issues.

dgusakov: Agree. Fixed

TheDZhon (approved):
Amazing! Thank you so much, @dgusakov :+1:

PR #434, opened 2022-06-03
DeployedBytecode linter

dgusakov:
Compare compiled bytecode to the deployed one before merge to master.

See https://github.com/lidofinance/lido-dao/actions/runs/2432947164 for demo run

### DIFF
--- a/.github/assert-deployed-bytecode.js
+++ b/.github/assert-deployed-bytecode.js
@@ -13,7 +13,8 @@ function isInsideEmpty32bytes(byteCode, index) {
 }
 
 function stripMetadata(byteCode) {
-  metaDataIndex = byteCode.indexOf('a264')
+  let metaDataLength = parseInt(byteCode.slice(-4), 16) * 2
+  let metaDataIndex = byteCode.length - metaDataLength - 4
   if (metaDataIndex > 0) {
     return byteCode.slice(0, metaDataIndex)
   }
@@ -40,9 +41,6 @@ async function assertByteCode(address, artifactName, deployTx) {
   const artifact = await artifacts.readArtifact(artifactName)
   let bytecodeFromArtifact = artifact.deployedBytecode.toLowerCase()
   const bytecodeFromRpc = (await web3.eth.getCode(address)).toLowerCase()
-  if (artifactName == "Lido") {
-    console.log(bytecodeFromRpc)
-  }
   const ignoreMetadata = IGNORE_METADATA_CONTRACTS.includes(artifactName)
   if (bytecodeFromRpc === bytecodeFromArtifact) {
     console.log(chalk.green(`Compiled bytecode for ${chalk.yellow(address)}(${artifactName}) MATCHES deployed bytecode!`))
--- a/.github/assert-deployed-bytecode.js
+++ b/.github/assert-deployed-bytecode.js
@@ -1,0 +1,74 @@
+const { assert } = require('chai')
+const chalk = require('chalk')
+const { readJSON } = require('../scripts/helpers/fs')
+const { APPS_TO_NAMES, CONTRACTS_TO_NAMES } = require('./deployed-bytecode-consts')
+
+// function deployArgToBytes(arg) {
+//   let coded = ''
+//   if (typeof arg == 'number') {
+//     const hex = arg.toString(16)
+//     coded = '0'.repeat(64 - hex.length) + hex
+//   }
+//   if (typeof arg == 'string') {
+//     const hex = arg.slice(2)
+//     coded = '0'.repeat(64 - hex.length) + hex
+//   }
+//   return coded
+// }
+
+async function assertByteCode(address, artifactName) {
+  const artifact = await artifacts.readArtifact(artifactName)
+  let bytecodeFromArtifact = artifact.deployedBytecode.toLowerCase()
+  const bytecodeFromRpc = (await web3.eth.getCode(address)).toLowerCase()
+  // console.log(`From RPC:\n${bytecodeFromRpc}`)
+  // console.log(`From Artifact:\n${bytecodeFromArtifact}`)
+  assert.isTrue(
+    bytecodeFromRpc === bytecodeFromArtifact,
+    `Compiled bytecode for ${chalk.yellow(address)}(${artifactName}) doesn't match deployed bytecode!`
+  )
+  console.log(chalk.green(`Compiled bytecode for ${chalk.yellow(address)}(${artifactName}) matches deployed bytecode!`))
+}
+
+async function assertDeployedByteCodeMain() {
+  const deployInfo = await readJSON(`deployed-mainnet.json`)
+  // handle APPs
+  const resultsApps = await Promise.allSettled(
+    Object.entries(deployInfo).map(async ([key, value]) => {
+      if (key.startsWith('app:') && !key.startsWith('app:aragon')) {
+        const name = APPS_TO_NAMES.get(key.split(':')[1])
+        if (!name) {
+          throw `Unknown APP ${key}`
+        }
+        const address = value.baseAddress
+        if (!address) {
+          throw `APP ${key} has no baseAddress`
+        }
+        await assertByteCode(address, name)
+      }
+    })
+  )
+  // handle standalone contracts
+  const resultsContracts = await Promise.allSettled(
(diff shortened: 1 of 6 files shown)


.github/assert-deployed-bytecode.js
```diff
@@ -0,0 +1,104 @@
...
+const { readJSON } = require('../scripts/helpers/fs')
+const { APPS_TO_NAMES, CONTRACTS_TO_NAMES, IGNORE_METADATA_CONTRACTS } = require('./deployed-bytecode-consts')
+
+function compareSymbolWise(first, second, ignoreMetadata) {
+  let compareEnd = first.length
+  if (ignoreMetadata) {
+    compareEnd = first.indexOf('a264')
+  }
+  for (var i = 0; i < compareEnd; i++) {
+    if (first[i] != second[i] && first[i] != '0' && second[i] != '0') {
```
TheDZhon, line 13: Maybe it's too dirty for dealing with uninitialized immutable fields, though will work :thinking: 
I mean, immutable fields should have 32 bytes-wide alignment and maybe it could be a good idea to check that zeros are grouped together.

.github/assert-deployed-bytecode.js
```diff
@@ -0,0 +1,104 @@
+const { AssertionError } = require('chai')
+const chalk = require('chalk')
+const { web3 } = require('hardhat')
+const { readJSON } = require('../scripts/helpers/fs')
+const { APPS_TO_NAMES, CONTRACTS_TO_NAMES, IGNORE_METADATA_CONTRACTS } = require('./deployed-bytecode-consts')
+
+function compareSymbolWise(first, second, ignoreMetadata) {
```
TheDZhon, line 7: The function name may be misleading to the reader a bit.
I suggest using smth like `isBytecodeSimilar`

.github/assert-deployed-bytecode.js
```diff
@@ -0,0 +1,104 @@
...
+
+async function assertByteCode(address, artifactName, deployTx) {
+  const artifact = await artifacts.readArtifact(artifactName)
+  let bytecodeFromArtifact = artifact.deployedBytecode.toLowerCase()
+  const bytecodeFromRpc = (await web3.eth.getCode(address)).toLowerCase()
+  const ignoreMetadata = IGNORE_METADATA_CONTRACTS.includes(artifactName)
+  if (bytecodeFromRpc === bytecodeFromArtifact) {
+    console.log(chalk.green(`Compiled bytecode for ${chalk.yellow(address)}(${artifactName}) MATCHES deployed bytecode!`))
+  } else if (
+    bytecodeFromRpc.length == bytecodeFromArtifact.length &&
```
TheDZhon, line 28: why not move the length check into the function?

.github/assert-deployed-bytecode.js
```diff
@@ -0,0 +1,104 @@
...
+    compareSymbolWise(bytecodeFromArtifact, bytecodeFromRpc, ignoreMetadata)
+  ) {
+    console.log(chalk.hex('#FFA500')(`Compiled bytecode for ${chalk.yellow(address)}(${artifactName}) is SIMILAR to deployed bytecode!`))
+    if (deployTx) {
+      await assertByteCodeByDeployTx(address, deployTx, artifact)
+    } else if (!ignoreMetadata) {
```
TheDZhon, line 34: It might be a bit incorrect since if even we ignore IPFS metadata, we don't ignore the immutable vars.

TheDZhon (changes requested):
Super-useful feature, thank you for the rocket speed of delivery :rocket: 
I propose to improve the code slightly to move further :pray:

.github/assert-deployed-bytecode.js
```diff
@@ -0,0 +1,104 @@
...
+      }
+    })
+  )
+  // handle standalone contracts
+  const resultsContracts = await Promise.allSettled(
+    Object.entries(deployInfo).map(async ([key, value]) => {
+      if (!key.startsWith('app:') && key.endsWith('Address')) {
+        const name = CONTRACTS_TO_NAMES.get(key.replace('Address', ''))
+        if (!name) {
+          return
```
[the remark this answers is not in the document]
dgusakov: Current format of deployed-mainnet.json does not allow doing it. There are a lot of xxxAddress attrs that are not related to the contracts.

.github/assert-deployed-bytecode.js
```diff
@@ -0,0 +1,104 @@
+const { AssertionError } = require('chai')
+const chalk = require('chalk')
+const { web3 } = require('hardhat')
+const { readJSON } = require('../scripts/helpers/fs')
+const { APPS_TO_NAMES, CONTRACTS_TO_NAMES, IGNORE_METADATA_CONTRACTS } = require('./deployed-bytecode-consts')
+
+function compareSymbolWise(first, second, ignoreMetadata) {
+  let compareEnd = first.length
+  if (ignoreMetadata) {
+    compareEnd = first.indexOf('a264')
```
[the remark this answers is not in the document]
dgusakov: It will work for the particular case. But as you can see here https://docs.soliditylang.org/en/v0.6.12/metadata.html#encoding-of-the-metadata-hash-in-the-bytecode there can be additional fields there.
BTW how you cont the actual length? I can't figure out

.github/assert-deployed-bytecode.js
dgusakov: Can't get your idea here. This is a workaround for the wstETH. Yet, we can also add `ignoreMetadata` to `assertByteCodeByDeployTx`

.github/assert-deployed-bytecode.js
dgusakov: Fixed

.github/assert-deployed-bytecode.js
dgusakov: fixed

.github/assert-deployed-bytecode.js
dgusakov: fixed

.github/assert-deployed-bytecode.js
```diff
@@ -0,0 +1,127 @@
+const { AssertionError } = require('chai')
+const chalk = require('chalk')
+const { web3 } = require('hardhat')
+const { readJSON } = require('../scripts/helpers/fs')
+const { APPS_TO_NAMES, CONTRACTS_TO_NAMES, IGNORE_METADATA_CONTRACTS } = require('./deployed-bytecode-consts')
+
+const empty32bytesSlot = '00000000000000000000000000000000'
```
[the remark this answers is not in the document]
dgusakov: Fixed

.github/assert-deployed-bytecode.js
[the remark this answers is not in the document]
TheDZhon: I see where are you coming from... I was incorrect initialy about constant offset calculation, indeed.

Though, we still should avoid using `a264`:
>The CBOR mapping can also contain other keys, so it is better to fully decode the data instead of relying on it starting with 0xa264.

I also checked out docs for the [`solc 0.4.24`](https://docs.soliditylang.org/en/v0.4.24/metadata.html), and there is no `0xa264` at all. 

---

The game-changer is the following:
>Since the mapping might contain more keys (see below) and the beginning of that encoding is not easy to find, its length is added in a two-byte big-endian encoding. 

Looks like the best option here is to use the last two bytes encoded metadata length!
We could read them and now we know the exact offset to skip if needed.

I have checked and it works:

* Solidity 0.6.12 docs sample:
```
0xa2
0x64 'i' 'p' 'f' 's' 0x58 0x22 <34 bytes IPFS hash>
0x64 's' 'o' 'l' 'c' 0x43 <3 byte version encoding>
0x00 0x33
```
here:
`0x00 0x33` == 51
`length(0xa2...<3 byte version encoding>)` == 51

* Solidity 0.4.24 docs sample:
```
0xa1 0x65 'b' 'z' 'z' 'r' '0' 0x58 0x20 <32 bytes swarm hash> 0x00 0x29
```
here:
`0x00 0x29` == 41
`length(0xa1...<32 bytes swarm hash>` == 41

.github/assert-deployed-bytecode.js
dgusakov: Yep. Already done this way

TheDZhon (approved):
We take off soon :flight_departure:

PR #437, opened 2022-06-17
update scratch deploy (to include latest protocol upgrades)

arwer13:
NB: this is still a draft
- deploy insurance contracts
- deploy el rewards contract
- start protocol in dao-local-deploy
- add aux script vote-and-enact
- refactor

TODO?: add two-phase voting?


TheDZhon:
I propose to add two-phase voting also :+1:

arwer13:
@TheDZhon Good point. And we could merge into master next week after the omnibus completion.

PR #449, opened 2022-10-26
chore(security): Add storage layout linter

dgusakov:


.github/assert-git-changes.py
```diff
@@ -4,13 +4,15 @@
...
 git_changes = subprocess.getoutput("git status --porcelain")
-print(f"Changes:\n{git_changes}")
 
-if git_changes.find(target_dir) > 0:
-    print(f"Changes in {target_dir} detected! Failing")
-    exit(1)
+if git_changes:
+    print(f"Changes:\n{git_changes}")
+    if not target_dir:
+        print(f"Changes detected! Failing")
```
TheDZhon, line 12: I didn't get the idea of this change. :thinking: 

Previously the case of unset `target_dir` was handled with a more precise error message.

If we needed to handle storage layout changes by using this trick, then I would expect another more explicit approach, tbh.

.github/workflows/storage-layout-update.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: lidofinance/storage-layout-action@v1
+        with:
+          mode: generate
+          src-folder: ./contracts
+          ignore-folders: "{test_helpers,template,mocks}"
+
+      - name: Check for changes
+        id: changes
+        continue-on-error: true
+        run: .github/assert-git-changes.py
```
TheDZhon, line 33: I propose to set an env or pass additional flag to know exactly what are we looking for

.github/workflows/storage-layout-update.yml
```diff
@@ -0,0 +1,43 @@
...
+
+      - name: Create Pull Request
+        if: steps.changes.outcome != 'success'
```
TheDZhon, line 36: Can't be sure that changes are triggered by the changed storage layout

.github/workflows/storage-layout-check.yml
```diff
@@ -0,0 +1,31 @@
...
+
+      - name: Install yarn deps
+        run: yarn install --immutable
+      
+      - name: Check storage-layout
+        uses: lidofinance/storage-layout-action@v1
+        with:
+          mode: check
+          src-folder: ./contracts
+          ignore-folders: "{test_helpers,template,mocks}"
```
TheDZhon, line 31: Do we have a chance to ignore some non-upgradable contracts somehow?

TheDZhon (changes requested):
Thank you for this PR :+1: 
I would suggest to address some implementation issues and consider proposed changes :pray:

.github/assert-git-changes.py
dgusakov: Yes and no))

Current implementation allow both specifying target subdir or analyzing from root. Here we need to analyze root.
As an option, we can explicitly use .storage-linter and rename parameter to `target`

.github/workflows/storage-layout-check.yml
dgusakov: I think we can. We can specify names of the folders to ignore. With contract names it will a bit harder and I would rather not do it

TheDZhon: Folders don't work

.github/assert-git-changes.py
TheDZhon: I am a bit wary of implicit assumptions here.

dgusakov: `target_dir` was renamed to `target`

.github/workflows/storage-layout-check.yml
dgusakov: Than let's keep it as is

.github/workflows/storage-layout-update.yml
dgusakov: Added

.github/workflows/storage-layout-update.yml
dgusakov: Fixed by explicitly setting
```
        env:
          TARGET: .storage-layout
```

PR #454, opened 2022-11-17
feat: assert-bytecode

madlabman:
`artifacts.json` template may be created by the following command:

```bash
jq 'select(.deployedBytecode? | length > 2) | select(.abi? | length > 0) |
      select(.sourceName? | strings | test("test|lib|mock|interface|template|OrderedCallback") == false) |
      {artifactPath: input_filename, sourcePath: .sourceName, name: .contractName}' \
      artifacts/contracts/**/*.json > artifacts.json
```


TheDZhon:
Hey :wave: 

I didn't get the idea of PR. Previously we were trying to retrieve the on-chain bytecode leveraging [`web3.eth.getCode(addr)`](https://github.com/lidofinance/lido-dao/blob/df95e563445821988baf9869fde64d86c36be55f/.github/assert-deployed-bytecode.js#L43). Looks like the current implementation could perform only off-line checks, isn't it? If so, I would like to discuss the rationale behind this decision.

madlabman:
Hej, @TheDZhon 

The idea is the same, just the verification itself was moved to [a separate action](https://github.com/lidofinance/action-verify-bytecode).

TheDZhon (approved):

PR #483, opened 2023-01-10
feat: add bunker mode flag for later use

folkyatina:
Add bunker mode flag to oracle report to be utilised later.


arwer13 (approved):
I would add a TODO about processing the flag in handleCommitteeMemberReport, but it's good to go and without it.

PR #484, opened 2023-01-10
Restoring Withdrawal tests

folkyatina:
- restore basic Withdrawal queue tests
- fix a bug sneaked from WQv0 on request creation 
- fix a bug with shares precision calculation on finalization
- unify assertReverts for custom error solutions


contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -139,7 +139,7 @@ contract WithdrawalQueue {
     uint256 public finalizedRequestsCounter = 0;
 
     /// @notice queue for withdrawal requests
-    WithdrawalRequest[] public queue;
+    WithdrawalRequest[] internal queue;
```
TheDZhon, line 142: why internal?

contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -413,7 +415,7 @@ contract WithdrawalQueue {
             shares -= queue[_firstId - 1].cumulativeShares;
         }
 
-        eth = _min(eth, _toUint128((shares * _shareRate) / 1e9));
+        eth = _min(eth, _toUint128(shares * _shareRate / 1e9));
```
TheDZhon, line 418: maybe 1e27 since `sharesRate` has 1e27 precision multiplier?

contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -498,7 +500,7 @@ contract WithdrawalQueue {
             WithdrawalRequest memory prevRequest = queue[requestId - 1];
 
             cumulativeShares += prevRequest.cumulativeShares;
-            cumulativeShares += prevRequest.cumulativeEther;
+            cumulativeEther += prevRequest.cumulativeEther;
```
TheDZhon, line 503: haha, was my infamous WQ0 copypasta, thanks for catching :+1:

test/0.8.9/withdrawal-queue.test.js
```diff
@@ -1,30 +1,34 @@
...
     withdrawal = await WithdrawalQueue.at(withdrawalProxy.address)
+
     await withdrawal.initialize(daoAgent)
     await withdrawal.resumeRequestsPlacement({ from: daoAgent })
+
+    steth.mintShares(user, 1)
+    steth.approve(withdrawal.address, StETH(300), { from: user })
   })
 
   context('Enqueue', async () => {
```
TheDZhon, line 34: Should check that by-default the queue is paused (can't enqueue requests)

test/0.8.9/withdrawal-queue.test.js
```diff
@@ -1,30 +1,34 @@
...
+const { assertRevert } = require('../helpers/assert')
 
 const WithdrawalQueue = artifacts.require('WithdrawalQueue.sol')
-const Owner = artifacts.require('Owner.sol')
-const StETHMock = artifacts.require('StETHMockForWithdrawalQueue.sol')
+const StETHMock = artifacts.require('StETHMock.sol')
 const WstETH = artifacts.require('WstETH.sol')
 const OssifiableProxy = artifacts.require('OssifiableProxy.sol')
 
-contract('WithdrawalQueue', ([recipient, stranger, daoAgent]) => {
```
TheDZhon, line 13: Though, these tests should be corrected once ACL PR merged :eyes:

test/helpers/assert.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+  // assert.fail() for some reason does not have its error string printed 🤷
+  assert(
+    false,
+    `Expected "${expectedErrorCode}"${expectedReason ? ` (with reason: "${expectedReason}")` : ''
+    } but it did not fail`
+  )
+}
+
+// version of @aragon/contract-helpers-test assertRevert, but with custom errors support
```
TheDZhon, line 63: There was a patch (https://github.com/lidofinance/lido-dao/blob/master/patches/%40aragon%2Bcontract-helpers-test%2B0.1.0.patch)
looks like a duplicate solution

test/helpers/utils.js
```diff
@@ -57,6 +57,7 @@ const div15 = (bn) => bn.div(new BN(1000000)).div(new BN(1000000)).div(new BN(10
 
 const ETH = (value) => web3.utils.toWei(value + '', 'ether')
 const tokens = ETH
+const shareRate = (value) => web3.utils.toWei(value + '', 'gether')
```
TheDZhon, line 60: haven't met `gether` before :eyes: :+1:

TheDZhon (changes requested):
LGTM, though I have some questions and concerns :eyes:

contracts/0.8.9/WithdrawalQueue.sol
folkyatina: Nope, here we're converting `shareRate` with 1e27 precision to `eth` with 1e18 precision. So, we have to get rid of additional 1e9. Will add a comment.

contracts/0.8.9/WithdrawalQueue.sol
folkyatina: No need to expose internal cumulative nature of the queue to users. There are more convenient methods, like `getWithdrawalRequestStatus` and others.

test/0.8.9/withdrawal-queue.test.js
folkyatina: Yep, but here, we are just restoring the old tests, we are to write many more tests for the newer functions later, of cause.

test/0.8.9/withdrawal-queue.test.js
folkyatina: Yep, we'll need a lot more tests, actually. But will add it later.

contracts/0.8.9/WithdrawalQueue.sol
folkyatina: Ok. Here you are right because we have shares with 1e18 precision too, so we need to reduce 1e27 of zeroes. Thanks. Thats a catch.

test/helpers/assert.js
loga4: +1 about truffle

contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -413,7 +417,7 @@ contract WithdrawalQueue {
             shares -= queue[_firstId - 1].cumulativeShares;
         }
 
-        eth = _min(eth, _toUint128((shares * _shareRate) / 1e9));
+        eth = _min(eth, _toUint128(shares * _shareRate / SHARE_RATE_PRECISION));
```
TheDZhon, line 420: ♥

TheDZhon (approved):
:+1:

PR #487, opened 2023-01-12
feat: add AccessControl to managing ACL in WQ

folkyatina:
Add role-based access control to WithdrawalQueue

There are three roles:
- PAUSE_ROLE
- RESUME_ROLE
- FINALIZE_ROLE 

and pause now is for finalization also


contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -91,24 +93,19 @@ contract WithdrawalQueue {
...
 
-    /// Lido DAO Agent contract address
-    /// Used to call administrative levers
-    bytes32 internal constant LIDO_DAO_AGENT_POSITION = keccak256("lido.WithdrawalQueue.lidoDAOAgent");
-
-    /// Requests placement resume/pause control storage slot
-    bytes32 internal constant REQUESTS_PLACEMENT_RESUMED_POSITION =
-        keccak256("lido.WithdrawalQueue.requestsPlacementResumed");
+    /// Withdrawal queue resume/pause control storage slot
+    bytes32 internal constant RESUMED_POSITION = keccak256("lido.WithdrawalQueue.resumed");
```
arwer13, line 97: Should claims be paused as well? They are not not. If not, I like full name with reference to what part of functionality if paused.

contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -390,7 +439,7 @@ contract WithdrawalQueue {
      *
      * @return hint rate index for this request
      */
-    function findRateHint(uint256 _requestId) public view returns (uint256 hint) {
+    function findClaimingRateHint(uint256 _requestId) public view returns (uint256 hint) {
```
arwer13, line 442: maybe `findClaimRateHint` is a bit cleare

contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -148,42 +149,54 @@ contract WithdrawalQueue {
...
-        if (_lidoDAOAgent == address(0)) {
-            revert LidoDAOAgentZeroAddress();
+    /**
+     * @param _admin admin address that can change every role.
+     * @param _pauser address that will be able to pause the withdrawals
+     * @param _resumer address that will be able to resume the withdrawals after pause
+     * @param _finalizer address that can finalize requests in the queue
+     * @dev Reverts with `AdminZeroAddress()` if `_admin` equals to `address(0)`
+     */
+    function initialize(address _admin, address _pauser, address _resumer, address _finalizer) external {
```
arwer13, line 171: Should the contract be in "paused" state upon deployment and initialization? If so, let's mention it explicitly.

contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -91,24 +93,19 @@ contract WithdrawalQueue {
...
-    /**
-     * @notice All state-modifying calls are allowed only from owner protocol.
-     * @dev should be Lido
-     */
-    address payable public immutable OWNER;
+    // ACL
+
+    bytes32 public constant PAUSE_ROLE = keccak256("PAUSE_ROLE");
+    bytes32 public constant RESUME_ROLE = keccak256("RESUME_ROLE");
+    bytes32 public constant FINALIZE_ROLE = keccak256("FINALIZE_ROLE");
```
arwer13, line 108: Would like to ponder if we need finalizer role or just set Lido address, but good to go as it is.

arwer13 (changes requested):
LGTM, except for a few comments.

contracts/0.8.9/WithdrawalQueue.sol
folkyatina: Hypothetically, it can be assigned to some other intermediary contract, to oracle or even be invoked manually. So I'd like to have some agility here. At least, assuming that we have a generalised part for handling roles here.

contracts/0.8.9/WithdrawalQueue.sol
folkyatina: Probably.

contracts/0.8.9/WithdrawalQueue.sol
folkyatina: Don't think that it worth it. At least in the name. REQUESTS_PLACEMENT_AND_FINALIZATION_RESUMED_POSITION looks too monstrous from my PoV. But I'd like to mention it in the docs. Thanks.

contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -91,24 +93,19 @@ contract WithdrawalQueue {
...
-    bytes32 internal constant LIDO_DAO_AGENT_POSITION = keccak256("lido.WithdrawalQueue.lidoDAOAgent");
-
-    /// Requests placement resume/pause control storage slot
-    bytes32 internal constant REQUESTS_PLACEMENT_RESUMED_POSITION =
-        keccak256("lido.WithdrawalQueue.requestsPlacementResumed");
+    /// Withdrawal queue resume/pause control storage slot
+    bytes32 internal constant RESUMED_POSITION = keccak256("lido.WithdrawalQueue.resumed");
 
     /// Lido stETH token address to be set upon construction
     address public immutable STETH;
```
TheDZhon, line 100: Should be placed in unstructured storage eventually

contracts/0.8.9/WithdrawalQueue.sol
TheDZhon, line 97: Claims should be paused together with placements, but this short name is totally ok (speaking about the previous convos).

contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -148,64 +149,94 @@ contract WithdrawalQueue {
...
-
+    constructor(address _stETH, address _wstETH) {
         // init immutables
         STETH = _stETH;
         WSTETH = _wstETH;
-        OWNER = _owner;
 
-        // petrify the implementation by assigning a zero Lido agent address
-        _initialize(address(0));
+        // petrify the implementation by assigning a zero addresses for every role
```
TheDZhon, line 160: a zero address

contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -148,64 +149,94 @@ contract WithdrawalQueue {
...
+        emit WithdrawalQueuePaused();
+    }
+
+    /// @notice Returns whether the requests placement and finalization is paused or not
+    function isPaused() external view returns (bool) {
+        return !RESUMED_POSITION.getStorageBool();
+    }
+
+    /// @notice Returns the length of the withdrawal request queue 
     function queueLength() external view returns (uint256) {
```
TheDZhon, line 214: maybe add the helpers `unfinalizedQueueLength` and `unfinalizedStETH` also for the frontend devs?

TheDZhon (approved):
LGTM for testnets :+1: 
Though, left some notes on the fields for the future.

contracts/0.8.9/WithdrawalQueue.sol
folkyatina: Ok. Let's do the safe way at first. But from my point of view, it's not fair to pause already locked ether.

contracts/0.8.9/WithdrawalQueue.sol
folkyatina: Should it? Not sure that it can be changed in a some way.

PR #490, opened 2023-01-13
feat: sanity check for bufferedEtherReserve

folkyatina:
If we're trying to update bufferedEtherReserve to be less than needed for the last unfinalized request, make it equal to last unfinalized request's stETH amount.

It will allow us to avoid clogging the queue with a large request.


contracts/0.4.24/Lido.sol
```diff
@@ -455,29 +444,32 @@ contract Lido is StETH, AragonApp {
         // EL values
         uint256 _withdrawalVaultBalance,
         // decision
-        uint256 _withdrawalsReserveAmount,
+        uint256 _newBufferedEtherReserveAmount,
```
arwer13, line 447: Maybe `_newBufferedEtherDepositReserve`?
or even  `_newBufferedEtherNonDepositReserve`
or `_newBufferNonDepositReserve`, `_newNonDepositBufferReserve`
cause it's reserve not for deposits, but on contrary for NOT deposits

In general, like when there is no need to go through usage and see for what "reserve" is, if it has unique usage.

contracts/0.4.24/Lido.sol
```diff
@@ -455,29 +444,32 @@ contract Lido is StETH, AragonApp {
...
     ) external {
         require(msg.sender == getOracle(), "APP_AUTH_FAILED");
         _whenNotStopped();
 
-        // update withdrawals reserve
-        WITHDRAWAL_RESERVE_POSITION.setStorageUint256(_withdrawalsReserveAmount);
-
         uint256 preBeaconBalance = BEACON_BALANCE_POSITION.getStorageUint256();
 
+        // updating saved stats from beacon balance checking its sanity
```
arwer13, line 456: "updating" --> "update" to make uniform with other nearby comments

but sorry, yeah, this is not of importance

contracts/0.4.24/Lido.sol
```diff
@@ -645,13 +645,32 @@ contract Lido is StETH, AragonApp {
...
+        uint256 minEtherReserve = 0;
+
+        // We are trying to avoid the situation when withdrawal queue is blocked by large request
+        // while ether is constantly spent to deposit new validators
+        // so if we pass a reserve amount less than amount of eth required to finalize the last request
+        // in the queue, we are updating it to that amount
+        address withdrawalQueueAddress = getWithdrawalQueue();
+
+        if (withdrawalQueueAddress != address(0)) {
+            IWithdrawalQueue withdrawalQueue = IWithdrawalQueue(withdrawalQueueAddress);
```
arwer13, line 659: are some many separator new lines intended in the body of if?

arwer13 (approved):

contracts/0.4.24/Lido.sol
```diff
@@ -620,7 +620,7 @@ contract Lido is StETH, AragonApp {
     }
 
     /**
-     * @dev updates beacon state
+     * @dev updates beacon state according to fresh report
```
TheDZhon, line 623: to the current report?

TheDZhon (approved):
Glanced briefly, no blockers for merge

contracts/0.4.24/Lido.sol
folkyatina: Not sure. Generally, I don't like too long names. Sometimes it's better to follow and check the overall logic and docs instead of trying to figure out the exact meaning if it's not clear from the start.

PR #500, opened 2023-01-18
Fix: const gas `calculateCurrentStakeLimit` ver

### LINKED ISSUE #499: Volatile `gasLimit` for `_submit()` method
**Context**
For the `_submit()` transactions, the estimated gas limit is not correct sometimes.
The problem was introduced with the Merge-ready protocol upgrade.
On average, every 8th transaction is reverted

**Purpose**
The difference appears due to the condition within the `calculateCurrentStakeLimit() ` method of the [`StakeLimitUtils`](https://github.com/lidofinance/lido-dao/blob/master/contracts/0.4.24/lib/StakeLimitUtils.sol#L105) library.
Depending on the `block.number` and with the same state, the gas consumption changes when condition is met, imposing additional `80 gas`. 

**Solution**
The possible solution to overcome the issue is to make the method gas consumption constant with respect to the current `block.number` (if the same state invariant is preserved)

TheDZhon:
Fixes #499.

### DIFF
--- a/contracts/0.4.24/lib/StakeLimitUtils.sol
+++ b/contracts/0.4.24/lib/StakeLimitUtils.sol
@@ -210,10 +210,12 @@ library StakeLimitUtils {
     /**
      * @notice find a minimum of two numbers with a constant gas consumption
      * @dev doesn't use branching logic inside
+     * @param _lhs left hand side value
+     * @param _rhs right hand side value
      */
-    function _constGasMin(uint256 l, uint256 r) internal pure returns (uint256 min) {
-        uint256 lIsLess = _toUInt256(l < r);
-        min = l * lIsLess + r * (1 - lIsLess);
+    function _constGasMin(uint256 _lhs, uint256 _rhs) internal pure returns (uint256 min) {
+        uint256 lhsIsLess = _toUInt256(_lhs < _rhs);
+        min = _lhs * lhsIsLess + _rhs * (1 - lhsIsLess);
     }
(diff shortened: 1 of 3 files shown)


contracts/0.4.24/lib/StakeLimitUtils.sol
```diff
@@ -94,17 +94,18 @@ library StakeLimitUnstructuredStorage {
...
         if (_data.maxStakeLimitGrowthBlocks != 0) {
             stakeLimitIncPerBlock = _data.maxStakeLimit / _data.maxStakeLimitGrowthBlocks;
         }
 
-        limit = _data.prevStakeLimit + ((block.number - _data.prevStakeBlockNumber) * stakeLimitIncPerBlock);
-        if (limit > _data.maxStakeLimit) {
-            limit = _data.maxStakeLimit;
-        }
+        limit = _constGasMin(
+            _data.prevStakeLimit + ((block.number - _data.prevStakeBlockNumber) * stakeLimitIncPerBlock),
```
folkyatina, line 106: Can probably use another variable with clear naming for this formula. It will improve readability a bit.

folkyatina (approved):
LGTM

contracts/0.4.24/lib/StakeLimitUtils.sol
TheDZhon: Yeah, refined this part, thank you :+1:

PR #503, opened 2023-01-20
Fix scratch deploy for shapella devnet-3

arwer13:
Fix and refactor scratch deployment of shapella version of the protocol.


krogla (approved):
lgtm

PR #505, opened 2023-01-23
Staking router external params to uint256

loga4:


contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -316,7 +323,7 @@ contract DepositSecurityModule {
         int256 guardianIndex = _getGuardianIndex(msg.sender);
 
         if (guardianIndex == -1) {
-            bytes32 msgHash = keccak256(abi.encodePacked(PAUSE_MESSAGE_PREFIX, blockNumber, stakingModuleId));
+            bytes32 msgHash = keccak256(abi.encodePacked(PAUSE_MESSAGE_PREFIX, blockNumber, uint24(stakingModuleId)));
```
TheDZhon, line 326: it's a bit inconsistent maybe to have such a type cast here because `blockNumber` is `uint256`. for example
maybe use `uint256` for `stakingModuleId` too?

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -391,7 +398,7 @@ contract DepositSecurityModule {
         uint256 onchainKeysOpIndex = STAKING_ROUTER.getStakingModuleKeysOpIndex(stakingModuleId);
         require(keysOpIndex == onchainKeysOpIndex, "keys op index changed");
 
-        _verifySignatures(depositRoot, blockNumber, blockHash, stakingModuleId, keysOpIndex, sortedGuardianSignatures);
+        _verifySignatures(depositRoot, blockNumber, blockHash, uint24(stakingModuleId), keysOpIndex, sortedGuardianSignatures);
```
TheDZhon, line 401: same as above

TheDZhon (approved):
LGTM, though, not sure about type casts for signature checking stuff

TheDZhon (approved):

contracts/0.8.9/DepositSecurityModule.sol
loga4: done

PR #507, opened 2023-01-24
Hardhat task to generate userdoc artifacts

manneredboor:
I needed to generate userdoc artifacts for dao-voting-ui descriptions feature.
I have made hardhat task for this purpose, contributing it in case it will be useful in the future.


TheDZhon (approved):
Thank you for the contribution :heart:

PR #509, opened 2023-01-26
fix: changing recipient works consistent

rkolpakov:
Function `changeRecipient` now changes `requestsByRecipient` mapping by moving `requestId` from sender to recipient.


contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -318,7 +320,7 @@ contract WithdrawalQueue is AccessControlEnumerable {
 
     /// @notice Returns all withdrawal requests placed for the `_recipient` address
     function getWithdrawalRequests(address _recipient) external view returns (uint256[] memory requestsIds) {
```
TheDZhon, line 322: maybe put a warning regarding unexpected gas spending (similar to the provided [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v4.4/contracts/utils/structs/EnumerableSet.sol#L342:L345))

TheDZhon, line 322: maybe should have also a range-bounded version (aka `pagination`)

contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -416,6 +418,9 @@ contract WithdrawalQueue is AccessControlEnumerable {
         if (request.claimed) revert RequestAlreadyClaimed();
 
         request.recipient = payable(_newRecipient);
+
```
TheDZhon, line 421: maybe add an early return if the new recipient is the same as the old one

test/0.8.9/withdrawal-queue.test.js
```diff
@@ -156,4 +156,50 @@ contract('WithdrawalQueue', ([recipient, stranger, daoAgent, user]) => {
       assertBn(await ethers.provider.getBalance(recipient), balanceBefore.add(bn(ETH(50))))
     })
   })
+
+  context('Transfer request', async () => {
```
TheDZhon, line 160: I would add some additional cases:
- auth error
- already claimed request (can't transfer)
- passing the same recipient
- wrong request id

TheDZhon (commented):
Hey :wave: 

Thank you for chasing up :articulated_lorry:

PR #521, opened 2023-01-31
Abolish interface contracts

arwer13:


contracts/0.4.24/Lido.sol
```diff
@@ -8,16 +8,50 @@ pragma solidity 0.4.24;
...
+        );
+
+    function reportRewardsMinted(uint256[] _stakingModuleIds, uint256[] _totalShares) external;
+
+    function deposit(uint256 maxDepositsCount, uint256 stakingModuleId, bytes depositCalldata) external payable returns (uint256);
+
+    function getWithdrawalCredentials() external view returns (bytes32);
+}
+
+interface IWithdrawalVault {
```
folkyatina, line 38: Move it closer to  ILidoExecutionLayerRewardsVault?

folkyatina (approved):

contracts/0.4.24/Lido.sol
arwer13: Done

arwer13:
@folkyatina 
> What do you think about merge two `IStakingModule` to one in `commons/interfaces`? We already have `IEIP712.sol` there.

The problem is that `IStakingModule` has `bytes calldata _calldata` arg in `requestValidatorsKeysForDeposits` what makes it compilable either in 0.4.24 or 0.8.9.
We cannot either make common interface and inherit it in 0.4.24 and 0.8.9 defining `requestValidatorsKeysForDeposits` with `bytes _calldata` and with `bytes calldata _calldata` respectively.

PR #525, opened 2023-02-01
Fix missing gas reports for contracts with the new Solidity version

Psirex:
The gas report wasn't evaluated for some contracts because `eth-gas-reporter` uses an outdated ABI parser, which doesn't support some features (custom errors in our case) of Solidity 0.8.x. The issue was solved by bumping the version of the dependency `@ethersproject/abi` in the `eth-gas-reporter` package: https://github.com/Psirex/eth-gas-reporter/commit/8ae255a8f94bcdb9bd9d669d2c1dbb012c761369


TheDZhon (approved):
Thank you, it works like a charm.

PR #527, opened 2023-02-01
non zero withdrawal vault

bulbozaur:
Non zero withrawal vault


contracts/0.4.24/Lido.sol
```diff
@@ -511,8 +519,17 @@ contract Lido is StETHPermit, AragonApp {
     * @dev withdrawal vault address is encoded as a last 160 bits of withdrawal credentials type 0x01
     * @return address of the vault or address(0) if the vault is not set
     */
-    function getWithdrawalVault() external view returns (address) {
-        return _getWithdrawalVault();
+    function getWithdrawalVault() public view returns (address) {
+        return WITHDRAWAL_VAULT_POSITION.getStorageAddress();
+    }
+
+    function setWithdrawalVault(address _withdrawalVault) external {
```
TheDZhon, line 526: maybe set all the contracts with `setProtocolContracts` to save bytecode size?

bulbozaur: then there will be conflict in initialize where _initialize_v2 and _setProtocolContracts called

TheDZhon: Didn't get it (we can read old contracts and pass them again for initialization purposes or am I missing smth?)

folkyatina: I see. We can likely merge  initialized_v2 and setProtocolContracts together. Just pass already set contracts in finalize_v2.

bulbozaur: or just let double writing int initialize method, as far as it won't be called never on already deployed contract, but will be replaced with next upgrade anyway. removing of 2 setters (setWithdrawalVault and setWithdrawaQueue) and concentrating it in setProtocolContracts saves 0.25kb of bytecode as well)

PR #529, opened 2023-02-02
Oracle report sanity checks

Psirex:


TheDZhon:
https://hackmd.io/@lido/SyluPxNji?type=view

For the reference

PR #530, opened 2023-02-02
Versioned.sol inheritance

bulbozaur:


contracts/0.4.24/utils/Versioned.sol
```diff
@@ -0,0 +1,46 @@
...
+    /// @notice Returns the current contract version.
+    function getContractVersion() public view returns (uint256) {
+        return CONTRACT_VERSION_POSITION.getStorageUint256();
+    }
+
+    function _checkContractVersion(uint256 version) internal view {
+        require(version == getContractVersion(), "UNEXPECTED_CONTRACT_VERSION");
+    }
+
+    function _writeContractVersion(uint256 version) internal {
```
arwer13, line 42: Maybe `_setContractVersion`? Because we use set but write everywhere else

arwer13 (changes requested):

contracts/0.4.24/utils/Versioned.sol
TheDZhon: agree

bulbozaur: fixed

TheDZhon (approved):
Awesome, thanks! :rocket:

PR #534, opened 2023-02-02
Use overrides for eth-gas-reporter instead of a forked repo

Psirex:
Pin the inner `@ethersproject/abi` dependency of the `eth-gas-reporter` lib using the [`package.json overrides`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) functionality


TheDZhon (approved):
Thank you, it's a pure gem :gem:

PR #539, opened 2023-02-03
Feat: LidoLocator behind Proxy

mymphe:
This PR includes a new contract: Proxied LidoLocator. All addresses are immutable to avoid storage-read operations for gas efficiency. Whenever Locator needs changing, simply upgrade the implementation.


contracts/0.8.9/LidoLocator.sol
```diff
@@ -0,0 +1,173 @@
...
+     * [3] ELRewardsVault
+     * [4] Oracle;
+     * [5] SafetyNetsRegistry;
+     * [6] SelfOwnedStETHBurner;
+     * [7] StakingRouter;
+     * [8] Treasury;
+     * [9] WithdrawalQueue;
+     * [10] WithdrawalVault;
+     */
+    constructor(address[] memory _addresses) {
```
TheDZhon, line 48: maybe pass a struct for initialization?

contracts/0.8.9/LidoLocator.sol
```diff
@@ -0,0 +1,173 @@
...
+ * @title LidoLocator
+ * @author mymphe
+ * @notice Service Locator of Lido
+ */
+contract LidoLocator is ILidoLocator {
+    error ErrorIncorrectLength();
+    error ErrorZeroAddress();
+
+    address internal immutable lido;
+    address internal immutable compositePostRebaseBeaconReceiver;
```
TheDZhon, line 20: maybe `UPPER_CASE_WITH_UNDERLINES` here and there?

contracts/0.8.9/LidoLocator.sol
```diff
@@ -0,0 +1,173 @@
...
+
+/* See contracts/COMPILERS.md */
+pragma solidity 0.8.9;
+
+import {ILidoLocator} from "../common/interfaces/ILidoLocator.sol";
+
+/**
+ * @title LidoLocator
+ * @author mymphe
+ * @notice Service Locator of Lido
```
TheDZhon, line 13: should add a notice about immutable storage and its purpose

contracts/0.8.9/LidoLocator.sol
```diff
@@ -0,0 +1,173 @@
...
+        if (_addresses[10] == address(0)) revert ErrorZeroAddress();
+        withdrawalVault = _addresses[10];
+
+    }
+
+    /**
+     * @notice get the address of the Lido contract
+     * @return address of the Lido contract
+     */
+    function getLido() external view returns (address) {
```
TheDZhon, line 90: can be pure here and there

contracts/0.8.9/test_helpers/LidoLocatorUpdatedMock.sol
```diff
@@ -0,0 +1,137 @@
...
+
+/* See contracts/COMPILERS.md */
+pragma solidity 0.8.9;
+
+
+/**
+ * @title LidoLocatorUpdatedMock
+ * @notice an updated impl of LidoLocator with a different interface
+ */
+contract LidoLocatorUpdatedMock {
```
TheDZhon, line 13: can't get a reason for this mock (it equals to the real contract)

contracts/common/interfaces/ILidoLocator.sol
```diff
@@ -0,0 +1,21 @@
...
+// See contracts/COMPILERS.md
+// solhint-disable-next-line
+pragma solidity 0.4.24||0.8.9;
+
+interface ILidoLocator {
+    function getLido() external view returns(address);
+    function getDepositSecurityModule() external view returns (address);
+    function getELRewardsVault() external view returns (address);
+    function getOracle() external view returns (address);
+    function getCompositePostRebaseBeaconReceiver() external view returns (address);
```
TheDZhon, line 14: don't need this one anymore

contracts/common/interfaces/ILidoLocator.sol
```diff
@@ -0,0 +1,21 @@
...
+
+// See contracts/COMPILERS.md
+// solhint-disable-next-line
+pragma solidity 0.4.24||0.8.9;
+
+interface ILidoLocator {
+    function getLido() external view returns(address);
+    function getDepositSecurityModule() external view returns (address);
+    function getELRewardsVault() external view returns (address);
+    function getOracle() external view returns (address);
```
TheDZhon, line 13: there will be two oracles (accounting and validator exit bus, see contracts/0.8.9/oracle)

TheDZhon (changes requested):
Thank you for the outstanding speed of delivery :rocket: 
Left some suggestions

contracts/common/interfaces/ILidoLocator.sol
TheDZhon: it's better to disambiguate them and make two separate getters

contracts/0.8.9/LidoLocator.sol
folkyatina: I'd rather stick with public immutables but in camel case.

folkyatina: And without getters.

contracts/0.8.9/LidoLocator.sol
```diff
@@ -0,0 +1,79 @@
...
+    address public immutable validatorExitBus;
+    address public immutable withdrawalQueue;
+    address public immutable withdrawalVault;
+
+    /**
+     * @notice declare service locations
+     * @dev accepts a struct to avoid the "stack-too-deep" error
+     * @param _config struct of addresses
+     */
+    constructor(Config memory _config) {
```
skozin, line 52: `calldata` would be cheaper

skozin: Btw, all checks below can be replaced by a universal one that doesn't depend on the number of addresses in the struct:

```solidity
constructor(Config calldata _config) {
    if (!_checkCalldataNonZero()) revert ErrorZeroAddress();
    // make all assignments
}

function _checkCalldataNonZero() internal pure returns (bool allNonZero) {
    assembly {
        allNonZero := 1
        // skip selector
        let offset := 4
        let maxOffset := calldatasize()
        for {} lt(offset, maxOffset) {} {
            if eq(calldataload(offset), 0) {
                allNonZero := 0
                break
            }
            offset := add(offset, 32)
        }
    }
}
```

skozin: Ah, my bad, it's a constructor :(

skozin: No `calldata` here unfortunately

skozin: Btw, how about this?

```solidity
constructor(Config memory _config) {
    accountingOracle = _assertNonZero(_config.accountingOracle);
    // ...
}

function _assertNonZero(address _addr) internal pure {
    if (_addr == address(0)) revert ErrorZeroAddress();
}
```

It would still avoid code duplication and the possibility of forgotten checks.

mymphe: Good idea. Thank you!

contracts/common/interfaces/ILidoLocator.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-FileCopyrightText: 2023 Lido <info@lido.fi>
+
+// SPDX-License-Identifier: GPL-3.0
+
+// See contracts/COMPILERS.md
+// solhint-disable-next-line
+pragma solidity 0.4.24||0.8.9;
```
TheDZhon, line 7: @arwer13 is proposing another approach here, hmm
#543

TheDZhon (approved):
LGTM :+1: 
Though please have a look at the proposed improvements

contracts/common/interfaces/ILidoLocator.sol
```diff
@@ -0,0 +1,30 @@
...
+    function lido() external view returns(address);
+    function safetyNetsRegistry() external view returns(address);
+    function selfOwnedStEthBurner() external view returns(address);
+    function stakingRouter() external view returns(address);
+    function treasury() external view returns(address);
+    function validatorExitBus() external view returns(address);
+    function withdrawalQueue() external view returns(address);
+    function withdrawalVault() external view returns(address);
+    function coreComponents() external view returns(
+        address,
```
TheDZhon, line 23: we can document returns in the interface file I believe

PR #561, opened 2023-02-06
Lido: restore getFee and getFeeDistributions

arwer13:
Sadly, restoring the functions even in this optimised manner eats 🐘 595 bytes  of Lido contract bytecode size.

‼️ Actually, I'm less sure now that restoring the functions which might return now incorrect values due to precision loss is better for integrations.


TheDZhon (approved):
Thank you! :+1:

PR #562, opened 2023-02-06
Oracle: support "fast lane" member subset

skozin:
The fast lane subset is a subset consisting of `quorum` oracles that changes on each frame. Fast lane members can, and are expected to, submit a report during the first part of the frame. Non-fast-lane members are only allowed to submit a report after the "fast-lane" part of the frame passes.

This is done to encourage each oracle from the full set to participate in reporting on a regular basis, and identify any malfunctioning members.

The subset selection is implemented as a sliding window of the `quorum` width over member indices (mod total members). The window advances by one index each reporting frame.


skozin:
(wrong branch)

PR #563, opened 2023-02-06
Oracle: support "fast lane" member subset

skozin:
The fast lane subset is a subset consisting of `quorum` oracles that changes on each frame. Fast lane members can, and are expected to, submit a report during the first part of the frame. Non-fast-lane members are only allowed to submit a report after the "fast-lane" part of the frame passes.

This is done to encourage each oracle from the full set to participate in reporting on a regular basis, and identify any malfunctioning members.

The subset selection is implemented as a sliding window of the `quorum` width over member indices (mod total members). The window advances by one index each reporting frame.

### DIFF
--- a/contracts/0.8.9/oracle/HashConsensus.sol
+++ b/contracts/0.8.9/oracle/HashConsensus.sol
@@ -107,8 +107,6 @@ contract HashConsensus is AccessControlEnumerable {
     }
 
     struct MemberState {
-        // address of the oracle member
-        address addr;
         // the last reference slot a report from this member was received for
         uint64 lastReportRefSlot;
         // the variant index of the last report from this member
@@ -154,8 +152,11 @@ contract HashConsensus is AccessControlEnumerable {
     /// @dev Reporting frame configuration
     FrameConfig internal _frameConfig;
 
-    /// @dev Oracle committee members array
-    MemberState[] internal _members;
+    /// @dev Oracle committee members states array
+    MemberState[] internal _memberStates;
+
+    /// @dev Oracle committee members' addresses array
+    address[] internal _memberAddresses;
 
     /// @dev Mapping from an oracle committee member address to the 1-based index in the
     /// members array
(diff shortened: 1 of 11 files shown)


contracts/0.8.9/oracle/HashConsensus.sol
```diff
@@ -489,35 +583,88 @@ contract HashConsensus is AccessControlEnumerable {
...
+        pastEndIndex = startIndex + _quorum;
+    }
+
+    /// @dev Tests whether the member with the given `index` is in the fast lane subset for the
+    /// given reporting `frameIndex`.
+    ///
+    function _isFastLaneMember(uint256 index, uint256 frameIndex) internal view returns (bool) {
+        uint256 totalMembers = _memberStates.length;
+        (uint256 flLeft, uint256 flPastRight) = _getFastLaneSubset(frameIndex, totalMembers);
+        return Math.pointInHalfOpenIntervalModN(index, flLeft, flPastRight, totalMembers);
```
avsetsin, line 639: If `quorum == totalMembers`, then we will always have `false` here

contracts/0.8.9/oracle/HashConsensus.sol
```diff
@@ -489,35 +583,88 @@ contract HashConsensus is AccessControlEnumerable {
...
-            --_reportVariants[member.lastReportVariantIndex].support;
+            --_reportVariants[memberState.lastReportVariantIndex].support;
         }
 
         _setQuorumAndCheckConsensus(quorum, newTotalMembers);
     }
 
+    function _setFastLaneConfig(uint256 fastLaneLengthSlots) internal {
+        if (fastLaneLengthSlots != _frameConfig.fastLaneLengthSlots) {
+            _frameConfig.fastLaneLengthSlots = fastLaneLengthSlots.toUint64();
```
avsetsin, line 619: Consider checking that `fastLaneLengthSlots < epochsPerFrame`

PR #570, opened 2023-02-07
Accounting oracle gas optimization

skozin:
* Accounting oracle: optimize gas consumption by calculating offchain and passing the max number of extra data items with unique `nodeOpId` per (`itemType`, `moduleId`) pair.
* Lib: simplify `ResizableArray` by removing the growth support that's not needed anymore.


ujenjt:
LGTM

PR #572, opened 2023-02-07
Improve compatibility of the memory allocation helper

skozin:
Align free memory pointer to 32 bytes when allocating memory, as it's currently done by the Solidity compiler (and was done by the `BytesLib` dependency that we're not using anymore).

* [x] Tests


contracts/common/lib/MemUtils.sol
```diff
@@ -15,7 +15,9 @@ library MemUtils {
         assembly {
             result := mload(0x40)
             mstore(result, _len)
-            mstore(0x40, add(add(result, _len), 32))
+            let freeMemPtr := add(add(result, 32), _len)
+            // align free mem ptr to 32 bytes as the compiler does now
+            mstore(0x40, and(add(freeMemPtr, 31), not(31)))
```
TheDZhon, line 20: :+1: 

Got a flashback:
```cpp
#ifndef align_up
#define align_up(num, align) \
    (((num) + ((align) - 1)) & ~((align) - 1))
#endif
```

skozin: love the macro parentheses craziness

contracts/common/test_helpers/Assertions.sol
```diff
@@ -0,0 +1,214 @@
+// SPDX-FileCopyrightText: 2023 Lido <info@lido.fi>
+// SPDX-License-Identifier: GPL-3.0
+
+/* See contracts/COMPILERS.md */
+pragma solidity 0.8.9;
```
TheDZhon, line 5: not sure it should be placed in the `common` though

TheDZhon (approved):
:heart:

contracts/common/test_helpers/Assertions.sol
skozin: yeah the tests are using 0.8.9 and that's the problem. however, the `MemUtils` itself supports 0.4.24

skozin: not sure what to do here. should we move the tests to `0.8.9` given that the tested code is in `common`?

skozin: writing assertions supporting 0.4.24 would be a pain

TheDZhon: I guess we can proceed as is, too minorish for me

PR #579, opened 2023-02-08
Renaming and LidoOracle initialize fix

arwer13:


contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -288,13 +292,13 @@ contract LidoOracle is AragonApp {
      * @notice A function to finalize upgrade v3 -> v4 (the compat-only deprecated impl).
      * Can be called only once.
      */
-    function finalizeUpgrade_v4(address _newOracle) external {
+    function finalizeUpgrade_v4(address _newOracle, address _newOracleConsensusContract) external {
```
skozin, line 295: Here, `_newOracle` must already be initialized. The accounting oracle upgrade should be performed as follows:

1. `HashConsensus` is deployed and configured with the same members as the legacy contract.
2. `AccountingOracle` is deployed and initialized, being provided the `LidoOracle` proxy address (that must still have the v3 impl) and the `HashConsensus` address.
3. `LidoOracle` proxy is upgraded to the compat impl v4.

See: https://github.com/lidofinance/lido-dao/blob/387c94a/contracts/0.8.9/oracle/AccountingOracle.sol#L427-L440

PR #580, opened 2023-02-09
test: hashConsensus setQuorum tests

Jeday:
Tests for setQuorum for basic functionality and its effect during discussion:

- [x] setQuorum calls are reflected in getQuorum and events
- [x] setQuorum changes the effective quorum in getConsensusState
- [x] setQuorum triggers consensus on decrease
- [x] setQuorum doesn’t re-trigger the same consensus
- [x] setQuorum doesn’t re-trigger consensus if hash is already being processed 

image


TheDZhon (approved):
GTG :+1:

TheDZhon (approved):

PR #583, opened 2023-02-09
LidoOracle, AccountingOracle: deploy related initialize() fixes

arwer13:


contracts/0.8.9/oracle/AccountingOracle.sol
```diff
@@ -522,11 +548,13 @@ contract AccountingOracle is BaseOracle {
             );
         }
 
-        ILegacyOracle(LEGACY_ORACLE_POSITION.getStorageAddress()).handleConsensusLayerReport(
-            data.refSlot,
-            data.clBalanceGwei * 1e9,
-            data.numValidators
-        );
+        if (LEGACY_ORACLE != address(0)) {
```
skozin, line 551: how can it be zero? I'd prefer not to have this check if it's not 100% necessary

arwer13: I'm not sure too actually. It adds complexity due to the requirement to have LegacyOracle deployed and initialized in tests and deployment from scratch. And gives kind of feeling that `LidoOracle.handleConsensusLayerReport` is kind of tested. Although it get tested indeed only during tests after upgrade.

But I guess it's likely to be cheaper to forbid zero address than to discuss it.

contracts/0.4.24/oracle/LidoOracle.sol
```diff
@@ -266,19 +268,31 @@ contract LidoOracle is AragonApp {
...
+     * @param _lidoLocator Address of the Lido Locator contract.
+     * @param _accountingOracleConsensusContract Address of consensus contract of the new accounting oracle contract.
+     * @param _lastCompletedEpochId Last completed epoch id position. Might want to set non-zero because
+     *        on initialization the new oracle checks its initial epoch against last completed epoch of legacy oracle
      */
-    function initialize(address _lido, address _newOracle) external onlyInit {
+    function initialize(
+        address _lidoLocator,
+        address _accountingOracleConsensusContract,
+        uint256 _lastCompletedEpochId
```
ujenjt, line 279: let's remove `_lastCompletedEpochId `

PR #589, opened 2023-02-09
More efficient accounting oracle

skozin:
Avoids non-trivial processing of exited/stuck validators report data in the accounting oracle by using `bytes` packing in the Staking Router's methods that receive the data from the oracle.

Gets rid of the hardly-explainable and non-semantical `extraDataMaxNodeOpsCountByModule` report field. `ResizableArray` is not needed anymore.

Should be significantly more gas-efficient than the current impl.

PR #591, opened 2023-02-10
AccountingOracle: move data boundaries to OracleReportSanityChecker

arwer13:
Also rename `maxExitedValidatorsPerDay` to `maxExitedValidatorsRatePerDay `


contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -273,6 +291,28 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+    function setMaxExitedValidatorsRatePerDay(uint256 _maxExitedValidatorsRatePerDay)
+        external
+        onlyRole(MAX_EXITED_VALIDATORS_RATE_PER_DAY_ROLE)
+    {
+        LimitsList memory limitsList = _limits.unpack();
+        limitsList.maxExitedValidatorsRatePerDay = _maxExitedValidatorsRatePerDay;
+        _updateLimits(limitsList);
+    }
+
+    /// @notice Sets the new value for the maxValidatorExitRequestsPerReport
```
skozin, line 305: `maxValidatorExitRequestsPerReport`  => `maxAccountingExtraDataListItemsCount`

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -72,9 +78,11 @@ struct LimitsListPacked {
     uint16 maxValidatorExitRequestsPerReport;
     uint64 requestTimestampMargin;
     uint64 maxPositiveTokenRebase;
+    uint64 maxExitedValidatorsRatePerDay;
```
TheDZhon, line 81: maybe extra because we already have a churn limit (see above)

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -72,9 +78,11 @@ struct LimitsListPacked {
...
+    uint64 maxAccountingExtraDataListItemsCount;
```
TheDZhon, line 82: can be smaller (the idea is to fit the structure into the single slot)

skozin:
This will conflict with #589 if the latter is to be merged. If we want to merge #589, I'd postpone merging this PR until #589 is merged since changes here are incompatible with it.

skozin:
E.g. #589 needs the value of `maxAccountingExtraDataListItemsCount`, not only the function for checking it.

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
skozin: if #589 is merged, the type of `maxAccountingExtraDataListItemsCount` can be set as low as `uint24`

skozin: churn limit is 1800 per day (?), thus `maxExitedValidatorsRatePerDay` can be a `uint16` (needs checking)

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
skozin: it's checked as a part of `Lido.handleOracleReport`, right?

TheDZhon: nope, I only appearing validators

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
TheDZhon: churn limit is 60 validators at max (roughly all Eth TVL is staked), i.e., 13500 per day
perfectly feats into `uint16`

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
skozin: then I'd keep it

PR #600, opened 2023-02-10
fix: remove extra pause check from `_mintShares`

TheDZhon:


arwer13:
I'd like to add comment about the invariant at `isStakingPaused`. Or event document the invariant somewhere else if we have a place where the protocol invariants are kept.

PR #601, opened 2023-02-11
fix: handle report resubmit case

Jeday:
## Scenario

* initial report is submitted at `refSlot1`
* next report is submitted at `refSlot2`,  initial report is missed, warning event fired
  * WarnProcessingMissed(`refSlot1`) 
* next report is re-agreed by consensus and resubmitted with new data
  * expected: WarnProcessingMissed(`refSlot1`) 
  * actual:  WarnProcessingMissed(`refSlot2`)

## Solution

At resubmit we lack information about slot missed at prev submitting. So proposed solution is to prevent fire of `WarnProcessingMissed` event at resubmit. What will be achieved:

- Upholding 1 missed report - 1 event condition
- WarnProcessingMissed argument can be trusted to contain actual missed slot 

### Testing notes

Corresponding tests are located in the branh [feature/shapella-upgrade-tests-base-oracle](https://github.com/lidofinance/lido-dao/tree/feature/shapella-upgrade-tests-base-oracle)

### DIFF
--- a/contracts/0.8.9/oracle/BaseOracle.sol
+++ b/contracts/0.8.9/oracle/BaseOracle.sol
@@ -173,7 +173,7 @@ abstract contract BaseOracle is IReportAsyncProcessor, AccessControlEnumerable, 
             revert RefSlotMustBeGreaterThanProcessingOne(refSlot, prevProcessingRefSlot);
         }
 
-        if (prevProcessingRefSlot != prevSubmittedRefSlot) {
+        if (refSlot != prevSubmittedRefSlot && prevProcessingRefSlot != prevSubmittedRefSlot) {
             emit WarnProcessingMissed(prevSubmittedRefSlot);
         }
 


ujenjt:
looks valid, @skozin could you check plz

ujenjt:
also, we don't have an event for slot re-agreement occurrences. Maybe let's add one for monitoring purposes?

TheDZhon (approved):
I think the fix should be applied, though waiting for Sam's feedback

Jeday:
> also, we don't have an event for slot re-agreement occurrences. Maybe let's add one for monitoring purposes?

This would be more relevant in `HashConsensus` contract. I think it can be implemented fairly easily. Though this brings us to question: if consensus can be reached again with different result and report submitted, why can't report be withdrawn if consensus is lost?

skozin:
> also, we don't have an event for slot re-agreement occurrences. Maybe let's add one for monitoring purposes?

good idea, will add

skozin:
I'd say that firing `WarnProcessingMissed(refSlot1)` twice is not the expected behavior, it's also incorrect. And the expected one is not firing the event twice, exactly the one this PR implements.

skozin (approved):

PR #619, opened 2023-02-15
Add support of Foundry tests

Psirex:


TheDZhon (approved):
Wow, this is cool :guitar:

PR #622, opened 2023-02-15
Test: `MAX_NODE_OPERATORS_COUNT` reward distribution gas for NodeOperatorsRegistry

TheDZhon:
Just measure that we can fit into the single block

image


TheDZhon:
Merged into #624

PR #626, opened 2023-02-17
Deposits accounting

Psirex:
Deposit accounting & contracts changes:
- `StakingRouter` now accepts the precise amount of deposits to make. If `StakingRouter` can't make the requested number of deposits, it will revert the transaction.

- `StakingRouter` doesn't return ETH to the `Lido` contract after the deposit, so the `Lido.receiveStakingRouterDepositRemainder()` method was removed.

- `StakingRouter.getStakingModuleMaxDepositsCount()` method signature changed. It now accepts two arguments: `stakingModuleId` and the max amount of ether that might be deposited.
    
- `IStakingModule.obtainDepositData()` now must return the exact number of requested keys. If `StakingModule` doesn't have enough deposit data items, it **MUST** revert
    
- `IStakingModule.obtainDepositData()` now doesn't return `depositsCount` as the first item. Only `publicKeys` and `signatures` returned now
   
- Assertion that deposited correct amount of ETH was moved from the `BeaconChainDepositor` contract to `StakingRouter` for clarity

- `Lido._markAsUnbuffered()` method was removed. Now, the buffered ether value is updated in place on deposit, like in other methods.

- Assert for unaccounted ether was removed from `Lido`. Previously such a check was relevant cause the deposits happened in the `Lido` contract itself. In the current implementation, ETH is sent to `StakingRouter`, where deposits happen, which contains assert that exactly the received amount of ETH was deposited.
    
- `Lido._getUnaccountedEther()` method was removed
    
- `Lido.getDepositableEther()` returns a value not multiple to 32 ether
    
- Check that `stakingModuleId` is less than `type(uint24).max` was removed from `Lido.deposit()` cause external interface of staking modules declares id as `uint256`


test/helpers/assert.js
```diff
@@ -44,8 +44,8 @@ chai.util.addMethod(chai.assert, 'notEmits', function (receipt, eventName, args
   this.isUndefined(event, `Expected that event "${eventName}" with args ${args}  wouldn't be emitted, but it was.`)
 })
 
-chai.util.addMethod(chai.assert, 'reverts', async function (receipt, reason) {
-  await assertRevert(receipt, reason)
+chai.util.addMethod(chai.assert, 'reverts', async function (receipt, reason, customErrorArgs) {
+  await assertRevert(receipt, customErrorArgs !== undefined ? `${reason}(${customErrorArgs.join(', ')})` : reason)
```
TheDZhon, line 48: have `assert.revertsWithCustomError()` already :thinking:

TheDZhon (approved):
It's awesome :open_mouth: 

Good to know that it saves a bytecode size also:
image

test/helpers/assert.js
Psirex: Yep, but regular `reverts()` work perfectly with custom errors, so I think we can remove `revertsWithCustomError()` eventually. But the introduced change in the `reverts()` method solves a bit different problem: it allows passing arguments for custom errors as standalone parameters. It might be useful when interpolated error string becomes too long. For example, instead of writing:
```js
await assert.reverts(
    foo.test(),
    `VeryLongErrorMessageWithManyCustomArguments(${firstArgumentWithLongName},${secondArgumentWithLongName},${thirdArgumentWithLongName})`
)
```
might be rewritten in the below form:
```js
await assert.reverts(
    foo.test(),
    'VeryLongErrorMessageWithManyCustomArguments',
    [firstArgumentWithLongName, secondArgumentWithLongName, thirdArgumentWithLongName]
)
```
what improves the reliability of the test (no long strings and horizontal scrolling).

PR #627, opened 2023-02-17
feat: view to get claimable ether for request

folkyatina:


contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -209,7 +209,21 @@ abstract contract WithdrawalQueue is AccessControlEnumerable, PausableUntil, Wit
         uint256 hint;
     }
 
-    /// @notice Claim withdrawals batch once finalized (claimable).
+    /// @notice Return batch of claimable eth amounts that is locked for each request
+    /// @param _claimWithdrawalInputs list of withdrawal request ids and hints to claim
+    function getClaimableEther(ClaimWithdrawalInput[] calldata _claimWithdrawalInputs)
```
TheDZhon, line 214: Just a consideration: integrations would need to repack the args to ClaimWithdrawalInput manually
though, they could have passed two arrays without additional conversions in the middle

contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -209,7 +209,21 @@ abstract contract WithdrawalQueue is AccessControlEnumerable, PausableUntil, Wit
         uint256 hint;
     }
 
-    /// @notice Claim withdrawals batch once finalized (claimable).
+    /// @notice Return batch of claimable eth amounts that is locked for each request
```
TheDZhon, line 212: should note that requests must be in finalized state

contracts/0.8.9/WithdrawalQueueBase.sol
```diff
@@ -424,31 +424,56 @@ abstract contract WithdrawalQueueBase {
...
+    /// @notice Calculates discounted ether value for `_requestId` using a provided `_hint`
+    /// @return request link to request in storage
+    /// @return ethWithDiscount discounted eth for `_requestId`. Returns 0 if request is not claimable
+    function _calculateClaimableEth(uint256 _requestId, uint256 _hint)
+        internal
+        view
+        returns (WithdrawalRequest storage, uint256)
+    {
+        if (_hint == 0) revert InvalidHint(_hint);
+        if (_requestId == 0 || _requestId > getLastRequestId()) revert InvalidRequestId(_requestId);
```
TheDZhon, line 460: looks like you don't check for `_requestId` < `getLastFinalizedRequestId()` if this func called from the `getClaimableEther` method.

This would mean (due to the shortcut below) that `claimableEth` set to zero
Not sure that it's a good UX just to avoid reverting tx

contracts/0.8.9/WithdrawalQueueBase.sol
```diff
@@ -424,31 +424,56 @@ abstract contract WithdrawalQueueBase {
...
+
+        WithdrawalRequest storage request = _getQueue()[_requestId];
+
+        // shortcuts
+        if (_requestId > getLastFinalizedRequestId()) return (request, 0);
+        if (request.claimed) return (request, 0);
```
TheDZhon, line 466: is it really possible or should it be treated like `assert`?

TheDZhon (approved):
Looks good :+1: 
Left some questions regarding UX, maybe you can invest some time to tinker a bit.

contracts/0.8.9/WithdrawalQueue.sol
folkyatina: Reasonable point. Will remove the structure.

contracts/0.8.9/WithdrawalQueue.sol
folkyatina: They should not, actually. It will return 0 as we have no claimable eth in non-final requests.

contracts/0.8.9/WithdrawalQueueBase.sol
folkyatina: It's not clear why getClaimableEth should be reverted if request has no claimable eth. IMHO, it should return 0. It will even be convenient for the user. You can load all request id's to the method without learning its exact status.

contracts/0.8.9/WithdrawalQueueBase.sol
folkyatina: It's definitely possible.

contracts/0.8.9/WithdrawalQueue.sol
TheDZhon: got your point, yeah

contracts/0.8.9/WithdrawalQueueBase.sol
TheDZhon: fair enough

contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -366,4 +406,16 @@ abstract contract WithdrawalQueue is AccessControlEnumerable, PausableUntil, Wit
...
         }
     }
+
+    /// @notice returns claimable ether under the request with _requestId.
+    function _getClaimableEther(uint256 _requestId, uint256 _hint) internal view returns (uint256) {
+        if (_requestId == 0 || _requestId > getLastRequestId()) revert InvalidRequestId(_requestId);
+
+        if (_requestId > getLastFinalizedRequestId()) return 0;
+
+        WithdrawalRequest storage request = _getQueue()[_requestId];
```
TheDZhon, line 416: I guess we can load to `memory` the whole structure without significant gas losses

anyway, we read `claimed` right here (slot 1) and `cumulativeStETH` (slot 0) inside `_calculateClimableEther`
so we may pass only `memory` struct to `_calculateClaimableEther` without the separate `_requestId`

TheDZhon (approved):
Like the new refactored structure.
Left some minor considerations.

contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -184,56 +187,94 @@ abstract contract WithdrawalQueue is AccessControlEnumerable, PausableUntil, Wit
...
-        _emitTransfer(msg.sender, address(0), _requestId);
+    /// @notice Claim a batch of withdrawal requests once finalized (claimable) sending locked ether to the owner
+    /// @param _requestIds array of request ids to claim
+    /// @param _hints checkpoint hint for each id.
+    ///   Can be retrieved with `findCheckpointHints()` or `findCheckpointHintsUnbounded()`
+    /// @dev
+    ///  Reverts if any requestId or hint in arguments are not valid
+    ///  Reverts if any request is not finalized or already claimed
+    ///  Reverts if msg sender is not an owner of the requests
+    function claimWithdrawals(uint256[] calldata _requestIds, uint256[] calldata _hints) external {
```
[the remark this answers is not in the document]
folkyatina: Yep. We can. But I thought about UX safety here. It seems safer to have a default method that claims to the sender and if you really need it you can use an extended version. Making everyone use an extended version by default can increase the probability of sending funds to a wrong address.

contracts/0.8.9/WithdrawalQueue.sol
folkyatina: But we ll need the requestid to check hint validity, I believe, inside `_calculateClaimableEth`

contracts/0.8.9/WithdrawalQueue.sol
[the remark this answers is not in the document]
TheDZhon: I am not against the method, I propose to call another inside

contracts/0.8.9/WithdrawalQueue.sol
TheDZhon: Ah, I see, the structure doesn't contain it, of course

contracts/0.8.9/WithdrawalQueue.sol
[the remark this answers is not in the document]
folkyatina: Oh. Got you. 

I decided that it's a bit more readable as is, because unwinding a loop doesn't really hurt the readability, but splitting the logic into 3 pieces instead of 2 does. 
Also, public method generates a tiny bit of additional bytecode and cost a tiny bit more gas. So, it's not a strong opinion, but I prefer it like this.

PR #628, opened 2023-02-17
refactor: nft renaming

folkyatina:


TheDZhon (approved):
New name is a way more better :+1:

PR #629, opened 2023-02-17
Fix and multiple improvements to OracleReportSanityChecker

arwer13:
- make checks for extraDataListItems and nodeOperatorsPerExtraDataItem depend on two separate limits
- add check for max (due packing) value of a limit in setters
- refactor order of the limits
- add 2 simple tests for the setters


TheDZhon (approved):
Nice one :+1:

PR #630, opened 2023-02-17
Fix: add OracleDaemonConfig to LidoLocator

TheDZhon:
Add a dedicated getter method for the `OracleDaemonConfig` contract.

- [x] Implement the contract
- [x] Update interface
- [x] Check that tests are not broken
- [x] Update tests for the contract itself


arwer13 (approved):
LGTM 👍

PR #631, opened 2023-02-17
Fix for nft balanceOf to be decreased after claim

folkyatina:


contracts/0.8.9/WithdrawalQueueBase.sol
```diff
@@ -442,6 +442,7 @@ abstract contract WithdrawalQueueBase {
         if (_recipient == address(0)) _recipient = request.owner;
 
         request.claimed = true;
+        _getRequestsByOwner()[request.owner].remove(_requestId);
```
TheDZhon, line 445: maybe it worth checking the return status of the `remove` call (the same was proposed in #633)

TheDZhon (approved):
:+1:

contracts/0.8.9/WithdrawalQueueBase.sol
folkyatina: Probably, it is.

folkyatina:
Merged into #632

PR #632, opened 2023-02-17
Fix issues from preliminary audit

folkyatina:


contracts/0.8.9/WithdrawalQueue.sol
```diff
@@ -336,10 +331,9 @@ abstract contract WithdrawalQueue is AccessControlEnumerable, PausableUntil, Wit
         _grantRole(FINALIZE_ROLE, _finalizer);
         _grantRole(BUNKER_MODE_REPORT_ROLE, _bunkerReporter);
 
-        RESUME_SINCE_TIMESTAMP_POSITION.setStorageUint256(PAUSE_INFINITELY); // pause it explicitly
         BUNKER_MODE_SINCE_TIMESTAMP_POSITION.setStorageUint256(BUNKER_MODE_DISABLED_TIMESTAMP);
 
-        emit InitializedV1(_admin, _pauser, _resumer, _finalizer, msg.sender);
+        emit InitializedV1(_admin, _pauser, _resumer, _finalizer, _bunkerReporter);
```
TheDZhon, line 336: should we bypass `grantRole` inside `_initialize` if zero addresses are provided?

TheDZhon (approved):
Good to go :+1: 
Left a Q about `grantRole`

contracts/0.8.9/WithdrawalQueue.sol
folkyatina: Yep. Good catch. Save the blockspace.

TheDZhon (approved):
:+1:

PR #633, opened 2023-02-17
Forbid transfer to themselves for Withdrawal nft

folkyatina:


contracts/0.8.9/WithdrawalQueueERC721.sol
```diff
@@ -191,8 +193,8 @@ contract WithdrawalQueueERC721 is IERC721Metadata, WithdrawalQueue {
         delete _getTokenApprovals()[_requestId];
         request.owner = payable(_to);
 
-        _getRequestsByOwner()[_to].add(_requestId);
         _getRequestsByOwner()[_from].remove(_requestId);
+        _getRequestsByOwner()[_to].add(_requestId);
```
TheDZhon, line 197: as a kind of safety measure, I would suggest checking the returns of `add` and `remove` methods of the OZ `EnumerableSet` here and there, and maybe assert on them

TheDZhon (approved):
The fix is essential :heart: 
Suggested a bit more hardening.

contracts/0.8.9/WithdrawalQueueERC721.sol
folkyatina: I think we can. It will be the third ring of defence, but relatively cheap one.

TheDZhon (approved):
:rocket:

PR #645, opened 2023-02-21
NOR: signing keys lib refactor

krogla:
optimized version of SigningKeys lib


TheDZhon (approved):
Let's roll it out :star_struck:

PR #650, opened 2023-02-24
Feat: fix/share rate simulation

TheDZhon:
- [x] Interface changes
- [x] Implementation
- [x] Don't break existing tests
- [ ] Test coverage to show why we need it
- [ ] Double-check burner test coverage


contracts/0.8.9/oracle/AccountingOracle.sol
```diff
@@ -232,6 +233,10 @@ contract AccountingOracle is BaseOracle {
         /// at the reference slot.
         uint256 elRewardsVaultBalance;
 
+        /// @dev The shares amount requested to burn through Burner as observed
+        /// at the reference slot.
```
skozin, line 237: maybe tell here how to obtain it? it's very impl-specific value anyway so referring the specific fn won't make the docs more fragile imo

contracts/0.8.9/Burner.sol
```diff
@@ -312,9 +311,9 @@ contract Burner is IBurner, AccessControlEnumerable {
             coverSharesBurnRequested -= sharesToBurnNowForCover;
             sharesToBurnNow += sharesToBurnNowForCover;
         }
-        if ((memNonCoverSharesBurnRequested > 0) && (sharesToBurnNow < _sharesToBurnLimit)) {
+        if ((memNonCoverSharesBurnRequested > 0) && (sharesToBurnNow < _sharesToBurn)) {
```
skozin, line 314: don't need inner parentheses here?

skozin (approved):

contracts/0.8.9/Burner.sol
TheDZhon: agree, removed :)
ty!

contracts/0.8.9/oracle/AccountingOracle.sol
TheDZhon: Yeah, added some more explanations :+1:

PR #658, opened 2023-02-27
refactor: remove needless grantRole calls in exit bus initialize()

arwer13:
As long as pauser and resumer roles granting it's gonna be removed for WithdrawalQueue let's remove it here as well.
Also, after writing template upgrade, I favour this option more.


TheDZhon (approved):
:+1:

PR #651, opened 2023-02-24
Fix: withdrawal queue finalization correctness

skozin:
WIP

PR #661, opened 2023-02-28
Fix/upgrade test scenario EL rewards after the merge

DiRaiks:
- un skip, fix/upgrade `execution_layer_rewards_after_the_merge` test
- fix comment in `AccountgOracle` contract

image

### DIFF
--- a/contracts/0.8.9/oracle/AccountingOracle.sol
+++ b/contracts/0.8.9/oracle/AccountingOracle.sol
@@ -243,8 +243,8 @@ contract AccountingOracle is BaseOracle {
 
         /// @dev The share/ETH rate with the 10^27 precision (i.e. the price of one stETH share
         /// in ETH where one ETH is denominated as 10^27) used for finalizing withdrawal requests
-        /// up to (and including) the one passed in the lastWithdrawalRequestIdToFinalize field.
-        /// Must be set to zero if lastWithdrawalRequestIdToFinalize is zero.
+        /// up to (and including) the one passed in the lastFinalizableWithdrawalRequestId field.
+        /// Must be set to zero if lastFinalizableWithdrawalRequestId is zero.
         uint256 simulatedShareRate;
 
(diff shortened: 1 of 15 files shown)


contracts/0.8.9/oracle/AccountingOracle.sol
```diff
@@ -243,8 +243,8 @@ contract AccountingOracle is BaseOracle {
 
         /// @dev The share/ETH rate with the 10^27 precision (i.e. the price of one stETH share
         /// in ETH where one ETH is denominated as 10^27) used for finalizing withdrawal requests
-        /// up to (and including) the one passed in the lastWithdrawalRequestIdToFinalize field.
-        /// Must be set to zero if lastWithdrawalRequestIdToFinalize is zero.
+        /// up to (and including) the one passed in the lastFinalizableWithdrawalRequestId field.
```
TheDZhon, line 246: this is a subject of change after #651 is merged
though, it's better to have in diff till that

test/scenario/execution_layer_rewards_after_the_merge.test.js
```diff
@@ -32,8 +33,35 @@ const makeAccountingReport = ({ refSlot, numValidators, clBalanceGwei, elRewards
   extraDataHash: ZERO_HASH,
   extraDataItemsCount: 0,
 })
-
-contract.skip('Lido: merge acceptance', (addresses) => {
+function calcReportDataHash(reportItems) {
```
TheDZhon, line 36: can we import these helpers somehow?
It would be less painful to update the func signature in the future if needed

TheDZhon (approved):
:rocket: Thank you so much for ressurecting these tests.
Please consider the suggestion below regarding common helpers for oracle data encoding.

test/scenario/execution_layer_rewards_after_the_merge.test.js
DiRaiks: done

PR #662, opened 2023-02-28
Hash consensus: allow updating a future initial epoch

skozin:
Allows updating the initial epoch of a `HashConsensus` instance while the current initial epoch is still in the future. This simplifies the protocol upgrade process by allowing to pre-deploy the consensus contract before the exact time of the protocol upgrade is known.


arwer13:
Thank you!

We also have another problem. `BaseOracle._initialize` calls `_setConsensusContract` which checks `refSlot < lastProcessingRefSlot` where `refSlot` is taken from `IConsensusContract(addr).getCurrentFrame()` which calls `_computeFrameIndex` which requires `epoch >= config.initialEpoch` what is  false at the moment of AccountingOracle initialization.

arwer13:
I also not sure if it is a good idea to check on AccountingOracle migration strictly `initialEpoch == legacyProcessedEpoch + epochsPerFram`. There might be multiple scenarios when its wrong. For example (the day counts are approx):
1. Aragon Voting started at hardfork-4days. EVMScript has `updateInitialEpoch(hardfork +1d)`
2. For any reason at hardfork-1d LidoOracle committee didn't report
3. Voting passed, it's time to enact because the hardfork is comming
4. Enactment reverts due to the check `initialEpoch == legacyProcessedEpoch + epochsPerFram`

skozin:
@arwer13 See the latest commit with the fix for the revert you've mentioned.

As for the initial epoch check, the assumed process of deployment is the following:

1. `HashConsensus`, `AccountingOracle`, and accounting oracle's proxy are deployed before the upgrade voting is started. The initial epoch of `HashConsensus` is far in the future, thus the consensus is guaranteed to be inoperable.
2. The upgrade voting is started.
3. The upgrade voting is passed.
4. As part of the upgrade voting execution, the following happens (assuming we have the migrator smart contract where this happens):
    - The new oracle's initial epoch is calculated as the last legacy oracle's epoch plus the frame size. The new oracle's initial epoch might be in the past.
    - The initial epoch of `HashConsensus` is updated to the calculated value by calling `updateInitialEpoch`.
    - Legacy oracle is disabled (upgraded to the compact impl).
    - The `AccountingOracle` is initialized, passing the `HashConsensus` address.

Thus, if the initial epoch is calculated correctly in the migrator contract, the check should pass regardless of the vote timing.

arwer13 (approved):
👍

TheDZhon (approved):
:rocket:

PR #673, opened 2023-03-06
No extrema checking during withdrawal finalization

folkyatina:


TheDZhon:
Looks like we are on track again.

image

PR #674, opened 2023-03-07
Withdrawal queue: trust oracle in ETH calc when finalizing withdrawals

skozin:
An attempt at reducing the code complexity, to be compared with #673.


TheDZhon:
Parking this one in favor of #673 

image

PR #680, opened 2023-03-10
ERC-1271 contract signature verification for stETH

skozin:
Allows stETH token to verify permit signatures issued by contracts. This is required to support account abstraction, among other things.

See:

* https://eips.ethereum.org/EIPS/eip-1271
* https://eip1271.io
* [Uniswap v3 permit impl](https://github.com/Uniswap/v3-periphery/blob/9ca9575/contracts/base/ERC721Permit.sol#L77)
* [Balancer v2 permit impl](https://github.com/balancer/balancer-v2-monorepo/blob/f423511/pkg/solidity-utils/contracts/openzeppelin/ERC20Permit.sol#L51)


contracts/common/lib/SignatureUtils.sol
```diff
@@ -0,0 +1,65 @@
...
+                let success := staticcall(gas(), signer, add(data, 32), mload(data), outDataOffset, 32)
+                if eq(success, 1) {
+                    retval := mload(outDataOffset)
+                }
+            }
+            return retval == ERC1271_IS_VALID_SIGNATURE_SELECTOR;
+        } else {
+            return ECDSA.recover(msgHash, v, r, s) == signer;
+        }
+    }
```
TheDZhon, line 57: Looked briefly into the Aragon impl:
https://github.com/aragon/aragon-apps/blob/master/apps/agent/contracts/SignatureValidator.sol#L113-L139

Looks like they are limiting the forwarded gas while the standard [discourages](https://eips.ethereum.org/EIPS/eip-1271#security-considerations) this approach:

> Since there are no gas-limit expected for calling the isValidSignature() function, it is possible that some implementation will consume a large amount of gas. It is therefore important to not hardcode an amount of gas sent when calling this method on an external contract as it could prevent the validation of certain signatures.

TheDZhon (approved):
luscious :+1:

PR #690, opened 2023-03-16
Fix: Lido.handleOracleReport reorder burn and fee distribution

TheDZhon:
The idea is to harden the accounting processing
Shares should be burnt BEFORE the fee distribution.


folkyatina (approved):
👍

PR #692, opened 2023-03-17
Accounting oracle: remove an erroneous and excess check

skozin:
The requirement for the number of exited validators to not decrease is already checked in the staking router, and it's compared per module which is a more strict check.

TODO:

* [x] Tests for the return value of `StakingRouter.updateExitedValidatorsCountByStakingModule`


test/0.8.9/staking-router/report-exited-keys.test.js
```diff
@@ -26,6 +26,9 @@ contract('StakingRouter', ([admin, depositor]) => {
     })
 
     router = deployed.stakingRouter
+    router.updateExitedValidatorsCountByStakingModuleWithResult = contractMethodWithResult(
```
skozin, line 29: Modifying function names makes reading tests harder since it's not clear which underlying function is called.

Can we do smth like this instead?

```js
// in utils

function addResultHandler(method) {
  method.transactAndGetResult = async (...args) => {
    const result = await method.call(...args)
    await method(...args)
    return result
  }
}

// in the before handler

addResultHandler(router.updateExitedValidatorsCountByStakingModule)

// in a test

const result = await router
  .updateExitedValidatorsCountByStakingModule
  .transactAndGetResult(...args)
```

This way, it's clear which function is invoked, and that it's involked in some magical way.

contracts/0.8.9/StakingRouter.sol
```diff
@@ -299,27 +299,72 @@ contract StakingRouter is AccessControlEnumerable, BeaconChainDepositor, Version
...
+    ///    `StakingRouter.onValidatorsCountsByNodeOperatorReportingFinished` which, in turn, calls
+    ///    `IStakingModule.onExitedAndStuckValidatorsCountsUpdated` on all modules.
+    ///
     function updateExitedValidatorsCountByStakingModule(
         uint256[] calldata _stakingModuleIds,
         uint256[] calldata _exitedValidatorsCounts
     )
         external
         onlyRole(REPORT_EXITED_VALIDATORS_ROLE)
+        returns (uint256)
```
Psirex, line 349: Super minor: `newlyExitedValidatorsCount` might be declared here, in the function declaration, then lines 355 and 386 might be removed

contracts/0.8.9/StakingRouter.sol
```diff
@@ -299,27 +299,72 @@ contract StakingRouter is AccessControlEnumerable, BeaconChainDepositor, Version
...
-            stakingModuleId = _stakingModuleIds[i];
+            uint256 stakingModuleId = _stakingModuleIds[i];
             StakingModule storage stakingModule = _getStakingModuleById(stakingModuleId);
 
             uint256 prevReportedExitedValidatorsCount = stakingModule.exitedValidatorsCount;
             if (_exitedValidatorsCounts[i] < prevReportedExitedValidatorsCount) {
                 revert ExitedValidatorsCountCannotDecrease();
             }
 
+            newlyExitedValidatorsCount += _exitedValidatorsCounts[i] - prevReportedExitedValidatorsCount;
```
Psirex, line 366: Super minor: subtraction might be wrapped in `unchecked {}` for gas savings.

Psirex (commented):
LGTM. Especially liked the detailed comment about the validators' update process 👍🏻👍🏻👍🏻

contracts/0.8.9/StakingRouter.sol
Psirex: Agree 👍🏻

TheDZhon:
Please consider fixing the conflicts with the target branch :pray:

TheDZhon (approved):
Neat :+1:

PR #698, opened 2023-03-21
Fix gas estimation for methods with try/catch in StakingRouter

Psirex:


contracts/0.8.9/StakingRouter.sol
```diff
@@ -289,6 +290,12 @@ contract StakingRouter is AccessControlEnumerable, BeaconChainDepositor, Version
                 address moduleAddr = _getStakingModuleById(_stakingModuleIds[i]).stakingModuleAddress;
                 try IStakingModule(moduleAddr).onRewardsMinted(_totalShares[i]) {}
                 catch (bytes memory lowLevelRevertData) {
+                    /// @dev This check is required to prevent incorrect gas estimation of the method.
+                    ///      Without it, Ethereum nodes that use binary search for gas estimation may
+                    ///      return an invalid value when the onRewardsMinted() reverts because of the
+                    ///      "out of gas" error. Here we assume that the onRewardsMinted() method doesn't
+                    ///      have reverts with empty error data except "out of gas".
+                    if (lowLevelRevertData.length == 0) revert ModuleOutOfGas();
```
TheDZhon, line 298: Shall we rename the `ModuleOutOfGas` error to something like `UnrecoverableModuleError`?

Because the real reasons might be different while we still don't want gracefully handle them in the modules' code, I guess.

TheDZhon (approved):
:guitar: let's roll it out

PR #699, opened 2023-03-21
Withdrawal Queue Gas Tests

Jeday:
Added gas test for batch operations on WQ.
### Usage notes

Tests are skipped when not run with `test:gas`. Run time is pretty long and transactions pump block price and may interfere with other tests. Fiddle with `MAX_BATCH_SIZE` and `batchIncrement` to generate different results and insights.

### Findings

- `requestWithdrawals` fails on >~270 batch size. Could be hardhat limitation.
- `finalize` spikes in gas usage on first time.
-  `claimWithdrawals`overestimates gas, convergence at ~21% less gas used

### Results
image
image

### DIFF
--- a/test/0.8.9/withdrawal-queue-gas.test.js
+++ b/test/0.8.9/withdrawal-queue-gas.test.js
@@ -75,6 +75,8 @@ contract('WithdrawalQueue', ([owner, user]) => {
         user,
         {
           from: user,
+          // smap of large transactions causes local network baseFee rise in next blocks
+          // increase gasPrice a bit on every tx to make sure execute
           gasPrice: gasPrice++,
           gasLimit: 1000000000,
         },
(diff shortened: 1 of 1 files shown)


test/0.8.9/withdrawal-queue-gas.test.js
```diff
@@ -0,0 +1,192 @@
...
+      console.table(results)
+    })
+
+    itParam('batch size ${value}', batchSizes, async (batch_size) => {
+      const args = [
+        Array(batch_size).fill(REQ_AMOUNT),
+        user,
+        {
+          from: user,
+          gasPrice: gasPrice++,
```
TheDZhon, line 78: Why do we use `gasPrice` as a simple increment counter?

Jeday: gas fee rise with these many txs, so it's a simple hack

TheDZhon (approved):
LGTM, thanks!

PR #708, opened 2023-03-27
Fix: discard report processing on lost consensus

skozin:
On consensus loss, discard the report if its processing hasn't started yet. This increases the chance that the DAO can react in time to an invalid report that still passes the onchain sanity checks.

HashConsensus tests:
- [x] upon changing processor, no report is submitted to a new processor if consensus was lost for the slot
- [x] submitting a different report by a member can lead to consensus loss (partially tested)
- [x] increasing quorum can lead to consensus loss
- [x] consensus cannot be lost if the report is already processing
- [x] `submitReprot` reverts on zero slot

BaseOracle tests:
- [x] `discardConsensusReport` access control
- [x] `discardConsensusReport` does nothing if there's no report for the current frame
- [x] `discardConsensusReport` does nothing if the passed slot is in the future
- [x] `discardConsensusReport` reverts if the slot is less than the last submitted one
- [x] `discardConsensusReport` discards the report that's not yet being processed and emits the event
- [x] `discardConsensusReport` reverts if the report is already being processed
- [x] `_startProcessing` fails when there's no report to process

AccountingOracle tests:
- [x] cannot submit main data if the report was discarded
- [x] `getProcessingState` returns the same state for a discarded report as if it was not submitted

ValidatotExitBusOracle tests:
- [x] cannot submit data if the report was discarded
- [x] `getProcessingState` returns the same state for a discarded report as if it was not submitted


TheDZhon (approved):
LGTM

Thank you, @skozin and @Jeday :+1:

PR #712, opened 2023-03-28
Fix (AccountingOracle): extra data submission order check

skozin:
Fixes the check in `AccountingOracle` that enforces extra data submisison after the main data. It was possible to submit stale extra data even though a new report hash was already pushed by `HashConsensus`.


TheDZhon (approved):
Support the change :+1:

PR #713, opened 2023-03-28
Fix/mixbytes audit

loga4:
Fix mixbytes audit findings

### DIFF
--- a/contracts/0.8.9/EIP712StETH.sol
+++ b/contracts/0.8.9/EIP712StETH.sol
@@ -51,8 +51,8 @@ contract EIP712StETH is IEIP712StETH {
     constructor(address _stETH) {
         if (_stETH == address(0)) { revert ZeroStETHAddress(); }
 
-        bytes32 hashedName = keccak256(bytes("Liquid staked Ether 2.0"));
-        bytes32 hashedVersion = keccak256(bytes("2"));
+        bytes32 hashedName = keccak256("Liquid staked Ether 2.0");
+        bytes32 hashedVersion = keccak256("2");
         bytes32 typeHash = keccak256(
             "EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)"
         );
--- a/contracts/common/lib/Math256.sol
+++ b/contracts/common/lib/Math256.sol
@@ -8,11 +8,6 @@
 pragma solidity >=0.4.24 <0.9.0;
 
 library Math256 {
-    /// @dev Returns the absolute unsigned value of a signed value.
-    function abs(int256 n) internal pure returns (uint256) {
-        return uint256(n >= 0 ? n : -n);
-    }
-
     /// @dev Returns the largest of two numbers.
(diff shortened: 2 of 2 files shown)


contracts/0.8.9/LidoExecutionLayerRewardsVault.sol
```diff
@@ -73,7 +73,9 @@ contract LidoExecutionLayerRewardsVault {
       * @dev execution layer rewards may be sent as plain ETH transfers
       */
     receive() external payable {
-        emit ETHReceived(msg.value);
+        if (msg.value > 0) {
```
TheDZhon, line 76: Let's don't do that
We aren't going to redeploy the EL Vault unless a critical issue discovered

contracts/common/lib/Math256.sol
```diff
@@ -8,11 +8,6 @@
 pragma solidity >=0.4.24 <0.9.0;
 
 library Math256 {
-    /// @dev Returns the absolute unsigned value of a signed value.
-    function abs(int256 n) internal pure returns (uint256) {
```
TheDZhon, line 12: If it is a dead code indeed than thanks for removing it :+1:

contracts/0.8.9/EIP712StETH.sol
```diff
@@ -53,9 +53,9 @@ contract EIP712StETH is IEIP712StETH {
 
         bytes32 hashedName = keccak256(bytes("Liquid staked Ether 2.0"));
         bytes32 hashedVersion = keccak256(bytes("2"));
-        bytes32 typeHash = keccak256(
+        bytes32 typeHash = keccak256(bytes(
```
TheDZhon, line 56: I didn't get why we needed it, tbh :thinking: 

OZ doesn't do that (both for our outdated 4.4 version and the up-to-date 4.8.x)

contracts/0.8.9/LidoExecutionLayerRewardsVault.sol
loga4: yep, it's low, remove last commit

contracts/0.8.9/EIP712StETH.sol
loga4: only for uniformity (above we bring the string to bytes, but not here)

i think we can remove this commit

interesting  case:
internal comment, with bytes keccak256(bytes("test")) ~ 419gas, just string keccak256("test") ~ 337 gas.
- because with bytes it push text to stack and use sha3 op for encode
- for pure keccak256 it calculate hash immediately, i see only push32 {hash} op

TheDZhon: Yeah, it makes sense for consistency. 

Though, I would rather remove casting to `bytes` for the cases above. WDYT?

loga4: done

TheDZhon (approved):
LGTM

PR #714, opened 2023-03-28
Fix (SignatureUtils): check the EIP-1271 call result more strictly

skozin:
Makes sure that:

1. the returned data is exactly 32 bytes long;
2. only the first 4 bytes are non-zero;
3. these 4 bytes contain the magic value.

This doesn't support contracts that right-pad the magic value with zeroes to a size longer than 32 bytes like the Solidity ABI decoder currently allows, but that's a really exotic edge case since it breaks the ABI specification for the `bytes4` return data type.


TheDZhon (approved):
:+1:

PR #715, opened 2023-03-28
Add missing nonce updates to NodeOperatorsRegistry

Psirex:
According to requirements from the `IStakingModule.getNonce()` method, its return value must update each time the deposit data set is changed. This requirement was violated in the following methods:
- `updaeStuckValidatorsCount()`
- `updateExitedValidatorsCount()`
- `unsafeUpdateValidatorsCount()`
- `updateTargetValidatorsLimits()`
- `clearNodeOperatorPenalty()`

### DIFF
--- a/test/helpers/assert.js
+++ b/test/helpers/assert.js
@@ -12,7 +12,7 @@ chai.util.addMethod(chai.assert, 'emits', function (receipt, eventName, args = u
   if (args !== undefined) {
     chai.assert(
       findEventWithArgs(args, events) !== undefined,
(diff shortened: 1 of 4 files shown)


test/helpers/assert.js
```diff
@@ -174,4 +174,9 @@ function normalizeArg(arg) {
   return arg
 }
 
+function stringify(obj) {
+  // Helps overcome the problem with BigInt serialization. Details: https://github.com/GoogleChromeLabs/jsbi/issues/30
```
TheDZhon, line 178: :+1:

PR #719, opened 2023-03-30
fix: remove int64 limitations

krogla:
List of changes:
- Use the `uint256` type instead of `uint64` in `NodeOperatorsRegestry` to work with node operator validators' data. Validate that the resulting value doesn't exceed the `UINT64_MAX` in the `Packed64x4.set()` method.
- Replace calculations on signed variables with equivalent on unsigned int type
- Prevent math operation overflows using `SafeMath` or asserts before/after calculations

NodeOperatorsRegistry contract size reduced from **23.971** to **23.169** KB

Gase changes for most often used methods based on unit tests
|Method Name|Prev Version|New Version|
|-|-|-|
|`obtainDepositData()`|194,414|189,988|
|`onExitedAndStuckValidatorsCountsUpdated()`|8,300,043|8,148,381|
|`updateExitedValidatorsCount()`|118,450|115,006|


contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -577,10 +577,10 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
...
             // upd totals
             Packed64x4.Packed memory summarySigningKeysStats = _loadSummarySigningKeysStats();
-            summarySigningKeysStats.set(
-                SUMMARY_EXITED_KEYS_COUNT_OFFSET,
-                uint64(int64(summarySigningKeysStats.get(SUMMARY_EXITED_KEYS_COUNT_OFFSET)) + totalExitedValidatorsDelta)
-            );
+            // no need to check SUMMARY_EXITED_KEYS_COUNT <= UIN64_MAX due to it is the sum of
+            // operators's TOTAL_EXITED_KEYS_COUNT, each of which does not exceed operator's TOTAL_DEPOSITED_KEYS_COUNT,
+            // which in turn is limited by SafeMath64 when calculating SUMMARY_DEPOSITED_KEYS_COUNT
+            summarySigningKeysStats.set(SUMMARY_EXITED_KEYS_COUNT_OFFSET, uint64(uint256(int256(summarySigningKeysStats.get(SUMMARY_EXITED_KEYS_COUNT_OFFSET)) + totalExitedValidatorsDelta)));
```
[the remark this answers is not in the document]
Psirex: As a rough idea to get rid of chain casts: we can add a method to the `Packed64x4` lib, which will accept delta as `int128` or `int256` and apply it to value. Then we can write something like: `summarySigningKeysStats.slide(SUMMARY_EXITED_KEYS_COUNT_OFFSET, 
 totalExitedValidatorsDelta)`.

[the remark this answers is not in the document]
krogla: yep, but the basic idea is that all numbers we have are uint64 values, so when we convert them to int64, the math can break if the absolute values are higher than uin64_max/2.

to avoid this, we just use a cast to higher bits numbers to do all the math. But again, no one result value will exceed uint64. So converting back to uint64 is safe. Thus, theoretically, even uint65 could be enough)

TheDZhon:
@Psirex please consider resolving the conflicts by integrating the changes from the upstream branch (have divergent changes of lastly pushed commits) :pray:

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -254,28 +250,16 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
...
-            summarySigningKeysStats.set(
-                SUMMARY_MAX_VALIDATORS_COUNT_OFFSET,
-                summarySigningKeysStats.get(SUMMARY_MAX_VALIDATORS_COUNT_OFFSET).add(vettedSigningKeysCountAfter)
-            );
-            summarySigningKeysStats.set(
-                SUMMARY_DEPOSITED_KEYS_COUNT_OFFSET,
-                summarySigningKeysStats.get(SUMMARY_DEPOSITED_KEYS_COUNT_OFFSET).add(depositedSigningKeysCount)
-            );
-            summarySigningKeysStats.set(
+            summarySigningKeysStats.add(SUMMARY_MAX_VALIDATORS_COUNT_OFFSET, vettedSigningKeysCountAfter);
```
TheDZhon, line 253: :+1:

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -927,10 +932,10 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
 
         Packed64x4.Packed memory signingKeysStats = _loadOperatorSigningKeysStats(_nodeOperatorId);
 
-        stakingLimit = signingKeysStats.get(TOTAL_VETTED_KEYS_COUNT_OFFSET);
-        stoppedValidators = signingKeysStats.get(TOTAL_EXITED_KEYS_COUNT_OFFSET);
-        totalSigningKeys = signingKeysStats.get(TOTAL_KEYS_COUNT_OFFSET);
-        usedSigningKeys = signingKeysStats.get(TOTAL_DEPOSITED_KEYS_COUNT_OFFSET);
+        stakingLimit = uint64(signingKeysStats.get(TOTAL_VETTED_KEYS_COUNT_OFFSET));
+        stoppedValidators = uint64(signingKeysStats.get(TOTAL_EXITED_KEYS_COUNT_OFFSET));
```
TheDZhon, line 936: Do we want to save these names for the future?
I know that this method is from the past, though, changing names won't break the ABIs

TheDZhon (approved):
LGTM :+1: 

Please consider addressing the minor outdated naming issue

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
Psirex: Updated the naming for returning variables

TheDZhon: awesome, ty :+1:

PR #720, opened 2023-03-30
Statemind report minor fixes

Psirex:


contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -699,15 +690,21 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
     }
 
     function _checkLimitValue(uint256 _value, uint256 _maxAllowedValue) internal pure {
         if (_value > _maxAllowedValue) {
-            revert IncorrectLimitValue(_value, _maxAllowedValue);
+            revert IncorrectLimitValue(_value, 0, _maxAllowedValue);
+        }
+    }
+
+    function _checkLimitValue(uint256 _value, uint256 _minAllowedValue, uint256 _maxAllowedValue) internal pure {
```
TheDZhon, line 705: Maybe having just a single version for both min and max is better?
Overloads can be dangerous in a long-run :thinking:

TheDZhon (approved):
Thank you, looks pretty reasonable and neat.
Only a single consideration regarding intrinsic limits for sanity checks

PR #721, opened 2023-03-30
Fix: stake limit resume handling

TheDZhon:
image

---

Test cases are in [`test/0.4.24/staking-limit.test.js`](https://github.com/lidofinance/lido-dao/blob/59b0476fef2c89a09e3b7cca264923eb6f9635b9/test/0.4.24/staking-limit.test.js#L244-L286)

### DIFF
--- a/contracts/0.4.24/lib/StakeLimitUtils.sol
+++ b/contracts/0.4.24/lib/StakeLimitUtils.sol
@@ -40,10 +40,10 @@ library StakeLimitState {
       * @dev Internal representation struct (slot-wide)
       */
     struct Data {
-        uint32 prevStakeBlockNumber;
-        uint96 prevStakeLimit;
-        uint32 maxStakeLimitGrowthBlocks;
-        uint96 maxStakeLimit;
+        uint32 prevStakeBlockNumber;      // block number of the previous stake submit
+        uint96 prevStakeLimit;            // limit value (<= `maxStakeLimit`) obtained on the previous stake submit
+        uint32 maxStakeLimitGrowthBlocks; // limit regeneration speed expressed in blocks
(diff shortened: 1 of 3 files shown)


contracts/0.4.24/lib/StakeLimitUtils.sol
```diff
@@ -145,13 +145,19 @@ library StakeLimitUtils {
...
-        // or new limit is lower than previous, then
-        // reset prev stake limit to the new max stake limit
-        if ((_data.maxStakeLimit == 0) || (_maxStakeLimit < _data.prevStakeLimit)) {
+        // reset prev stake limit to the new max stake limit if
+        if (
+            // staking was paused or
+            _data.prevStakeBlockNumber == 0 ||
+            // staking was unlimited or
+            _data.maxStakeLimit == 0 ||
+            // new limit is lower than the previous
```
folkyatina, line 154: Comment is misleading here. It make you think that prevStakeLimit is prevMaxStakeLimit, but it's a current level in the Diablo mana sphere

folkyatina (approved):
Fine, but can optionally use some  work on naming and comments. 
As for me, `prevStakeLimit` is really misleading name here and can use some rebranding.

loga4 (approved):
yep, all is good

contracts/0.4.24/lib/StakeLimitUtils.sol
TheDZhon: yeah, good catch!
ty, fixed :+1:

TheDZhon:
> Fine, but can optionally use some work on naming and comments. As for me, `prevStakeLimit` is really misleading name here and can use some rebranding.

Agree, added a few comments, but decided to not proceed with a large-scale refactoring

PR #725, opened 2023-04-03
Fix: restrict staking module name length to 31 bytes

loga4:
this fix increase codesize by 21bytes

Screenshot 2023-04-04 at 01 40 11


TheDZhon:
Got 18 bytes back :floppy_disk:

PR #727, opened 2023-04-03
Fix: check nor active status 

loga4:


TheDZhon (approved):
:+1: 

FYI: made a few edits in the test to improve readability

PR #730, opened 2023-04-08
Fix: too high simulated share rate

TheDZhon:
Allow the simulated share rate to be either higher or lower within the defined bounds.
A brief explanation was added to the code (see the key extracts below). 
Test cases to illustrate both possible scenarios are provided with the current PR.

```solidity
        // the simulated share rate can be either higher or lower than the actual one
        // in case of new user-submitted ether & minted `stETH` between the oracle reference slot
        // and the actual report delivery slot
        //
        // lower can be for a negative token rebase (token rebase >= one off CL balance decrease)
        // higher can be for a positive token rebase (token rebase <= max positive token rebase)
        //
        // user-submitted ether & minted `stETH` don't exceed the current staking rate limit
        // (see Lido.getCurrentStakeLimit())
        //
        // thus, the `simulatedShareRateDeviationBPLimit` (L) should be set as follows:
        // L = SRL * max(CLD, MPR),
        // where:
        // - CLD is one-off CL balance decrease (as BP),
        // - MPR is max positive token rebase (as BP),
        // - SRL is staking rate limit normalized by TVL (`maxStakeLimit / totalPooledEther`)
        //   totalPooledEther should be chosen as a reasonable lower bound of the protocol TVL
        //
        uint256 simulatedShareDiff = Math256.absDiff(actualShareRate, _simulatedShareRate);
        uint256 simulatedShareDeviation = (MAX_BASIS_POINTS * simulatedShareDiff) / actualShareRate;

        if (simulatedShareDeviation > _limitsList.simulatedShareRateDeviationBPLimit) {
            revert IncorrectSimulatedShareRate(_simulatedShareRate, actualShareRate);
        }
```

### DIFF
--- a/contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
+++ b/contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
@@ -633,20 +633,29 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
             revert ActualShareRateIsZero();
         }
 
-        if (_simulatedShareRate > actualShareRate) {
-            // the simulated share rate can't be higher than the actual one
-            // invariant: rounding only can lower the simulated share rate
-            revert TooHighSimulatedShareRate(_simulatedShareRate, actualShareRate);
-        }
-
-        uint256 simulatedShareDiff = actualShareRate - _simulatedShareRate;
+        // the simulated share rate can be either higher or lower than the actual one
+        // in case of new user-submitted ether & minted `stETH` between the oracle reference slot
+        // and the actual report delivery slot
+        //
+        // lower can be for a negative token rebase (token rebase >= one off CL balance decrease)
+        // higher can be for a positive token rebase (token rebase <= max positive token rebase)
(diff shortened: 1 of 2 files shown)


contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -633,20 +633,29 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
         }
 
-        if (_simulatedShareRate > actualShareRate) {
-            // the simulated share rate can't be higher than the actual one
-            // invariant: rounding only can lower the simulated share rate
-            revert TooHighSimulatedShareRate(_simulatedShareRate, actualShareRate);
-        }
-
-        uint256 simulatedShareDiff = actualShareRate - _simulatedShareRate;
+        // the simulated share rate can be either higher or lower than the actual one
```
TheDZhon, line 636: Provide a scenario when the actual share rate is lower (more users, less rewards per single users)

PR #731, opened 2023-04-09
Certora report's small fixes

Psirex:


contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -1439,6 +1439,11 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
         require(bytes(_name).length > 0 && bytes(_name).length <= MAX_NODE_OPERATOR_NAME_LENGTH, "WRONG_NAME_LENGTH");
     }
 
+    function _onlyValidRewardAddress(address _rewardAddress) internal view {
+        _onlyNonZeroAddress(_rewardAddress);
+        require(_rewardAddress != getLocator().lido(), "LIDO_REWARD_ADDRESS");
```
TheDZhon, line 1444: maybe add a comment why do we need it

TheDZhon (approved):
Thank you for the fixes!

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
Psirex: Agree, added a comment

PR #733, opened 2023-04-10
ERC4906 implementation for withdrawal nft

folkyatina:
- [x] ERC4905 BatchUpdateMetadata on finalize
- [x] token uri now contains all the state
- [x] basic tests for tokenURI
- [x] gas test for tokenURI

### DIFF
--- a/contracts/0.8.9/WithdrawalQueueBase.sol
+++ b/contracts/0.8.9/WithdrawalQueueBase.sol
@@ -41,7 +41,6 @@ abstract contract WithdrawalQueueBase {
     /// @dev timestamp of the last oracle report
     bytes32 internal constant LAST_REPORT_TIMESTAMP_POSITION = keccak256("lido.WithdrawalQueue.lastReportTimestamp");
 
-
     /// @notice structure representing a request for withdrawal.
     struct WithdrawalRequest {
         /// @notice sum of the all stETH submitted for withdrawals up to this request
@@ -217,11 +216,7 @@ abstract contract WithdrawalQueueBase {
         uint256 _maxTimestamp,
         uint256 _maxRequestsPerCall,
         BatchesCalculationState memory _state
-    )
-        external
-        view
-        returns (BatchesCalculationState memory)
-    {
+    ) external view returns (BatchesCalculationState memory) {
         if (_state.finished || _state.remainingEthBudget == 0) revert InvalidState();
 
         uint256 currentId;
@@ -246,7 +241,7 @@ abstract contract WithdrawalQueueBase {
         while (currentId < queueLength && currentId < nextCallRequestId) {
             WithdrawalRequest memory request = _getQueue()[currentId];
 
-            if (request.timestamp > _maxTimestamp) break;  // max timestamp break
+            if (request.timestamp > _maxTimestamp) break; // max timestamp break
 
             (uint256 requestShareRate, uint256 ethToFinalize, uint256 shares) = _calcBatch(prevRequest, request);
 
@@ -364,7 +359,7 @@ abstract contract WithdrawalQueueBase {
             _amountOfETH,
             requestToFinalize.cumulativeShares - lastFinalizedRequest.cumulativeShares,
             block.timestamp
-            );
+        );
     }
 
     /// @dev creates a new `WithdrawalRequest` in the queue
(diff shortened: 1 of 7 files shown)


contracts/0.8.9/WithdrawalQueueERC721.sol
```diff
@@ -338,4 +349,31 @@ contract WithdrawalQueueERC721 is IERC721Metadata, WithdrawalQueue {
...
+        // we still have no string.concat in 0.8.9, so we have to do it with bytes
+        bool finalized = _requestId <= getLastFinalizedRequestId();
+        return string(
+            bytes.concat(
+                bytes(baseURI),
+                bytes(_requestId.toString()),
+                bytes("?status="),
+                bytes(finalized ? "finalized" : "pending"),
+                bytes("&amount="),
+                bytes(
```
TheDZhon, line 367: Let's add both amounts unconditionally.

It's vital to have the requested `stETH` unchanged.
While claimable changes upon the finalization.

contracts/0.8.9/WithdrawalQueueBase.sol
```diff
@@ -494,7 +489,7 @@ abstract contract WithdrawalQueueBase {
     }
 
     /// @dev Calculates discounted ether value for `_requestId` using a provided `_hint`. Checks if hint is valid
-    /// @return claimableEther discounted eth for `_requestId`. Returns 0 if request is not claimable
```
TheDZhon, line 497: I believe that the 'returns 0...' part should be stated for `_getClaimableEther`

contracts/0.8.9/interfaces/IERC4906.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: MIT
```
TheDZhon, line 1: + Copyright of OZ

contracts/0.8.9/WithdrawalQueueERC721.sol
```diff
@@ -146,6 +143,18 @@ contract WithdrawalQueueERC721 is IERC721Metadata, WithdrawalQueue {
         emit NftDescriptorAddressSet(_nftDescriptorAddress);
     }
 
+    /// @notice Finalize requests from last finalized one up to `_lastRequestIdToFinalize`
+    /// @dev ether to finalize all the requests should be calculated using `finalizationValue()` and sent along
+    function finalize(uint256[] calldata _batches, uint256 _maxShareRate) external payable {
```
TheDZhon, line 148: Maybe it's better to leave the `finalize` method in `WithdrawalQueue` but override it here, emitting the event in addition?

TheDZhon (commented):
Looks good :+1: 

Please consider using both amounts together (not exclusively).
Also, have merge conflicts with tests.

contracts/0.8.9/WithdrawalQueueERC721.sol
```diff
@@ -338,4 +349,31 @@ contract WithdrawalQueueERC721 is IERC721Metadata, WithdrawalQueue {
...
+    function _constructTokenUri(uint256 _requestId) internal view returns (string memory) {
+        string memory baseURI = _getBaseURI().value;
+        if (bytes(baseURI).length == 0) return "";
+
+        // ${baseUri}/${_requestId}?state=finalized|unfinalized&amount=${amount}&created_at=${timestamp}
+        // we still have no string.concat in 0.8.9, so we have to do it with bytes
+        bool finalized = _requestId <= getLastFinalizedRequestId();
+        return string(
+            bytes.concat(
+                bytes(baseURI),
```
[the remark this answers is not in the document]
folkyatina: Good point. Double slash is better than no one.

contracts/0.8.9/WithdrawalQueueERC721.sol
```diff
@@ -18,22 +19,18 @@ import {AccessControlEnumerable} from "./utils/access/AccessControlEnumerable.so
...
+    /// @param _requestId is an id for particular withdrawal request
     function constructTokenURI(uint256 _requestId) external view returns (string memory);
 }
 
 /// @title NFT implementation on top of {WithdrawalQueue}
 /// NFT is minted on every request and burned on claim
 ///
 /// @author psirex, folkyatina
-contract WithdrawalQueueERC721 is IERC721Metadata, WithdrawalQueue {
+contract WithdrawalQueueERC721 is IERC721Metadata, WithdrawalQueue, IERC4906 {
```
Jeday, line 33: General style for inheritance order is interfaces then base contracts, right?

Jeday (approved):

contracts/0.8.9/WithdrawalQueueERC721.sol
folkyatina: Idea is great but implementation can be messy because of the payable. And, I'd rather prefer to put all the checks in the top method.

contracts/0.8.9/WithdrawalQueueERC721.sol
folkyatina: Thanks.

TheDZhon (approved):
:+1: Will merge with the gas tests included

PR #734, opened 2023-04-10
Fix: reduce slots withdrawals

loga4:


contracts/0.8.9/WithdrawalQueueBase.sol
```diff
@@ -70,10 +70,10 @@ abstract contract WithdrawalQueueBase {
         uint256 amountOfStETH;
         /// @notice amount of stETH shares locked on withdrawal queue for this request
         uint256 amountOfShares;
-        /// @notice address that can claim or transfer this request
-        address owner;
         /// @notice timestamp of when the request was created, in seconds
         uint256 timestamp;
+        /// @notice address that can claim or transfer this request
+        address owner;
```
TheDZhon, line 76: Let's don't do that at all.

We are not storing this struct in storage, so it's not worthful 
https://docs.soliditylang.org/en/v0.8.17/internals/layout_in_memory.html?highlight=struct#example-for-difference-in-struct-layout

TheDZhon (changes requested):
Thank you for the PR! 
Yet I propose to park this idea.

PR #743, opened 2023-04-11
Feat: split churn limit into exited and appeared

TheDZhon:
Disentangle the meaning of `appeared` and `activated` validators.

```solidity
    /// @notice The max possible number of validators that might be reported as `exited`
    ///     per single day, depends on the Consensus Layer churn limit
    /// @dev Must fit into uint16 (<= 65_535)
    uint256 exitedValidatorsPerDayLimit;

    /// @notice The max possible number of validators that might be reported as `appeared`
    ///     per single day, limited by the max daily deposits via DepositSecurityModule in practice
    ///     isn't limited by a consensus layer (because `appeared` includes `pending`, i.e., not `activated` yet)
    /// @dev Must fit into uint16 (<= 65_535)
    uint256 appearedValidatorsPerDayLimit;
 ```


Psirex (approved):
Looks nice 👍🏻

PR #744, opened 2023-04-11
fix int type initialization

loga4:


TheDZhon:
Decided not to merge it

PR #751, opened 2023-04-12
Statemind minor fixes round 2

Psirex:


TheDZhon:
Quite strange that we have merge conflicts with `lib/abi/*` stuff.
They were removed already during the `rc1` preps :thinking:

PR #757, opened 2023-04-20
Aragon Apps: followup updates

mymphe:
- fix all oracles showing up as fast-lane members
- use monospace font for bytes for uniform length


TheDZhon (approved):
:rocket:

PR #767, opened 2023-05-05
ci: Fix slither version

dgusakov:


dgusakov:
Will be fixed in https://github.com/lidofinance/lido-dao/pull/768

PR #772, opened 2023-05-24
Add check against underflow

claudijd:
This adds a simple check on AccountingOracleMock.sol to avoid it underflowing.

This is just the mock, so not live implementation risk here.


earuak:
The check ensures that the value of **data.refSlot** is always greater than or equal to the value of **_lastRefSlot,** thus avoiding calculation errors that could result in a negative value.

Adding checks to prevent underflows is a recommended practice in smart contracts to avoid security issues and calculation errors. This approach demonstrates an extra level of care in validating input data and can contribute to the robustness and reliability of the contract.

However, it's important to consider the entire context of the code and perform a thorough review of the proposed changes in the pull request to assess their suitability and overall impact on the system. Additionally, it is recommended to have the code reviewed by developers specialized in smart contract security before deploying it in a production environment.

PR #776, opened 2023-06-17
Add unit-tests for PausableUtil

claudijd:
This PR adds forge unit-tests for PausableUtil

### DIFF
--- a/test/0.8.9/utils/pausable-util.test.sol
+++ b/test/0.8.9/utils/pausable-util.test.sol
@@ -1,0 +1,251 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
+
+import "forge-std/Test.sol";
+import { PausableUntil } from "contracts/0.8.9/utils/PausableUntil.sol";
+
+contract ExposedPausableUntil is PausableUntil {
+
+    function infinity() public pure returns (uint256) {
+        return PAUSE_INFINITELY;
+    }    
+
+    function setPausedState(uint256 _resumeSince) public {
+        _setPausedState(_resumeSince);
+    }
+
+    function pauseUntil(uint256 _pauseUntilInclusive) public {
+        _pauseUntil(_pauseUntilInclusive);
+    }
+
+    function pauseFor(uint256 _duration) public {
+        _pauseFor(_duration);
+    }
+
+    function resume() public {
+        _resume();
+    }
+
+    function checkResumed() public view {
+        _checkResumed();
+    }
+
+    function checkPaused() public view {
+        _checkPaused();
+    }
+}
+
+contract PausableUntilTest is Test {
+    ExposedPausableUntil public pausableUntil;
+    ExposedPausableUntil public pausableUntilTwo;
+
+    function setUp() public {
+        pausableUntil = new ExposedPausableUntil();
+        pausableUntilTwo = new ExposedPausableUntil();
+    }
+
+    function testIsNotPausedByDefault() public {
+        assertEq(pausableUntil.isPaused(), false);
+        assertEq(pausableUntilTwo.isPaused(), false);
+    }
+
+    function testSetPausedState(uint256 _resumeSince, uint256 _randomTime) public {
+        // ignore obvious 1 state and underflow condition
+        vm.assume(_resumeSince != 1);
+        vm.assume(_resumeSince >= block.timestamp);
+
+        // make sure we're not already paused
+        assertEq(pausableUntil.isPaused(), false);
+
+        // pause until _resumeSince
+        pausableUntil.setPausedState(_resumeSince);
+        assertEq(pausableUntil.isPaused(), true);
+
+        // check a random point in time (either before or after _resumeSince)
+        vm.warp(_randomTime);
+        if (_randomTime < _resumeSince) {
+            assertEq(pausableUntil.isPaused(), true);
+        } else {
+            assertEq(pausableUntil.isPaused(), false);  
+        }
+
+        // go forward in time to _resumeSince - 1 block, check edge
+        vm.warp(_resumeSince - 1);
+        assertEq(pausableUntil.isPaused(), true);
+
+        // go forward in time to _resumeSince block, check edge
+        vm.warp(_resumeSince);
+        assertEq(pausableUntil.isPaused(), false);
+    }
+
+    function testPauseUntil(uint256 _pauseUntilInclusive, uint256 _randomTime) public {
+        vm.assume(_pauseUntilInclusive != 0);
+        vm.assume(_randomTime != 0);
+        vm.assume(_randomTime != _pauseUntilInclusive);
+                
+        if (_pauseUntilInclusive < block.timestamp) {
+            vm.expectRevert();
+            pausableUntil.pauseUntil(_pauseUntilInclusive);
+            assertEq(pausableUntil.isPaused(), false);
+        } else {
+            pausableUntil.pauseUntil(_pauseUntilInclusive);
+            assertEq(pausableUntil.isPaused(), true);
+
+            // check a random point in time (either before or after _pauseUntilInclusive)
+            vm.warp(_randomTime);
+            if (_randomTime < _pauseUntilInclusive) {
+                assertEq(pausableUntil.isPaused(), true);
+            } else {
+                assertEq(pausableUntil.isPaused(), false);  
+            }
+        }
+    }
+
+    function testPauseUntilInfinity(uint256 _randomTime) public {
+        assertEq(pausableUntil.isPaused(), false);
+        pausableUntil.pauseUntil(pausableUntil.infinity());
+        assertEq(pausableUntil.isPaused(), true);
+
+        // check a random point in time
+        vm.warp(_randomTime);
+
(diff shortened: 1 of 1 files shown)


test/0.8.9/utils/pausable-util.test.sol
```diff
@@ -0,0 +1,186 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.4.24 <0.9.0;
```
loga4, line 2: Let's lock in the version 0.8.9 (as PausableUntil.sol)

test/0.8.9/utils/pausable-util.test.sol
```diff
@@ -0,0 +1,186 @@
...
+    function testSetPausedState(uint256 _resumeSince, uint256 _random_time) public {
+        // ignore obvious 1 state and underflow condition
+        vm.assume(_resumeSince != 1);
+        vm.assume(_resumeSince >= block.timestamp);
+
+        // make sure we're not already paused
+        assertEq(pausableUntil.isPaused(), false);
+
+        // pause until _resumeSince
+        pausableUntil.setPausedState(_resumeSince);
```
loga4, line 53: Should we check events via [expectEmit](https://book.getfoundry.sh/cheatcodes/expect-emit), wdyt?

claudijd: I looked into this, but I'm having trouble trying to sort out the syntax of expectEmit.  I see the docs suggest a true, true, true, true pattern, but it's not clear to me what that's trying to acheive.  It would seem natural to me that I should be able to say vm.expectEmit(NAME_OF_EVENT) and that seems to not work.  I tried including in the Exposed contract, but got a name space collision on defining the events because they collide with PausableUntil.   I tried including them in PausableUntilTest, but forge bombs out on expectEmit not found.  Any advise you might be able to offer on how to syntactically do this, I'm happy to implement it.

test/0.8.9/utils/pausable-util.test.sol
```diff
@@ -0,0 +1,186 @@
...
+            assertEq(pausableUntil.isPaused(), true);
+        } else {
+            assertEq(pausableUntil.isPaused(), false);  
+        }
+    }
+
+    function testResumeWhenPaused() public {
+        assertEq(pausableUntil.isPaused(), false);
+
+        // pause for the next 3 blocks
```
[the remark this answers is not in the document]
claudijd: Fixed in https://github.com/lidofinance/lido-dao/pull/776/commits/4a733a85e98dbe94de3a11adb35ce930eab4ad9a

test/0.8.9/utils/pausable-util.test.sol
claudijd: Fixed in [00f858f](https://github.com/lidofinance/lido-dao/pull/776/commits/00f858f90116b225f9bd49c768670b5b2f47aed2)

test/0.8.9/utils/pausable-util.test.sol
loga4: Thank you so much for your time!

Let me try to help. If you look at the documentation, you can see that the entire event test flow is divided into 3 steps:

1. Call cheat code, like `vm.expectEmit(true, true, true, true);`  - func has 4 parameters:
 - checkTopic1 - do we need to check the first indexed topic from event
 - checkTopic2 - check the second
 - checkTopic3 - check the third
 - checkData - do we need to check/compare data in topics

2. Emit expected event with expected data. Here we "assign" expected event to foundry.

3. Call the real function that we want to test. Like `pausableUntil.setPausedState(_resumeSince);`.
 - If `checkData` was checked as true - foundry compare expected data from step2 and actual from step3.

So if briefly recap: 
 1. Prepare expected data ( your `vm.expectEmit(NAME_OF_EVENT)` divided into 2 steps : `vm.expectedEmit(...)+ emit Event(..)` )
 2. Call actual func from smart contract with event

test/0.8.9/utils/pausable-util.test.sol
```diff
@@ -0,0 +1,186 @@
...
+        pausableUntil.setPausedState(block.timestamp + 3);
+
+        // pick a random time
+        vm.warp(_random_time);
+
+        if (_random_time >= (original_timestamp + 3)) {
+            pausableUntil.checkResumed();
+        } else {
+            vm.expectRevert();
+            pausableUntil.checkResumed();
```
loga4, line 166: is redudant, lets refactor this block like

```
if (random_time < (original_timestamp + 3)) {
 vm.expectedRevert
}

pausableUntil.checkResumed();
```

test/0.8.9/utils/pausable-util.test.sol
```diff
@@ -0,0 +1,186 @@
...
+        if (_pauseUntilInclusive < block.timestamp) {
+            vm.expectRevert();
+            pausableUntil.pauseUntil(_pauseUntilInclusive);
+            assertEq(pausableUntil.isPaused(), false);
+        } else {
+            pausableUntil.pauseUntil(_pauseUntilInclusive);
+            assertEq(pausableUntil.isPaused(), true);
+        }
+
+        // check a random point in time (either before or after _pauseUntilInclusive)
```
loga4, line 87: can we add additional test/assert for PAUSE_INFINITELY value for `pauseFor` and `pauseUntil`?

test/0.8.9/utils/pausable-util.test.sol
```diff
@@ -0,0 +1,186 @@
...
+    function checkResumed() public view {
+        _checkResumed();
+    }
+
+    function checkPaused() public view {
+        _checkPaused();
+    }
+}
+
+contract PausableUntilTest is Test {
```
TheDZhon, line 33: can we have a fuzzed test that `pauseFor` effectively is dual to `pauseUntil`:

I mean the following in pseudocode
```solidity
pausableUntil1.pauseFor(x)
pausableUntil2.pauseUntill(block.timestamp + x - 1)

assertEq(
   pausableUntil1.getResumeSinceTimestamp(), 
   pausableUntil2.getResumeSinceTimestamp()
)
```

test/0.8.9/utils/pausable-util.test.sol
```diff
@@ -0,0 +1,186 @@
...
+    ExposedPausableUntil public pausableUntil;
+
+    function setUp() public {
+        pausableUntil = new ExposedPausableUntil();
+    }
+
+    function testIsPaused() public {
```
TheDZhon, line 40: I'd suggest renaming to `testIsPausedByDefault`

test/0.8.9/utils/pausable-util.test.sol
```diff
@@ -0,0 +1,186 @@
...
+        assertEq(pausableUntil.isPaused(), false);
+    }
+
+    function testSetPausedState(uint256 _resumeSince, uint256 _random_time) public {
+        // ignore obvious 1 state and underflow condition
+        vm.assume(_resumeSince != 1);
```
TheDZhon, line 46: why is the case with `1` so special? :thinking:

test/0.8.9/utils/pausable-util.test.sol
TheDZhon: Yeah, agree :+1:

test/0.8.9/utils/pausable-util.test.sol
TheDZhon: It would be great to have together with a special case check for the `PAUSE_INFINITELY` value here

test/0.8.9/utils/pausable-util.test.sol
```diff
@@ -0,0 +1,186 @@
...
+        } else {
+            assertEq(pausableUntil.isPaused(), false);  
+        }
+    }
+
+    function testResumeWhenPaused() public {
+        assertEq(pausableUntil.isPaused(), false);
+
+        // pause for the next 3 seconds
+        pausableUntil.setPausedState(block.timestamp + 3);
```
TheDZhon, line 131: would be great to not use a hardcoded constant because the real network has a twelve seconds cadence
(can fuzz it, maybe?)

test/0.8.9/utils/pausable-util.test.sol
```diff
@@ -0,0 +1,186 @@
...
+        vm.assume(_resumeSince >= _random_time);
+        vm.warp(_random_time);
+        pausableUntil.setPausedState(_resumeSince);
+        assertEq(pausableUntil.getResumeSinceTimestamp(), _resumeSince);
+    }
+
+    function testCheckResumed(uint256 _random_time) public {
+        // set paused for 3 seconds
+        uint256 original_timestamp = block.timestamp;
+        pausableUntil.setPausedState(block.timestamp + 3);
```
TheDZhon, line 157: the same consideration regarding `3` as noted above :pray:

TheDZhon (commented):
Thank you for being with us for so long :+1: 

Suggest a few edits to achieve bullet-proof coverage :shield:

mymphe:
Closed!
This repo is no longer maintained. LidoFinance protocol moved to [core repository](https://github.com/lidofinance/core).

PR #805, opened 2023-11-21
Scratch deploy: capability to set WithdrawalQueue baseUri

arwer13:
Just a capability / reminder to set it upon deploy. It happened to be missing for Holesky deployment.


scripts/scratch/16-initialize-non-aragon-contracts.js
```diff
@@ -125,14 +125,26 @@ async function deployNewContracts({ web3, artifacts }) {
...
-  const withdrawalQueue = await artifacts.require('WithdrawalQueueERC721').at(withdrawalQueueAddress)
   await makeTx(withdrawalQueue, 'initialize', withdrawalQueueArgs, { from: DEPLOYER })
 
+  //
+  // === WithdrawalQueue: setBaseURI ===
+  //
+  const withdrawalQueueBaseUri = state["withdrawalQueueERC721"].deployParameters.baseUri
+  if (withdrawalQueueBaseUri !== null && withdrawalQueueBaseUri !== "") {
+    const MANAGE_TOKEN_URI_ROLE = await withdrawalQueue.MANAGE_TOKEN_URI_ROLE()
+    await makeTx(withdrawalQueue, 'grantRole', [MANAGE_TOKEN_URI_ROLE, DEPLOYER], { from: DEPLOYER })
```
folkyatina, line 142: Is it easy to make it one tx?

scripts/scratch/dao-local-deploy.sh
```diff
@@ -7,6 +7,7 @@ export NETWORK=local
 export RPC_URL="http://127.0.0.1:8545"
 export GATE_SEAL_FACTORY=0x0000000000000000000000000000000000000000
```
folkyatina, line 8: Can we clarify why it's a ZERO_ADDRESS?

folkyatina (approved):
unstETH full name is wrong, everything else is gtg

deployed-testnet-defaults.json
```diff
@@ -119,7 +119,8 @@
   "withdrawalQueueERC721": {
     "deployParameters": {
       "name": "stETH Withdrawal NFT",
```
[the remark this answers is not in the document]
arwer13: Oh, thanks!

Than we also have incorrect names on Goerli and Holesky. Submitted 
 https://github.com/lidofinance/lido-dao/issues/806

scripts/scratch/16-initialize-non-aragon-contracts.js
arwer13: Would like to see it atomic indeed, but don't know a simple nice solution, thb

In general, think having UpgradeTemplate-like helper contract might be the way

deployed-testnet-defaults.json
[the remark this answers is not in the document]
arwer13: Fixed

scripts/scratch/16-initialize-non-aragon-contracts.js
folkyatina: It's possible by utilizing the Multicall contract that is usually deployed everywhere and well-known. Ape/brownie even have a built-in support for that.

arwer13: I'm not sure it works when the call goes from EOA-deployer which is the admin due to 

> Because you cannot delegatecall from an EOA, this significantly reduces the benefit of calling Multicall3 from an EOA—any calls the Multicall3 executes will have the MultiCall3 address as the msg.sender. This means you should only call Multicall3 from an EOA if the msg.sender does not matter.

from https://github.com/mds1/multicall#batch-contract-writes

Is it?

folkyatina: Yep. Correct. Let it be separate txs for now. 

But, the audited deployer/template contract may be worth considering in the next iteration.

arwer13: I think even unaudited one is a worthy improvement.

PR #821, opened 2024-02-07
Sepolia deposit adapter implementation

vp42ldo:
Sepolia deposit adapter implementation + tests


test/0.8.9/sepolia-deposit-adapter.test.js
```diff
@@ -0,0 +1,81 @@
...
+    snapshot = new EvmSnapshot(ethers.provider)
+    await snapshot.make()
+  })
+
+  afterEach(async () => {
+    await snapshot.rollback()
+  })
+
+  describe('SepoliaDepositAdapter Logic', () => {
+    it(`transfer Bepolia tokens`, async () => {
```
folkyatina, line 35: Looks like it should be in 'before'?

folkyatina (commented):
Already have all the essentials, but there is some room for improvement.

test/0.8.9/sepolia-deposit-adapter.test.js
vp42ldo: Well. that could be but it's not required for every test

folkyatina: Ok. Then, it could be a separate context just for the test that requires a token with its own before hook. 

Now, it looks like we're testing an impersonated version of a token, which we should not.

contracts/0.8.9/SepoliaDepositAdapter.sol
```diff
@@ -4,24 +4,82 @@
...
+
+    function get_deposit_root() external view returns (bytes32) {
+        return originalContract.get_deposit_root();
+    }
+
+    function get_deposit_count() external view returns (bytes memory) {
+        return originalContract.get_deposit_count();
     }
 
+    function test() external view returns (string memory) {
```
folkyatina, line 46:
```suggestion
    function name() external view returns (string memory) {
```

contracts/0.8.9/SepoliaDepositAdapter.sol
```diff
@@ -4,24 +4,82 @@
...
+          "Receive %s from %s",
+            msg.value,
+            msg.sender
+        );
+    }
 
+    function drain() external onlyOwner {
+        uint ownTokens = address(this).balance;
+        address payable _owner = payable(owner());
+        _owner.transfer(ownTokens);
```
folkyatina, line 61: I recalled why transfer() is generally not recommended. 

https://www.rareskills.io/post/smart-contract-security#:~:text=Transfer()%20and%20send()%20can%20break%20with%20multi%2Dsignature%20wallets

test/0.8.9/sepolia-deposit-adapter.test.js
vp42ldo: We are impersonating the account with tokens. Account is our EOA and not the same as temp account spinning with hardhat for tests. So, it's not possible to transfer tokens from EOA to contract or did I miss something?

contracts/0.8.9/SepoliaDepositAdapter.sol
vp42ldo: ok. I can do destination address as a parameter. Do you have other suggestions?

test/0.8.9/sepolia-deposit-adapter.test.js
```diff
@@ -0,0 +1,142 @@
...
+      return this.skip()
+    }
+
+    // depositAdapter = await SepoliaDepositAdapter.at(sepoliaDepositAdapterContract)
+    depositAdapter = await ethers.deployContract('SepoliaDepositAdapter', [sepoliaDepositContract])
+    log('depositAdapter address', depositAdapter.address)
+
+    bepoliaToken = await ethers.getContractAt('ISepoliaDepositContract', sepoliaDepositContract)
+
+    const result = await depositAdapter.name()
```
TheDZhon, line 31:
```suggestion
    const name = await depositAdapter.name()
```

test/0.8.9/sepolia-deposit-adapter.test.js
```diff
@@ -0,0 +1,142 @@
...
+
+    snapshot = new EvmSnapshot(ethers.provider)
+    await snapshot.make()
+  })
+
+  afterEach(async () => {
+    await snapshot.rollback()
+  })
+
+  describe('SepoliaDepositAdapter Logic', () => {
```
TheDZhon, line 42: I guess we might need to test:
- `receive` method
- events to be emitted, esp `DepositEvent`

contracts/0.8.9/SepoliaDepositAdapter.sol
```diff
@@ -4,24 +4,77 @@
...
+        uint balance = address(this).balance;
+        (bool success,) = owner().call{value: balance}("");
+        if (!success) {
+            revert TransferFailed();
+        }
+    }
+
+    function recoverBepolia() external onlyOwner {
+        uint bepoliaOwnTokens = originalContract.balanceOf(address(this));
+        bool success = originalContract.transfer(owner(), bepoliaOwnTokens);
```
TheDZhon, line 62: nitpick: usually it's better to use `SafeERC20` and `safeTransfer` to avoid issues with quasi-ERC20 conformant tokens (e.g. USDT)

contracts/0.8.9/SepoliaDepositAdapter.sol
```diff
@@ -4,24 +4,77 @@
...
+    ) external payable;
+
+    function get_deposit_root() external view returns (bytes32);
+
+    function get_deposit_count() external view returns (bytes memory);
+
+    function name() external view returns (string memory);
+}
+
+contract SepoliaDepositAdapter is Ownable {
```
TheDZhon, line 27: I would put a piece of comment for future readers: why we need this adapter and what is the idea behind it :eyes:

TheDZhon (changes requested):
:+1: Thank you for putting everything together!

I'd suggest one more round of refinements before declaring it as completed :pray:

test/0.8.9/sepolia-deposit-adapter.test.js
vp42ldo: It's the only item left. I can't detect `DepositEvent` was sent from the original contract. Still figuring this out.

vp42ldo: ok. this solved as well

contracts/0.8.9/SepoliaDepositAdapter.sol
```diff
@@ -4,24 +4,77 @@
...
+            revert TransferFailed();
+        }
+    }
+
+    function recoverBepolia() external onlyOwner {
+        uint bepoliaOwnTokens = originalContract.balanceOf(address(this));
+        bool success = originalContract.transfer(owner(), bepoliaOwnTokens);
+        if (!success) {
+            revert TransferFailed();
+        }
```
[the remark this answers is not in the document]
TheDZhon: @vp42ldo sorry for asking twice, but what is your attitude towards the suggestion above? :pray:

test/0.8.9/sepolia-deposit-adapter.test.js
```diff
@@ -0,0 +1,157 @@
...
```
TheDZhon, line 44: thank you for improving the tests :+1: 

it would be great to test revert conditions if possible; though not a blocker for the testnet contract, ofc

TheDZhon (approved):
LGTM :rocket:

PR #824, opened 2024-02-21
Update deploy script to include SepoliaDepositAdapter contract upgrade  

vp42ldo:
Update deploy script to include SepoliaDepositAdapter contract upgrade

### DIFF
--- a/scripts/deploy-sepolia-deposit-contract-adapter.js
+++ b/scripts/deploy-sepolia-deposit-contract-adapter.js
@@ -3,20 +3,41 @@ const { deployBehindOssifiableProxy } = require('./helpers/deploy')
 
 const DEPLOYER = process.env.DEPLOYER || ''
 
-async function deployNewContracts({ web3, artifacts }) {
-  if (!DEPLOYER) {
-    throw new Error('Deployer is not specified')
-  }
+async function updateAdapterImplementation(proxyOwner) {
+  const sepoliaDepositAdapter = "<new-deposit-adapter-address>"
+  const proxyAddress = "0x80b5DC88C98E528bF9cb4B7F0f076aC41da24651"
 
+  const OssifiableProxy = await artifacts.require('OssifiableProxy')
+  const proxy = await OssifiableProxy.at(proxyAddress)
+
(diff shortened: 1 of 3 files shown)


TheDZhon (commented):
:thinking: Scarcely am I sure if we need to hardcode the addresses here instead of putting/reading them from the `deployed-sepolia.json`

wdyt, @arwer13?

scripts/deploy-sepolia-deposit-contract-adapter.js
```diff
@@ -3,20 +3,41 @@ const { deployBehindOssifiableProxy } = require('./helpers/deploy')
...
+  const sepoliaDepositAdapter = "<new-deposit-adapter-address>"
+  const proxyAddress = "0x80b5DC88C98E528bF9cb4B7F0f076aC41da24651"
+
+  const OssifiableProxy = await artifacts.require('OssifiableProxy')
+  const proxy = await OssifiableProxy.at(proxyAddress)
 
+  await proxy.proxy__upgradeTo(sepoliaDepositAdapter, { from: proxyOwner })
+}
+
+async function deployAdaperBehindPrixy(depositAdapterProxyOwner) {
```
arwer13, line 16: typo in Prixy

arwer13 (changes requested):

vp42ldo:
> @TheDZhon @vp42ldo There is `updateProxyImplementation` from deploy.js for the case (if the adapter implementation not deployed yet). It will read the proxy address from the state file, deploy the implementation, bind it to the proxy and update the impl address in the state file.

There is an issue with `updateProxyImplementation`. DepositContractAdapter and it's proxy are not part of deployed-<network>.json as it's specific for Sepolia-deploy only. Currently there is no reference to the proxy address except a parameter to stakingRouter

arwer13:
Oh, we've forgotten to add it to the state file :/ I'd propose to add it now under `/depositContractAdapter`. It is part of the setup and thus must be presented in the state file as well.

TheDZhon:
Having it would be prudent, I agree with @arwer13  :pray:

arwer13 (approved):
lgtm, taking into account the deploy script is most likely one-time-usage and the migration to the new core repo and typescript (won't move it there)

PR #844, opened 2024-10-07
docs(readme): add repovation disclaimer

failingtwice:
- add a disclaimer section about repovation and link the last legacy commit


tamtamchik (approved):
💚

PR #826, opened 2024-02-23
new SimpleDVT module

krogla:
- scripts for deploy SimpleDVT module
- new Aragon app for SimpleDVT (mostly clone of NOR app)
- minor scripts and type fixes

PR #845, opened 2024-10-07
feat: add Simple DVT to scratch deploy

tamtamchik:
- Restore #826


contracts/0.4.24/template/LidoTemplate.sol
```diff
@@ -357,6 +355,28 @@ contract LidoTemplate is IsContract {
...
+            this,
+            _initialSemanticVersion,
+            _impl,
+            _contentURI
+        );
+
+        bytes32 appId = _getAppId(SIMPLE_DVT_APP_NAME, deployState.lidoRegistryEnsNode);
+        dao.setApp(dao.APP_BASES_NAMESPACE(), appId, _impl);
+        deployState.sdvt = NodeOperatorsRegistry(_proxy);
+    }
```
TheDZhon: there should be an event maybe (similar to `emit TmplAPMDeployed(address(registry))`)

contracts/0.4.24/template/LidoTemplate.sol
```diff
@@ -357,6 +355,28 @@ contract LidoTemplate is IsContract {
         deployState = state;
     }
 
+    function createSimpleDVTApp(
+        uint16[3] _initialSemanticVersion,
+        address _proxy,
+        address _impl,
+        bytes _contentURI
+    ) external onlyOwner {
+        APMRegistry lidoRegistry = deployState.lidoRegistry;
```
TheDZhon: Should there be any prerequisite requirements? E.g. requiring NOR to be deployed upfront

globals.d.ts
```diff
@@ -1,19 +1,21 @@
...
+    HARDHAT_FORKING_URL?: string;
 
     /**
      * Flags for changing the behavior of the integration tests
      */
 
     /* if "on" the integration tests will deploy the contracts to the empty Hardhat Network node using scratch deploy */
-    INTEGRATION_SCRATCH_DEPLOY?: "on" | "off";
-    /* if "on" the integration tests will enable assertions and checks for the simple DVT module */
-    INTEGRATION_SIMPLE_DVT_MODULE?: "on" | "off";
```
TheDZhon, line 16: Hmm, the template has an optionality to deploy sDVT or not, while here, we exclude it deliberately. Not sure if this is consistent.

lib/protocol/networks.ts
```diff
@@ -80,8 +80,7 @@ async function getLocalNetworkConfig(network: string, source: string): Promise<P
     locator: config["lidoLocator"].proxy.address,
     agentAddress: config["app:aragon-agent"].proxy.address,
     votingAddress: config["app:aragon-voting"].proxy.address,
-    easyTrackAddress: config["app:aragon-agent"].proxy.address,
-    sdvt: config["app:node-operators-registry"].proxy.address,
```
TheDZhon, line 84: why remove it here?

contracts/0.4.24/template/LidoTemplate.sol
```diff
@@ -614,6 +619,20 @@ contract LidoTemplate is IsContract {
         acl.createPermission(_state.stakingRouter, _state.operators, _state.operators.STAKING_ROUTER_ROLE(), voting);
         acl.createPermission(_state.agent, _state.operators, _state.operators.MANAGE_NODE_OPERATOR_ROLE(), voting);
 
+        if (deployState.sdvt != address(0)) {
```
TheDZhon: Not sure about this optionality branching (see my comments with the 'optionality' word)

scripts/scratch/steps/0150-plug-staking-modules.ts
```diff
@@ -5,21 +5,22 @@ import { makeTx } from "lib/deploy";
...
 import { readNetworkState, Sk } from "lib/state-file";
 
+const STAKING_MODULE_MANAGE_ROLE = streccak("STAKING_MODULE_MANAGE_ROLE");
+
 const NOR_STAKING_MODULE_TARGET_SHARE_BP = 10000; // 100%
 const NOR_STAKING_MODULE_MODULE_FEE_BP = 500; // 5%
 const NOR_STAKING_MODULE_TREASURY_FEE_BP = 500; // 5%
-const STAKING_MODULE_MANAGE_ROLE = streccak("STAKING_MODULE_MANAGE_ROLE");
+
+const SDVT_STAKING_MODULE_TARGET_SHARE_BP = 50; // 0.5%
```
TheDZhon: let's set to 4% as the current value for Mainnet (if doesn't break tests badly)

TheDZhon (approved):
Fantastic work it is :clap: 

Left some comments and considerations :thinking:

globals.d.ts
tamtamchik: The template can also be used to deploy on testnets, and in that case, the step with SDVT may be skipped. However, local scratch deployment now always enables the SDVT module to test the state with 2 modules. This is why I decided to remove the bias for running with or without SDVT, since on the mainnet it always runs with SDVT enabled. But this is a good catch! I'm not solid in my decision here.

lib/protocol/networks.ts
tamtamchik: `easyTrackAddress` goes to `app: Aragon-voting,` and sdvt now is taken directly from deployment artifacts. No need to link NOR manually anymore.

scripts/scratch/steps/0150-plug-staking-modules.ts
tamtamchik: Done in 57ea90e123a532684f21c0a040c63a594517cfda

contracts/0.4.24/template/LidoTemplate.sol
tamtamchik: Done in 57ea90e123a532684f21c0a040c63a594517cfda

contracts/0.4.24/template/LidoTemplate.sol
tamtamchik: Done in 57ea90e123a532684f21c0a040c63a594517cfda

contracts/0.4.24/template/LidoTemplate.sol
tamtamchik: Same here. Agree to remove optionality if there will be no deploys without SDVT in future.

TheDZhon (approved):
GG

PR #854, opened 2024-10-10
sync: develop from core repo

tamtamchik:


TheDZhon (approved):
Seems legit :eyes: 

Compared against [master](https://github.com/lidofinance/lido-dao/compare/master..sync/develop) and found only Negative Rebase protection changes

PR #855, opened 2024-10-10
ci: fix master coverage

tamtamchik:
Currently, the master branch coverage is not being collected because the workflow only runs on PRs. It also needs to run on push to master to update the reference coverage.


TheDZhon (approved):
Ship it :ship:

PR #857, opened 2024-10-11
test: Second opinion oracle test

tamtamchik:
Port of https://github.com/lidofinance/core/pull/211


vp4242:
Seems that the reason for failed tests that PostTotalShares event is no longer emitted after LegacyOracle deprecation.

test/integration/second-opinion.ts
```diff
@@ -0,0 +1,195 @@
...
+    const { lido, depositSecurityModule, oracleReportSanityChecker } = ctx.contracts;
+
+    await finalizeWithdrawalQueue(ctx, stEthHolder, ethHolder);
+
+    await norEnsureOperators(ctx, 3n, 5n);
+    await sdvtEnsureOperators(ctx, 3n, 5n);
+
+    const { chainId } = await ethers.provider.getNetwork();
+    // Sepolia-specific initialization
+    if (chainId === 11155111n) {
```
tamtamchik, line 54: I'd move this to the protocol discovery since this may be used somewhere else, and here, take data from the context or move it under the protocol context flag to explicitly set tests state for Sepolia network.

vp4242 (approved):
LGTM

PR #877, opened 2024-11-24
Add comments and update deposit module

0xtomoon:
Added explanation in functions in LidoLocator and SepoliaDepositAdapter

## Context

Added comments in some functions and updated deposit module

## Problem

No issues in the ticket


tamtamchik:
Outdated.

PR #882, opened 2024-11-27
feat: add proper upgrade to version 3 in Lido

folkyatina:
Proper upgrade to version 3 for Lido contract


contracts/0.4.24/Lido.sol
```diff
@@ -210,6 +210,7 @@ contract Lido is Versioned, StETHPermit, AragonApp {
     {
         _bootstrapInitialHolder();
         _initialize_v2(_lidoLocator, _eip712StETH);
```
tamtamchik: Can we remove _initialize_v2 altogether by moving it's logic to v3?

folkyatina: We can, but should we?

PR #885, opened 2024-11-28
fix: extract mint/burning to Lido

folkyatina:
it's easier to authenticate

### DIFF
--- a/contracts/0.4.24/Lido.sol
+++ b/contracts/0.4.24/Lido.sol
@@ -591,16 +591,41 @@ contract Lido is Versioned, StETHPermit, AragonApp {
         stakingRouter.deposit.value(depositsValue)(depositsCount, _stakingModuleId, _depositCalldata);
     }
 
+    /// @notice Mint stETH shares
+    /// @param _recipient recipient of the shares
+    /// @param _sharesAmount amount of shares to mint
+    /// @dev can be called only by accounting
+    function mintShares(address _recipient, uint256 _sharesAmount) public {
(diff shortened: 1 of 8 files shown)


failingtwice (approved):

contracts/0.4.24/Lido.sol
```diff
@@ -591,16 +591,41 @@ contract Lido is Versioned, StETHPermit, AragonApp {
...
+        // maybe TransferShare for cover burn and all events for withdrawal burn
+    }
+
     /// @notice Mint shares backed by external vaults
     ///
     /// @param _receiver Address to receive the minted shares
     /// @param _amountOfShares Amount of shares to mint
-    ///
-    /// @dev authentication goes through isMinter in StETH
+    /// @return stethAmount The amount of stETH minted
```
loga4, line 624: the function doesnt return `stethAmount`

contracts/0.4.24/Lido.sol
```diff
@@ -634,7 +657,9 @@ contract Lido is Versioned, StETHPermit, AragonApp {
 
         EXTERNAL_BALANCE_POSITION.setStorageUint256(extBalance - stethAmount);
 
-        burnShares(msg.sender, _amountOfShares);
+        _burnShares(msg.sender, _amountOfShares);
+
+        _emitTransferEvents(msg.sender, address(0), stethAmount, _amountOfShares);
```
loga4, line 662: We dont have events for `burnShares`, but we have them here - why?

PR #893, opened 2024-12-06
[SI] Vaults UX suggestions

DiRaiks:

## Context

- It would be very convenient to have a view on VaultHub that will allow us to get Vaults by the Owner's address
   - Why is it necessary - in theory, there can be a lot of Vaults. To display all the user's Vaults on the UI, we need to get all the Vaults and then make a request for each one to get the Owner's address
- View on maxMintableShares - external - to show how much stETH Vault allows to mint
- canMint(uint256 depositSize) - a view that will return how much you can mint from the deposit amount - an exact number.
  - Needed when calculating the amount and for sending the number to the mint() method
- canWithdraw() - similar to canMint
- Deposit ETH/WETH
- Withdrawal ETH/WETH
- Mint stETH/wstETH
- Burn + Permit

## Problem

## Solution

Your proposed solution


contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -117,6 +117,35 @@ abstract contract VaultHub is AccessControlEnumerableUpgradeable {
         return $.sockets[$.vaultIndex[IHubVault(_vault)]];
     }
 
+    /// @notice Returns all vaults owned by a given address
+    /// @param _owner Address of the owner
+    /// @return An array of vaults owned by the given address
+    function vaultsByOwner(address _owner) external view returns (IHubVault[] memory) {
```
DiRaiks: I'm not sure about the correctness of this method, but such a method will be very useful

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -194,13 +277,33 @@ contract Dashboard is AccessControlEnumerable {
...
     function burn(uint256 _tokens) external virtual onlyRole(DEFAULT_ADMIN_ROLE) {
         _burn(_tokens);
     }
 
+    /**
+     * @notice Burns stETH tokens from the sender backed by the vault using EIP-2612 Permit.
+     * @param _tokens Amount of tokens to burn
+     * @param _permit data required for the stETH.permit() method to set the allowance
+     */
+    function burnWithPermit(uint256 _tokens, PermitInput calldata _permit) external virtual onlyRole(DEFAULT_ADMIN_ROLE) {
```
Jeday: burnWsteth(WithPermit)

contracts/0.8.25/vaults/VaultHub.sol
Jeday: Needs coment that this is to be used as view only, not safe for contract intereaction because of cycle

Jeday: Maybe needs  `vaultsIndexesByOwner` and then this method around it.

contracts/0.8.25/vaults/Dashboard.sol
DiRaiks: I wasn't sure that burning wstETH works on the business logic of Vaults. But let's add it and check on the review

Jeday: like on WQ, you transfer wsteth to this contract balance, unwrap to steth and procced with burn

contracts/0.8.25/vaults/VaultHub.sol
DiRaiks: The thing is that Vaults are created without an owner, and then the owner and other roles are added. And it seems to me that storing another array in memory that will be updated frequently may be a bad decision.
If this is not the case, then it makes sense to have an additional owner field in the VaultSocket structure.

folkyatina: > Maybe needs `vaultsIndexesByOwner` and then this method around it.

We can't maintain the correctness of such index, because the owner can be changed in the vault, independently.

folkyatina: If the owner is the dashboard contract, how are you planning to use it?

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -126,6 +148,49 @@ contract Dashboard is AccessControlEnumerable {
...
+
+        return sharesMinted + sharesToMint > maxMintableShares ? maxMintableShares - sharesMinted : sharesToMint;
+    }
+
+    /**
+     * @notice Returns the amount of ether that can be withdrawn from the staking vault.
+     * @return The amount of ether that can be withdrawn.
+     */
+    function canWithdraw() external view returns (uint256) {
+        return address(stakingVault).balance - stakingVault.locked();
```
folkyatina:
```suggestion
        return Math256.min(address(stakingVault).balance, stakingVault.unlocked());
```

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -150,6 +215,15 @@ contract Dashboard is AccessControlEnumerable {
         _fund();
     }
 
+    /**
+     * @notice Funds the staking vault with wrapped ether. Approvals for the passed amounts should be done before.
+     * @param _wethAmount Amount of wrapped ether to fund the staking vault with
+     */
+    function fundByWeth(uint256 _wethAmount) external virtual onlyRole(DEFAULT_ADMIN_ROLE) {
```
folkyatina, line 240: Missing `transferFrom`?

contracts/0.8.25/vaults/Dashboard.sol
DiRaiks: Yes. I meant the Vault economy which takes into account the burning of tokens. However, I have already added

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -159,6 +233,15 @@ contract Dashboard is AccessControlEnumerable {
         _withdraw(_recipient, _ether);
     }
 
+    /**
+     * @notice Withdraws stETH tokens from the staking vault to wrapped ether. Approvals for the passed amounts should be done before.
+     * @param _tokens Amount of tokens to withdraw
+     */
+    function withdrawToWeth(uint256 _tokens) external virtual onlyRole(DEFAULT_ADMIN_ROLE) {
+        _withdraw(address(weth), _tokens);
```
folkyatina:
```suggestion
        _withdraw(address(this), _tokens);
```

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -159,6 +233,15 @@ contract Dashboard is AccessControlEnumerable {
         _withdraw(_recipient, _ether);
     }
 
+    /**
+     * @notice Withdraws stETH tokens from the staking vault to wrapped ether. Approvals for the passed amounts should be done before.
```
folkyatina: Sorry, didn't get what it should do. RN, ether is still on the dashboard contract. Should it be transferred to msg.sender or some other recepient?

DiRaiks: @folkyatina The idea is to send to the `fundByWeth` method the amount of WETH that needs to be "unwrap" to the ETH and then fund Vault
WETH withdraw method https://github.com/gnosis/canonical-weth/blob/master/contracts/WETH9.sol#L38

Now I understand that it seems that this code will not work correctly

contracts/0.8.25/vaults/Dashboard.sol
DiRaiks: added

contracts/0.8.25/vaults/Dashboard.sol
DiRaiks: Updated

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -194,13 +287,62 @@ contract Dashboard is AccessControlEnumerable {
...
+    /**
+     * @notice Burns wstETH tokens from the sender backed by the vault using EIP-2612 Permit.
+     * @param _tokens Amount of wstETH tokens to burn
+     * @param _wstETHPermit data required for the wstETH.permit() method to set the allowance
+     */
+    function burnWstETHWithPermit(uint256 _tokens, PermitInput calldata _wstETHPermit) external virtual onlyRole(DEFAULT_ADMIN_ROLE) {
+        wstETH.permit(msg.sender, address(this), _wstETHPermit.value, _wstETHPermit.deadline, _wstETHPermit.v, _wstETHPermit.r, _wstETHPermit.s);
+
+        wstETH.transferFrom(msg.sender, address(this), _tokens);
+        stETH.approve(address(wstETH), _tokens);
```
Jeday: You don't need approve for unwrap.

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -194,13 +287,62 @@ contract Dashboard is AccessControlEnumerable {
...
+        _burn(_tokens);
+    }
+
+    /**
+     * @notice Burns wstETH tokens from the sender backed by the vault using EIP-2612 Permit.
+     * @param _tokens Amount of wstETH tokens to burn
+     * @param _wstETHPermit data required for the wstETH.permit() method to set the allowance
+     */
+    function burnWstETHWithPermit(uint256 _tokens, PermitInput calldata _wstETHPermit) external virtual onlyRole(DEFAULT_ADMIN_ROLE) {
+        wstETH.permit(msg.sender, address(this), _wstETHPermit.value, _wstETHPermit.deadline, _wstETHPermit.v, _wstETHPermit.r, _wstETHPermit.s);
```
Jeday: Needs check for allowance to prevent ddos. See updated WQ contract.

Jeday (changes requested):

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -20,6 +40,9 @@ import {VaultHub} from "./VaultHub.sol";
  * TODO: need to add recover methods for ERC20, probably in a separate contract
  */
 contract Dashboard is AccessControlEnumerable {
+    /// @dev basis points base
+    uint256 internal constant BPS_BASE = 100_00;
```
tamtamchik:
```suggestion
    uint256 internal constant TOTAL_BASIS_POINTS = 100_00;
```
We appear to have a bias towards this. I see that `BPS_BASE` is used in VaultHub. Can we consistently use `TOTAL_BASIS_POINTS` everywhere? @folkyatina

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -122,8 +157,52 @@ contract Dashboard is AccessControlEnumerable {
...
+    function totalMintableShares() public view returns (uint256) {
+        return _totalMintableShares(stakingVault.valuation());
+    }
+
+    /**
+     * @notice Returns the maximum number of shares that can be minted with deposited ether.
+     * @param _ether the amount of ether to be funded
+     * @return the maximum number of stETH that can be minted by ether
+     */
+    function canMintShares(uint256 _ether) external view returns (uint256) {
```
tamtamchik: `canMintShares` seems to conflict with the requirement that the `_ether` parameter be passed. Perhaps it should be `calculateMintableShares` instead?

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -122,8 +157,52 @@ contract Dashboard is AccessControlEnumerable {
...
+
+        if (_totalShares < _sharesMinted) return 0;
+        return _totalShares - _sharesMinted;
+    }
+
+    /**
+     * @notice Returns the amount of ether that can be withdrawn from the staking vault.
+     * @return The amount of ether that can be withdrawn.
+     */
+    function canWithdraw() external view returns (uint256) {
```
tamtamchik: Maybe rename it to `calculateWithdrawableAmount` for clarity?

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -146,6 +225,20 @@ contract Dashboard is AccessControlEnumerable {
         _fund();
     }
 
+    /**
+     * @notice Funds the staking vault with wrapped ether. Approvals for the passed amounts should be done before.
+     * @param _wethAmount Amount of wrapped ether to fund the staking vault with
+     */
+    function fundByWeth(uint256 _wethAmount) external virtual onlyRole(DEFAULT_ADMIN_ROLE) {
+        require(weth.allowance(msg.sender, address(this)) >= _wethAmount, "ERC20: transfer amount exceeds allowance");
```
tamtamchik:
```suggestion
        if (weth.allowance(msg.sender, address(this)) < _wethAmount) revert("ERC20: transfer amount exceeds allowance");
```
To unify the style of the checks inside the contract.

contracts/0.8.25/vaults/Delegation.sol
```diff
@@ -98,8 +98,11 @@ contract Delegation is Dashboard, IReportReceiver {
     /**
      * @notice Constructor sets the stETH token address.
      * @param _stETH Address of the stETH token contract.
+     * @param _weth Address of the weth token contract.
+     * @param _wstETH Address of the wstETH token contract.
+     * @param _vaultHub Address of the vault hub contract.
```
tamtamchik: Why do we need `_vaultHub` here?

tamtamchik (commented):
🙄 Partial review... more to come

contracts/0.8.25/vaults/Dashboard.sol
tamtamchik: Or maybe `getMintableShares`

contracts/0.8.25/vaults/Dashboard.sol
tamtamchik: Or maybe `getWithdrawableEther`

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -274,6 +487,18 @@ contract Dashboard is AccessControlEnumerable {
         vaultHub.burnStethBackedByVault(address(stakingVault), _tokens);
     }
 
+    /**
+     * @dev calculates total shares vault can mint
+     * @param _valuation custom vault valuation
+     */
+    function _totalMintableShares(uint256 _valuation) internal view returns (uint256) {
```
tamtamchik, line 504: I think this must be replaced for _maxMintableShares from the VaultHub... maybe we heed to make it public in the hub
```
function _maxMintableShares(address _vault, uint256 _reserveRatio) internal view returns (uint256) {
        uint256 maxStETHMinted = (IStakingVault(_vault).valuation() * (TOTAL_BASIS_POINTS - _reserveRatio)) /
            TOTAL_BASIS_POINTS;
        return STETH.getSharesByPooledEth(maxStETHMinted);
    }
```

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -122,8 +157,52 @@ contract Dashboard is AccessControlEnumerable {
         return vaultSocket().treasuryFeeBP;
     }
 
+    function valuation() external view returns (uint256) {
+        return stakingVault.valuation();
+    }
+
+    /**
+     * @notice Returns the total of stETH shares that can be minted on the vault bound by valuation and vault share limit.
+     * @dev This is a public view method for the _maxMintableShares method in VaultHub
```
tamtamchik: But it is not... this is a separate method that duplicates the logic

contracts/0.8.25/vaults/Delegation.sol
Jeday: `claimManagementDue` uses it

contracts/0.8.25/vaults/Dashboard.sol
Jeday: fixed

contracts/0.8.25/vaults/Delegation.sol
tamtamchik: Yes, but it is set in the `initialize` via `vaultHub = VaultHub(stakingVault.vaultHub());`, and it is not assigned in the `Delegation` constructor anyways...

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -146,6 +229,20 @@ contract Dashboard is AccessControlEnumerable {
         _fund();
     }
 
+    /**
+     * @notice Funds the staking vault with wrapped ether. Approvals for the passed amounts should be done before.
+     * @param _wethAmount Amount of wrapped ether to fund the staking vault with
+     */
+    function fundByWeth(uint256 _wethAmount) external virtual onlyRole(DEFAULT_ADMIN_ROLE) {
```
tamtamchik, line 240:
```suggestion
    function fundWeth(uint256 _wethAmount) external virtual onlyRole(DEFAULT_ADMIN_ROLE) {
```

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -155,6 +252,17 @@ contract Dashboard is AccessControlEnumerable {
         _withdraw(_recipient, _ether);
     }
 
+    /**
+     * @notice Withdraws stETH tokens from the staking vault to wrapped ether.
+     * @param _recipient Address of the recipient
+     * @param _ether Amount of ether to withdraw
+     */
+    function withdrawToWeth(address _recipient, uint256 _ether) external virtual onlyRole(DEFAULT_ADMIN_ROLE) {
```
tamtamchik, line 264:
```suggestion
    function withdrawWeth(address _recipient, uint256 _ether) external virtual onlyRole(DEFAULT_ADMIN_ROLE) {
```

tamtamchik (approved):
Approved. 

However, there are a couple of naming suggestions to consider. 

PS. Please note that at the moment, tests are failing because `managementDue` is being removed in the target branch, causing the tests not to work. T_T

Jeday (approved):

PR #894, opened 2024-12-09
chore: fix tests in master

tamtamchik:
`Should deposit 100 ETH to node operators` test is broken on-chain, so here is a fix to make it green again.

Other:
- Some changes to Slither workflow to move configuration out of action.


rnmsslido (approved):
LGTM for changes to `.github/workflows`.
Before merging, please ensure additional reviews are obtained covering the other changes in this PR.

PR #896, opened 2024-12-11
feat: local devnet

eddort:
> PR is currently under development.

This PR adds the ability to deploy the protocol on a local devnet.
The main repository is located here: https://github.com/lidofinance/lido-local-devnet/pull/6.


hardhat.config.ts
```diff
@@ -57,7 +63,19 @@ const config: HardhatUserConfig = {
     },
   },
   etherscan: {
-    apiKey: process.env.ETHERSCAN_API_KEY || "",
```
tamtamchik, line 60: Looks like this change breaks verification on other than `local-devnet` networks.

deployed-holesky.json
```diff
@@ -175,7 +175,7 @@
   "app:simple-dvt": {
     "stakingRouterModuleParams": {
       "moduleName": "SimpleDVT",
-      "moduleType": "simple-dvt-onchain-v1",
+      "moduleType": "curated-onchain-v1",
```
tamtamchik, line 178: Can this be verified somehow?

tamtamchik (changes requested):
👏  Looks good, but it maybe breaks verification on non local-devnet networks

deployed-holesky.json
eddort: Of course, I'll give you a couple of examples now

Mainnet deployed: https://github.com/lidofinance/core/blob/9492d8f21ee15743a2e5a05cfc9012801498b121/deployed-mainnet.json#L149
Holesky: https://holesky.etherscan.io/address/0x11a93807078f8BB880c1BD0ee4C387537de4b4b6#readProxyContract#F2

We can get the type and check it with cast
```sh
cast --to-ascii 0x637572617465642d6f6e636861696e2d76310000000000000000000000000000
```

The type itself is the data needed for KAPI, it is basically the version of the parser. We have only two types curated and csm

hardhat.config.ts
eddort: fixed

tamtamchik (approved):
💚

scripts/scratch/steps/0120-initialize-non-aragon-contracts.ts
```diff
@@ -108,6 +109,10 @@ export async function main() {
     { from: deployer },
   );
 
+  // Initialize WithdrawalVault
+  const withdrawalVault = await loadContract("WithdrawalVault", withdrawalVaultAddress);
+  await makeTx(withdrawalVault, "initialize", [withdrawalVaultAdmin], { from: deployer });
```
eddort: Here is fix v0 -> v1
Etherscan examples:

https://etherscan.io/address/0xB9D7934878B5FB9610B3fE8A5e441e8fad7E293f#readProxyContract
https://holesky.etherscan.io/address/0xF0179dEC45a37423EAD4FaD5fCb136197872EAd9#readProxyContract

PR #904, opened 2024-12-20
feat: withdrawal credentials

mkurayan:
### Summary

This PR introduces an early-stage prototype for a potential implementation of [EIP-7685: General Purpose Execution Layer Requests](https://eips.ethereum.org/EIPS/eip-7685). Specifically, it implements [EIP-7002: Execution Layer Triggerable Withdrawals](https://eips.ethereum.org/EIPS/eip-7002) within the Lido WithdrawalVault contract, which serves as the withdrawal credentials address for Lido validators.

### Approach

The implementation follows the first approach outlined in the [ADR for Withdrawal Credentials Contract](https://hackmd.io/@lido/ryTE7SPXJl). This approach was selected due to the following advantages:
- **Clear and specific interfaces:** Each request type has a dedicated, well-defined interface.
- **Custom logic and parameter validation:** Request methods can validate parameters and incorporate tailored logic as needed.
- **Granularity and modularity:** WithdrawalCredentials contracts, such as the existing Lido WithdrawalVault and future Vaults, can expose only the necessary subset of request handlers. This allows for unique permission models and specialized logic.

### Implementation Details

In this iteration, the `WithdrawalVault` contract supports only full withdrawal requests. Partial withdrawals are not included because they are not part of the proposed [Triggerable Withdrawal V1](https://hackmd.io/@lido/SJKVeT990) implementation within the Lido protocol.

The following pseudo-code demonstrates the approach:

```solidity
library TriggerableWithdrawals {
    // Address of the EIP-7002 pre-deployed contract.
    address constant WITHDRAWAL_REQUEST = 0x0c15...;

    function addFullWithdrawalRequests(
        bytes[] calldata pubkeys,
        uint256 totalWithdrawalFee
    ) internal {
        // Implementation omitted for brevity.
        // WITHDRAWAL_REQUEST.call{value: fee}(withdrawalRequests);
    }
    
    function addPartialWithdrawalRequests(
        bytes[] calldata pubkeys,
        uint64[] calldata amounts,
        uint256 totalWithdrawalFee
    ) internal {
        // Implementation omitted for brevity.
        // WITHDRAWAL_REQUEST.call{value: fee}(withdrawalRequests);
    }
}

contract WithdrawalVault {
    function addFullWithdrawalRequests(
        bytes[] calldata pubkeys
    ) external payable {
        if (msg.sender != address(VALIDATORS_EXIT_BUS)) {
            revert NotValidatorExitBus();
        }

        TriggerableWithdrawals.addFullWithdrawalRequests(pubkeys, msg.value);
    }
}
```

### Key Considerations

The modular design of separate libraries for different withdrawal request types ensures that withdrawal credentials contracts can import and use only the minimal functionality required. For example:
- **Contract A:** Supports only full withdrawal requests.
- **Contract B:** Supports full withdrawals, partial withdrawals, and consolidation requests.

### Notes

This PR is an initial prototype and is subject to further iterations based on feedback and evolving requirements.


contracts/0.8.9/lib/WithdrawalRequests.sol
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * @dev Adds partial withdrawal requests for the provided public keys with corresponding amounts.
+     *      A partial withdrawal is any withdrawal where the amount is greater than zero.
+     *      This allows withdrawal of any balance exceeding 32 ETH (e.g., if a validator has 35 ETH, up to 3 ETH can be withdrawn).
+     *      However, the protocol enforces a minimum balance of 32 ETH per validator, even if a higher amount is requested.
+     * @param pubkeys An array of public keys for the validators requesting withdrawals.
+     * @param amounts An array of corresponding withdrawal amounts for each public key.
+     */
+    function addPartialWithdrawalRequests(
```
folkyatina: What if I want a batch that is mix of partial and full withdrawals?

contracts/0.8.9/lib/WithdrawalRequests.sol
```diff
@@ -0,0 +1,122 @@
...
+                revert InvalidPubkeyLength(pubkey);
+            }
+
+            uint256 feeToSend = feePerRequest;
+
+            if (i == keysCount - 1) {
+                feeToSend += unallocatedFee;
+            }
+
+            bytes memory callData = abi.encodePacked(pubkey, amount);
```
folkyatina: memory expansion in a loop can eat a lot of gas very quickly. Maybe worth to optimize like in BeaconChainDepositor.

contracts/0.8.9/WithdrawalVault.sol
```diff
@@ -27,6 +28,7 @@ contract WithdrawalVault is Versioned {
 
     ILido public immutable LIDO;
     address public immutable TREASURY;
+    address public immutable VALIDATORS_EXIT_BUS;
```
folkyatina: It may be better to store locator here

contracts/0.8.9/lib/WithdrawalRequests.sol
```diff
@@ -0,0 +1,122 @@
...
+    function _addWithdrawalRequests(
+        bytes[] calldata pubkeys,
+        uint64[] memory amounts
+    ) internal {
+        uint256 keysCount = pubkeys.length;
+        if (keysCount == 0) {
+            revert NoWithdrawalRequests();
+        }
+
+        uint256 minFeePerRequest = getWithdrawalRequestFee();
```
folkyatina: Doesn't fee increase with each request?

contracts/0.8.9/lib/WithdrawalRequests.sol
```diff
@@ -0,0 +1,122 @@
...
+            (bool success, ) = WITHDRAWAL_REQUEST.call{value: feeToSend}(callData);
+
+            if (!success) {
+                revert WithdrawalRequestAdditionFailed(pubkey, amount);
+            }
+
+            emit WithdrawalRequestAdded(pubkey, amount);
+        }
+
+        assert(address(this).balance == prevBalance);
```
folkyatina: Is it possible to calculate the fee exactly at the moment of signing the tx? Shouldn't we ease the constraint and add some refund capability here?

contracts/0.8.9/lib/WithdrawalRequests.sol
mkurayan: No, the fee will not increase with each request.  Inside the transaction, all requests will have the same fee.

EIP 7002 uses block-by-block behavior. 

If block N processes X requests, then at the end of block N  the number of withdrawal requests that the chain has processed relative to the “targeted” number increases by X - TARGET_WITHDRAWAL_REQUESTS_PER_BLOCK, and so the fee in block N+1 increases by a factor of e**((X - TARGET_WITHDRAWAL_REQUESTS_PER_BLOCK) / WITHDRAWAL_REQUEST_FEE_UPDATE_FRACTION).

contracts/0.8.9/lib/WithdrawalRequests.sol
mkurayan: Great question. 

Calls to the system contract require a fee payment defined by the current contract state. Overpaid fees are not returned to the caller. It is not generally possible to compute the exact required fee amount ahead of time. 

When adding a withdrawal request from a contract, the contract can perform a read operation to check for the current fee and then pay an exact amount.

When adding a withdrawal request from EOA request may result in overpayment of fees. But the gas cost of returning the overage would likely be higher than the overage itself. 

Proposed in this PR implementation does not imply any restriction on the caller, but makes sure that the withdrawal credential balance (accounting) will not be affected.

In the proposed [Triggerable Withdrawal](https://hackmd.io/@lido/SJKVeT990) implementation, the dedicated bot will provide the fee within the withdrawal request transaction and, a small overpayment is possible. This is the simplest approach, which is currently considered for [Triggerable Withdrawal](https://hackmd.io/@lido/SJKVeT990)  implementation.

Other withdrawal contracts may consider the first approach without overpayment.

contracts/0.8.9/lib/WithdrawalRequests.sol
mkurayan: The separation of Full and Partial withdrawals was suggested to simplify an interaction with the withdrawal contract.

An approach, proposed in the [Triggerable Withdravals](https://github.com/lidofinance/core/pull/9040) document suggests that the Validator Exit Bus will only use full withdrawal. Usages of partial withdrawals might require a different mechanism inside the Lido protocol, which is why this PR proposes to separate those types of requests.

If you consider that a mix of partial and full withdrawals might be useful in the future we can consider multiple approaches. 

1. Add the `addWithdravalCredentials` method which will consume both partial and full withdrawals.

2. In addition to the first approach, remove `addPartialWithdrawalRequests` method, so the `addWithdravalCredentials` will be used for both partials and mix (partial + full) withdrawal requests.

What do you think will be an optimal approach?

folkyatina: 1st one will be fine. I think that this kind of constraint is not required in vaults.

contracts/0.8.9/lib/WithdrawalRequests.sol
folkyatina: For vaults, it will be good to exclude `msg.value` or fees' amount assumptions from this method. I mean, it's ok that we can't foresee the price, but in the same time, I'd like that the excessive amount just stay on the vault if any and will be available for withdraw later. 

So, I'd rather just check if the balance is enough to add all requests and does not impose any assumptions on if the balance was topped up in this tx or before.

mkurayan: Good point, I agree. The fee allocation strategy should be decoupled from the withdrawals library, allowing each contract to determine its own approach based on its design and requirements. To address these changes, I suggest adding a new `totalWithdrawalFee` parameter, which represents the total amount that will be spent on the EIP-7002 withdrawal requests.

The following pseudo-code demonstrates the approach:
```solidity
library WithdrawalRequests {
    function addWithdrawalRequests(
        bytes[] calldata pubkeys,
        uint64[] memory amounts,
        uint256 totalWithdrawalFee
    ) internal {
        // Implementation that uses totalWithdrawalFee to cover EIP-7002 withdrawal requests
        ...
    }

    function getWithdrawalRequestFee() internal pure returns (uint256) {
        // Returns the minimum required fee per request
        ...
    }
}

contract A {
    function addWithdrawalRequests(
        bytes[] calldata pubkeys,
        uint64[] calldata amounts
    ) external payable {
        // Use the entire sent amount (msg.value) as the total fee for withdrawal requests
        uint256 totalWithdrawalFee = msg.value;
        WithdrawalRequests.addWithdrawalRequests(pubkeys, amounts, totalWithdrawalFee);
    }
}

contract B {
    function addWithdrawalRequests(
        bytes[] calldata pubkeys,
        uint64[] calldata amounts
    ) external {
        // Use the minimum required fee per request
        uint256 minFeePerRequest = WithdrawalRequests.getWithdrawalRequestFee();
        uint256 totalWithdrawalFee = minFeePerRequest * pubkeys.length;
        WithdrawalRequests.addWithdrawalRequests(pubkeys, amounts, totalWithdrawalFee);
    }
}

// Additional contracts can implement other fee allocation strategies as needed.
```

contracts/0.8.9/lib/WithdrawalRequests.sol
mkurayan: Yes, I am also considering the first approach, noted, thanks!

contracts/0.8.9/lib/TriggerableWithdrawals.sol
```diff
@@ -0,0 +1,148 @@
...
+    event WithdrawalRequestAdded(bytes pubkey, uint256 amount);
+
+    /**
+     * @dev Adds full withdrawal requests for the provided public keys.
+     *      The validator will fully withdraw and exit its duties as a validator.
+     * @param pubkeys An array of public keys for the validators requesting full withdrawals.
+     */
+    function addFullWithdrawalRequests(
+        bytes[] calldata pubkeys,
+        uint256 totalWithdrawalFee
```
folkyatina: It think that this parameter is not required at all. We can safely assume that all required ether is on the balance of the contract and we'll revert if it's not true and if we need some additional constraints (like, msg.value == fee), we can add it in the contract that use that lib.

contracts/0.8.9/lib/TriggerableWithdrawals.sol
```diff
@@ -0,0 +1,148 @@
...
+        if (minFeePerRequest * keysCount > totalWithdrawalFee) {
+            revert FeeNotEnough(minFeePerRequest, keysCount, totalWithdrawalFee);
+        }
+
+        uint256 feePerRequest = totalWithdrawalFee / keysCount;
+        uint256 unallocatedFee = totalWithdrawalFee % keysCount;
+        uint256 prevBalance = address(this).balance - totalWithdrawalFee;
+
+        for (uint256 i = 0; i < keysCount; ++i) {
+            bytes memory pubkey = pubkeys[i];
```
folkyatina: Unnecessary memory extension?

contracts/0.8.9/lib/TriggerableWithdrawals.sol
mkurayan: This parameter was introduced to decouple the fee allocation strategy from the withdrawals library, [discussion](https://github.com/lidofinance/core/pull/904#discussion_r1894111883). It enables contracts to employ different allocation strategies.

The proposed **Validator Exitt Bus Triggerable Withdrawal** implementation assumes that the withdrawal fee is provided by the actor who triggers the withdrawals. In this approach, the `WithdrawalVault.sol` balance remains unaffected, preventing issues with the Oracle’s accounting. Consequently, the entire `msg.value` sent is used as the fee for withdrawal requests:

```solidity
// Simplified pseudo-code
function addWithdrawalRequests(
    bytes[] calldata pubkeys,
    uint64[] calldata amounts
) external payable {
    // Use the entire sent amount (msg.value) as the total fee for withdrawal requests
    uint256 totalWithdrawalFee = msg.value;
    WithdrawalRequests.addWithdrawalRequests(pubkeys, amounts, totalWithdrawalFee);
}
```

Other vaults could employ the strategy you mentioned, assuming all required Ether is already in the contract’s balance:

```solidity
// Simplified pseudo-code
function addWithdrawalRequests(
    bytes[] calldata pubkeys,
    uint64[] calldata amounts
) external {
    // Use the minimum required fee per request
    uint256 minFeePerRequest = WithdrawalRequests.getWithdrawalRequestFee();
    uint256 totalWithdrawalFee = minFeePerRequest * pubkeys.length;
    WithdrawalRequests.addWithdrawalRequests(pubkeys, amounts, totalWithdrawalFee);
}
```

When the withdrawal fee is specified explicitly, any fee allocation strategy can be used. The library ensures that the provided fee sufficiently covers all requests and that the exact fee amount is spent.

contracts/0.8.9/lib/WithdrawalRequests.sol
mkurayan: Done. Here is some additional discussion on this topic: https://github.com/lidofinance/core/pull/904#discussion_r1901709888

contracts/0.8.9/lib/TriggerableWithdrawals.sol
mkurayan: Yes, we can access the pubkeys and amounts directly but there are nine places in the code below where I have to write `pubkeys[i]` or `amounts[i]`, so for the code clarity pubkeys[i] and amounts[i] were extracted into the dedicated variables.

contracts/0.8.9/WithdrawalVault.sol
mkurayan: Agree, updated.

contracts/0.8.9/lib/TriggerableWithdrawals.sol
folkyatina: > When the withdrawal fee is specified explicitly, any fee allocation strategy can be used. The library ensures that the provided fee sufficiently covers all requests and that the exact fee amount is spent.

See no reason to pass it inside the function when it can definitely be checked before and after the function call in the WithdrawalVault itself. It's kinda alien constraint for the raw withdrawal request creation library. 

E.g. in the vaults we don't care where the funds for the gas will come from and we don't need to check it at all.

contracts/0.8.9/lib/TriggerableWithdrawals.sol
folkyatina: I understand this clearly, but Solidity does not have any implicit mechanic to reuse the memory, and here we have the code that unnecessarily allocates new memory and copies values from calldata into it, which scales with the keysCount in a loop, and that's quite inefficient. 

I suspect that the clarity isn't worth such a performance loss, especially, if we want to have the general purpose library.

contracts/0.8.9/lib/TriggerableWithdrawals.sol
```diff
@@ -0,0 +1,148 @@
...
+        uint256 keysCount = pubkeys.length;
+        if (keysCount == 0) {
+            revert NoWithdrawalRequests();
+        }
+
+        if(address(this).balance < totalWithdrawalFee) {
+            revert InsufficientBalance(address(this).balance, totalWithdrawalFee);
+        }
+
+        uint256 minFeePerRequest = getWithdrawalRequestFee();
```
folkyatina:
```suggestion
        uint256 feePerRequest = getWithdrawalRequestFee();
```

folkyatina (changes requested):
Still see some potential for improvement

contracts/0.8.9/lib/TriggerableWithdrawals.sol
mkurayan: The EIP-7002 specification does not impose an upper limit on the withdrawal request fee; this was intentionally designed for flexibility. The library implementation proposed in this PR follows the EIP-7002 specification and does not add any extra restrictions on withdrawal requests.

If we do not want the **general purpose** library, we can remove control over the request fee, and narrow the library's functionality to allow only requests with minimal fees. This would simplify the code slightly, but also limit the library's potential use cases.

I am not confident that we will not need control over the request fee in the future, @folkyatina what do you think?

contracts/0.8.9/lib/TriggerableWithdrawals.sol
mkurayan: I agree that for a **general purpose** library, we should optimize gas consumption. I will update the code accordingly. Thanks for pointing that out! 👍

mkurayan: Done.

contracts/0.8.9/lib/TriggerableWithdrawals.sol
mkurayan: After an internal discussion, we agreed to follow the eip 7002 specification and keep control over the request fee amount, but pass the request fee instead of the total withdrawal fee to simplify library implementation.

contracts/0.8.9/WithdrawalVault.sol
```diff
@@ -42,34 +47,44 @@ contract WithdrawalVault is Versioned {
...
      */
-    constructor(ILido _lido, address _treasury) {
-        if (address(_lido) == address(0)) {
-            revert LidoZeroAddress();
-        }
-        if (_treasury == address(0)) {
-            revert TreasuryZeroAddress();
-        }
+    constructor(address _lido, address _treasury) {
+        _requireNonZero(_lido);
```
tamtamchik:
```suggestion
        _onlyNonZeroAddress(_lido);
```

contracts/0.8.9/lib/WithdrawalRequests.sol
mkurayan: I agree, code optimized.

contracts/0.8.9/lib/TriggerableWithdrawals.sol
```diff
@@ -0,0 +1,163 @@
+// SPDX-FileCopyrightText: 2023 Lido <info@lido.fi>
```
tamtamchik:
```suggestion
// SPDX-FileCopyrightText: 2025 Lido <info@lido.fi>
```

contracts/0.8.9/lib/TriggerableWithdrawals.sol
```diff
@@ -0,0 +1,163 @@
...
+// SPDX-License-Identifier: GPL-3.0
+
+pragma solidity 0.8.9;
+library TriggerableWithdrawals {
+    address constant WITHDRAWAL_REQUEST = 0x0c15F14308530b7CDB8460094BbB9cC28b9AaaAA;
```
tamtamchik: Add some comment with "validation" link? 
```suggestion
    /// @dev https://eips.ethereum.org/EIPS/eip-7002#configuration
    address constant WITHDRAWAL_REQUEST = 0x0c15F14308530b7CDB8460094BbB9cC28b9AaaAA;
```

tamtamchik: May this contract address be changed on testnets to something else (as, for example, with the deposit contract on Holesky)? We'll have to use a separate library for it?

contracts/0.8.9/lib/TriggerableWithdrawals.sol
```diff
@@ -0,0 +1,163 @@
+// SPDX-FileCopyrightText: 2023 Lido <info@lido.fi>
+// SPDX-License-Identifier: GPL-3.0
+
+pragma solidity 0.8.9;
```
tamtamchik:
```suggestion
// solhint-disable-next-line lido/fixed-compiler-version
pragma solidity >=0.8.9 <0.9.0;
```
Also, consider moving to common libs?

contracts/0.8.9/lib/TriggerableWithdrawals.sol
```diff
@@ -0,0 +1,163 @@
...
+
+        for (uint256 i = 0; i < keysCount; i++) {
+            _copyPubkeyToMemory(pubkeys, callData, i);
+
+            (bool success, ) = WITHDRAWAL_REQUEST.call{value: feePerRequest}(callData);
+
+            if (!success) {
+                revert WithdrawalRequestAdditionFailed(callData);
+            }
+        }
```
tamtamchik: `emit WithdrawalRequestAdded(...)` ? Or is this done on the WITHDRAWAL_REQUEST contract side, or should it be done on the side of the contract that uses this library?

contracts/0.8.9/WithdrawalVault.sol
```diff
@@ -9,6 +9,9 @@ import "@openzeppelin/contracts-v4.4/token/ERC721/IERC721.sol";
 import "@openzeppelin/contracts-v4.4/token/ERC20/utils/SafeERC20.sol";
 
 import {Versioned} from "./utils/Versioned.sol";
+import {AccessControlEnumerable} from "./utils/access/AccessControlEnumerable.sol";
+import {TriggerableWithdrawals} from "./lib/TriggerableWithdrawals.sol";
+import { ILidoLocator } from "../common/interfaces/ILidoLocator.sol";
```
tamtamchik:
```suggestion
import {ILidoLocator} from "../common/interfaces/ILidoLocator.sol";
```

contracts/0.8.9/WithdrawalVault.sol
```diff
@@ -122,4 +137,52 @@ contract WithdrawalVault is Versioned {
...
+     */
+    function addFullWithdrawalRequests(
+        bytes calldata pubkeys
+    ) external payable onlyRole(ADD_FULL_WITHDRAWAL_REQUEST_ROLE) {
+        uint256 prevBalance = address(this).balance - msg.value;
+
+        uint256 minFeePerRequest = TriggerableWithdrawals.getWithdrawalRequestFee();
+        uint256 totalFee = pubkeys.length / TriggerableWithdrawals.PUBLIC_KEY_LENGTH  * minFeePerRequest;
+
+        if(totalFee > msg.value) {
```
tamtamchik:
```suggestion
        if (totalFee > msg.value) {
```

contracts/0.8.9/lib/TriggerableWithdrawals.sol
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    function _copyAmountToMemory(bytes memory target, uint64 amount) private pure {
+        assembly {
+            mstore(add(target, 80), shl(192, amount))
+        }
+    }
+
+    function _validateAndCountPubkeys(bytes calldata pubkeys) private pure returns (uint256) {
+        if(pubkeys.length % PUBLIC_KEY_LENGTH != 0) {
```
tamtamchik:
```suggestion
        if (pubkeys.length % PUBLIC_KEY_LENGTH != 0) {
```

contracts/0.8.9/lib/TriggerableWithdrawals.sol
```diff
@@ -0,0 +1,163 @@
...
+
+        if (feePerRequest == 0) {
+            feePerRequest = minFeePerRequest;
+        }
+
+        if (feePerRequest < minFeePerRequest) {
+            revert InsufficientRequestFee(feePerRequest, minFeePerRequest);
+        }
+
+        if(address(this).balance < feePerRequest * keysCount) {
```
tamtamchik:
```suggestion
        if (address(this).balance < feePerRequest * keysCount) {
```

contracts/0.8.9/WithdrawalVault.sol
```diff
@@ -122,4 +137,52 @@ contract WithdrawalVault is Versioned {
...
+     *      The validator will fully withdraw and exit its duties as a validator.
+     * @param pubkeys An array of public keys for the validators requesting full withdrawals.
+     */
+    function addFullWithdrawalRequests(
+        bytes calldata pubkeys
+    ) external payable onlyRole(ADD_FULL_WITHDRAWAL_REQUEST_ROLE) {
+        uint256 prevBalance = address(this).balance - msg.value;
+
+        uint256 minFeePerRequest = TriggerableWithdrawals.getWithdrawalRequestFee();
+        uint256 totalFee = pubkeys.length / TriggerableWithdrawals.PUBLIC_KEY_LENGTH  * minFeePerRequest;
```
tamtamchik:
```suggestion
        uint256 totalFee = pubkeys.length / TriggerableWithdrawals.PUBLIC_KEY_LENGTH * minFeePerRequest;
```

tamtamchik (commented):
Looks good! Minor comments about library placement, ~~constant WITHDRAWAL_REQUEST (may be immutable)~~, and some style fixes suggestions.

contracts/0.8.9/WithdrawalVault.sol
mkurayan: Thanks.

contracts/0.8.9/lib/TriggerableWithdrawals.sol
mkurayan: Thanks.

contracts/0.8.9/lib/TriggerableWithdrawals.sol
mkurayan: Thanks.

contracts/0.8.9/lib/TriggerableWithdrawals.sol
mkurayan: The library provides minimum functionality for submitting Triggerable Withdrawal requests, which does not include event emitting if necessary events emitting might be done separately in contracts that use `TriggerableWithdrawals` library.  

In the case of the proposed TW + VEB implementation, the kind of events that will be needed has not yet been defined.

contracts/0.8.9/lib/TriggerableWithdrawals.sol
mkurayan: Good point! Lib will be moved to common.

mkurayan: Done.

contracts/0.8.9/lib/TriggerableWithdrawals.sol
mkurayan: Updated, thanks!

contracts/0.8.9/lib/TriggerableWithdrawals.sol
tamtamchik: The contract will be pre-deployed (as per proposal), so the address will remain the same.

contracts/0.8.9/WithdrawalVault.sol
mkurayan: Thanks, applied 👍

contracts/common/lib/TriggerableWithdrawals.sol
```diff
@@ -0,0 +1,154 @@
...
+// solhint-disable-next-line lido/fixed-compiler-version
+pragma solidity >=0.8.9 <0.9.0;
+library TriggerableWithdrawals {
+    address constant WITHDRAWAL_REQUEST = 0x0c15F14308530b7CDB8460094BbB9cC28b9AaaAA;
+    uint256 internal constant WITHDRAWAL_REQUEST_CALLDATA_LENGTH = 56;
+    uint256 internal constant PUBLIC_KEY_LENGTH = 48;
+    uint256 internal constant WITHDRAWAL_AMOUNT_LENGTH = 8;
+
+    error MismatchedArrayLengths(uint256 keysCount, uint256 amountsCount);
+    error InsufficientBalance(uint256 balance, uint256 totalWithdrawalFee);
```
tamtamchik: Too wide error naming may clash (and it does within vaults), so maybe it's worth narrowing it a bit more.

```suggestion
    error InsufficientBalanceForWithdrawalFee(uint256 balance, uint256 totalWithdrawalFee);
```

contracts/common/lib/TriggerableWithdrawals.sol
```diff
@@ -0,0 +1,154 @@
...
+        if (feePerRequest == 0) {
+            feePerRequest = minFeePerRequest;
+        }
+
+        if (feePerRequest < minFeePerRequest) {
+            revert InsufficientRequestFee(feePerRequest, minFeePerRequest);
+        }
+
+        if (address(this).balance < feePerRequest * keysCount) {
+            revert InsufficientBalance(address(this).balance, feePerRequest * keysCount);
```
tamtamchik:
```suggestion
            revert InsufficientBalanceForWithdrawalFee(address(this).balance, feePerRequest * keysCount);
```

tamtamchik (commented):
Added some suggestions to errors naming, to prevent clashing

contracts/0.8.9/WithdrawalVault.sol
```diff
@@ -122,4 +141,66 @@ contract WithdrawalVault is Versioned {
...
+        uint256 refund = msg.value - totalFee;
+        if (refund > 0) {
+            (bool success, ) = msg.sender.call{value: refund}("");
+
+            if (!success) {
+                revert TriggerableWithdrawalRefundFailed();
+            }
+        }
+
+        assert(address(this).balance == prevBalance);
```
dgusakov, line 187: Can you return the refund amount for the ease of integration?

contracts/0.8.9/WithdrawalVault.sol
```diff
@@ -122,4 +141,66 @@ contract WithdrawalVault is Versioned {
...
+     *         - The provided total withdrawal fee is insufficient to cover all requests.
+     *         - Refund of the excess fee fails.
+     */
+    function addFullWithdrawalRequests(
+        bytes calldata pubkeys
+    ) external payable onlyRole(ADD_FULL_WITHDRAWAL_REQUEST_ROLE) {
+        uint256 prevBalance = address(this).balance - msg.value;
+
+        uint256 minFeePerRequest = TriggerableWithdrawals.getWithdrawalRequestFee();
+        uint256 totalFee = (pubkeys.length / TriggerableWithdrawals.PUBLIC_KEY_LENGTH) * minFeePerRequest;
```
dgusakov, line 166: Can we make it a separate view func so that integrating contracts can re-use it?

contracts/0.8.9/WithdrawalVault.sol
mkurayan: Yes we return the refund amount.

contracts/0.8.9/WithdrawalVault.sol
folkyatina, line 1: Since, it has no storage, we can safely upgrade solidity to 0.8.25 here, I think.

PR #911, opened 2024-12-27
rought draft: BRING YOUR OWN DSM (for permissiones staking vaults) [very very WIP]

failingtwice:
This PR includes a simple way to validate owner's validation of deposit data (Similary mechanism to DSM but with owner of the vault acting as the sole guardian)

### DIFF
--- a/contracts/0.8.25/lib/DepositLogistics.sol
+++ b/contracts/0.8.25/lib/DepositLogistics.sol
@@ -37,7 +37,7 @@ library DepositLogistics {
     /**
      * @notice Domain separator for the security signature
      */
-    bytes32 internal constant GUARDIAN_SIGNATURE_PREFIX = keccak256(abi.encodePacked("  "));
+    bytes32 internal constant GUARDIAN_SIGNATURE_PREFIX = keccak256(abi.encodePacked("GuardianSignaturePrefix"));
 
     /**
      * @notice Error thrown when the number of deposits is zero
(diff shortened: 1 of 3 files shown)


contracts/0.8.25/lib/DepositLogistics.sol
```diff
@@ -84,8 +105,26 @@ library DepositLogistics {
...
+        // i.e. no deposit front-running.
+
+        // the guardians signs a concatenation of 3 things:
+        // 1. deposit prefix (a constant that reduces the signature validity space)
+        // 2. the deposit root against which the deposit data was validated
+        // 3. the aggregate of all deposit data roots
+
+        bytes32 currentDepositRoot = _depositContract.get_deposit_root();
+        bytes32 securityMessageHash = keccak256(abi.encodePacked(GUARDIAN_SIGNATURE_PREFIX, currentDepositRoot, depositDataRoots));
+        address signer = ECDSA.recover(securityMessageHash, _guardianSignature.r, _guardianSignature.vs);
```
folkyatina, line 125: Looks like it's worth to push this implementation a bit further over the abstraction path. 

We can support ERC-1271 here and thus the other contract can be a guardian. 
We can have an implementation of the "Lido Vaults DSM" as the guardian contract, so we can interchange them.  
So, we can have some vaults protected by EOA guardian and some — by DSM committee. 

WDYT?

folkyatina (changes requested):
ERC-1271 maybe?

failingtwice:
Closed in favor of https://github.com/lidofinance/core/pull/930

PR #913, opened 2025-01-02
fix: get rid of double division on shares to eth conversion

folkyatina:
External shares concept introduce double division during the shares to eth conversion calculations and vise versa. 

The suggested way to fix it is to use internal ether and internal shares for conversion in `getPooledEthForShares()` and `getSharesForPoooledEther()`

### DIFF
--- a/contracts/0.4.24/Lido.sol
+++ b/contracts/0.4.24/Lido.sol
@@ -196,7 +196,7 @@ contract Lido is Versioned, StETHPermit, AragonApp {
      * @param _eip712StETH eip712 helper contract for StETH
      */
     function initialize(address _lidoLocator, address _eip712StETH) public payable onlyInit {
-        _bootstrapInitialHolder();
+        _bootstrapInitialHolder(); // stone in the elevator
 
         LIDO_LOCATOR_POSITION.setStorageAddress(_lidoLocator);
         emit LidoLocatorSet(_lidoLocator);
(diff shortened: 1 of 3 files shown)


tamtamchik (approved):

contracts/0.4.24/Lido.sol
```diff
@@ -947,6 +947,21 @@ contract Lido is Versioned, StETHPermit, AragonApp {
...
+    /// - avoid double precision loss on additional division on external ether calculations
+    /// - optimize gas cost of conversions between shares and ether
+    function _getShareRateNumerator() internal view returns (uint256) {
+        return _getInternalEther();
+    }
+
+    /// @dev the denominator (in shares) of the share rate for StETH conversion between shares and ether and vice versa.
+    function _getShareRateDenominator() internal view returns (uint256) {
+        uint256 externalShares = EXTERNAL_SHARES_POSITION.getStorageUint256();
+        uint256 internalShares = _getTotalShares() - externalShares;
```
failingtwice: This line looks a bit dangerous for an outside reader because if `_getTotalShares() == externalShares`, the denominator would be 0 and it would lead to division by 0 error.

Consider adding a comment about the limit on externalShares mentioning that `totalShares` **WILL ALWAYS BE GREATER** than `externalShares` and, thus, the denominator will never be 0.

contracts/0.4.24/StETH.sol
```diff
@@ -303,17 +303,17 @@ contract StETH is IERC20, Pausable {
      */
     function getSharesByPooledEth(uint256 _ethAmount) public view returns (uint256) {
         return _ethAmount
-            .mul(_getTotalShares())
-            .div(_getTotalPooledEther());
+            .mul(_getShareRateDenominator()) // denominator in shares
+            .div(_getShareRateNumerator()); // numerator in ether
```
failingtwice, line 307: This naming pattern with numerator/denominator is confusing because here `_getShareRateNumerator` is used as a denominator

folkyatina: See no real alternative. And, btw, it's still a numerator for share rate here.

contracts/0.4.24/Lido.sol
folkyatina: added a comment

PR #920, opened 2025-01-14
Feat/factory upgrade immutable

loga4:
feat: add immutable args for Clones


package.json
```diff
@@ -111,6 +111,7 @@
     "@openzeppelin/contracts": "3.4.0",
     "@openzeppelin/contracts-v4.4": "npm:@openzeppelin/contracts@4.4.1",
     "@openzeppelin/contracts-v5.0.2": "npm:@openzeppelin/contracts@5.0.2",
```
folkyatina: I believe, we don't need two v5 OZ libs for sure. Need to replace 5.0.2 with 5.2 and call it v5 or at least v5.2 to be able to update without changing imports all over the place

folkyatina (changes requested):
Looks like we have to get rid of OZ v5.0.2 in this PR

package.json
loga4: done

tamtamchik (approved):

PR #924, opened 2025-01-16
chore: accounting fixes after review

folkyatina:


tamtamchik (approved):
Looks good. `vaultsInOutDeltas` is used in AO, so DevNet 3 need to update the oracle code.

PR #925, opened 2025-01-17
[VAULTS] Simplify beacon chain deposit processing

failingtwice:
This PR gets rid of intense gas optimization for deposits in favor of simplicity and readability thanks to the Pectra's MAX_EB increase.


contracts/0.8.25/interfaces/IDepositContract.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-FileCopyrightText: 2024 Lido <info@lido.fi>
```
tamtamchik:
```suggestion
// SPDX-FileCopyrightText: 2025 Lido <info@lido.fi>
```

contracts/0.8.25/vaults/StakingVault.sol
```diff
@@ -304,22 +321,26 @@ contract StakingVault is IStakingVault, IBeaconProxy, BeaconChainDepositLogistic
...
+                deposit.pubkey,
+                bytes.concat(withdrawalCredentials()),
+                deposit.signature,
+                deposit.depositDataRoot
+            );
+        }
 
-        _makeBeaconChainDeposits32ETH(_numberOfDeposits, bytes.concat(withdrawalCredentials()), _pubkeys, _signatures);
-        emit DepositedToBeaconChain(msg.sender, _numberOfDeposits, _numberOfDeposits * 32 ether);
+        emit DepositedToBeaconChain(msg.sender, numberOfDeposits);
```
tamtamchik: Wouldn't a total deposited amount be needed for analytics?

test/integration/vaults-happy-path.integration.ts
```diff
@@ -460,3 +475,27 @@ describe("Scenario: Staking Vaults Happy Path", () => {
     expect(await stakingVault.locked()).to.equal(0);
   });
 });
+
+function getRoot(creds: string, pubkey: string, signature: string, size: bigint) {
```
tamtamchik: Duplication, move to lib?

tamtamchik (commented):
Looks great!

contracts/0.8.25/vaults/StakingVault.sol
failingtwice: I've decided not to include the total amount because each individual amount is included in the DepositContract's `DepositEvent`

tamtamchik: I'm not an expert, but wouldn't it be more convenient for analytics / UI to receive it as only one event instead of many? Just from this angle.

contracts/0.8.25/vaults/interfaces/IStakingVault.sol
```diff
@@ -20,9 +20,17 @@ interface IStakingVault {
         int128 inOutDelta;
     }
 
+    struct Deposit {
+        bytes pubkey;
+        bytes signature;
+        uint256 amount;
+        bytes32 depositDataRoot;
```
tamtamchik, line 27: Since depositDataRoot is now not calculated, maybe need some info from where to take it? 
Will it be taken from the deposit_data file?

folkyatina: It's probably even worth to include a view function that calculates the deposit_data_root to be conveniently used by external parties

failingtwice: It would requrie a ton of code (memory lib, helpers like little-endian conversion, root computation itself) to be added to this contract just to be wrapped in an external function.

contracts/0.8.25/vaults/StakingVault.sol
```diff
@@ -370,6 +390,57 @@ contract StakingVault is IStakingVault, BeaconChainDepositLogistics, OwnableUpgr
...
+
+        // Step 3. Compute the root of the pubkey
+        bytes32 pubkeyRoot = sha256(abi.encodePacked(_pubkey, bytes16(0)));
+
+        // Step 4. Compute the root of the signature
+        bytes32 sigSlice1Root = sha256(abi.encodePacked(_signature[0:64]));
+        bytes32 sigSlice2Root = sha256(abi.encodePacked(_signature[64:], bytes32(0)));
+        bytes32 signatureRoot = sha256(abi.encodePacked(sigSlice1Root, sigSlice2Root));
+
+        // Step 5. Compute the root-toot-toorootoo of the deposit data
```
tamtamchik, line 433:
```suggestion
        // Step 5. Compute the root-toot-totoro of the deposit data
```

tamtamchik (approved):
🌱

PR #927, opened 2025-01-21
feat: triggerable withdrawals vebo part (val-1456)

Amuhar:

## Context

## Problem

## Solution

Your proposed solution


contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -0,0 +1,115 @@
...
+
+        uint256 lastDeliveredKeyIndex = requestStatus.deliveredItemsCount - 1;
+
+        bytes memory pubkeys = new bytes(keyIndexes.length * PUBLIC_KEY_LENGTH);
+
+        for (uint256 i = 0; i < keyIndexes.length; i++) {
+            if (keyIndexes[i] > lastDeliveredKeyIndex) {
+                revert KeyWasNotUnpacked(keyIndexes[i], lastDeliveredKeyIndex);
+            }
+            uint256 requestOffset = keyIndexes[i] * PACKED_REQUEST_LENGTH + 16;
```
mkurayan: A minor suggestion to check `requestOffset` is not out of range.

```
        if (requestOffset + PUBLIC_KEY_LENGTH > data.length) {
            revert OutOfRangeDataAccess();
        }
```

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -0,0 +1,115 @@
...
+
+            for (uint256 j = 0; j < PUBLIC_KEY_LENGTH; j++) {
```
mkurayan: You can optimize coping with calldatacopy, like this.

 ```
        uint256 destOffset = i * PUBLIC_KEY_LENGTH;

        assembly {
            let dest := add(pubkeys, add(32, destOffset))
            calldatacopy(dest, add(data.offset, requestOffset), PUBLIC_KEY_LENGTH)
        }
```

File: contracts/0.8.9/oracle/ValidatorsExitBus.sol [hunk 3]
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    function triggerExitHashVerify(bytes calldata data, uint256[] calldata keyIndexes) external payable {
+        bytes32 dataHash = keccak256(abi.encode(data));
+        RequestStatus storage requestStatus = _storageExitRequestsHashes()[dataHash];
+
+        address locatorAddr = LOCATOR_CONTRACT_POSITION.getStorageAddress();
+        address withdrawalVaultAddr = ILidoLocator(locatorAddr).withdrawalVault();
+        uint256 minFee = IWithdrawalVault(withdrawalVaultAddr).getWithdrawalRequestFee();
+        uint requestsFee = keyIndexes.length * minFee;
```
mkurayan: better to use uint256 instead of uint

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -0,0 +1,115 @@
...
+    // TODO: make type optimization
+    struct DeliveryHistory {
+      uint256 blockNumber;
+      /// @dev Key index in exit request array
+      uint256 lastDeliveredKeyIndex;
+
+      // TODO: timestamp
+    }
+    // TODO: make type optimization
+    struct RequestStatus {
```
mkurayan, line 41: Should we optimize storage? 

Current:
```
    mapping(bytes32 => RequestStatus) requests;

    struct DeliveryHistory {
      uint256 blockNumber;
      uint256 lastDeliveredKeyIndex;
    }

    struct RequestStatus {
      uint256 totalItemsCount;
      uint256 deliveredItemsCount;
      uint256 contractVersion;
      DeliveryHistory[] deliverHistory;
    } 
```

Suggested:
```
    mapping(bytes32 => RequestStatusPacked) requestsBasic; 
    mapping(bytes32 => DeliveryHistoryPacked[]) deliveryHistories;

    struct RequestStatusPacked {
        uint16 contractVersion;
        uint16 totalItemsCount;
        uint16 deliveredItemsCount;

        uint16 deliverHistoryLength;
        
        uint32 latestDeliveredBlockNumber;
        uint32 latestDeliveredTimestamp;
    }

    struct DeliveryHistoryPacked {
      uint16 lastDeliveredKeyIndex;
      uint32 blockNumber;
      uint32 blockTimestamp;
    }
```

I suspect that our reports will typically be processed in a single pass, so storing an array of `DeliveryHistory` with one entry per report in storage seems excessive. Instead, we store `deliverHistoryLength` and information about the last report processing (which is usually the only one) in `RequestStatusPacked`. This way, in 99% of cases, we don't need to keep an additional history array in storage.

mkurayan:
I found it a bit inconvenient to jump between *ValidatorsExitBus.sol* and *ValidatorsExitBusOracle.sol*. For example, the description of the report data format (`DATA_FORMAT_LIST`) is outlined in *ValidatorsExitBusOracle.sol*, but when data in `DATA_FORMAT_LIST` is processed in *ValidatorsExitBus.sol*, you have to switch between files to understand the context.

IMHO, the logic in these files seems tightly coupled, and separating it into two files does not simplify things.

I suggest combining *ValidatorsExitBus.sol* and *ValidatorsExitBusOracle.sol* into a single file.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        IWithdrawalVault(withdrawalVaultAddr).addFullWithdrawalRequests{value: requestsFee}(pubkeys);
+
+        if (refund > 0) {
+          (bool success, ) = msg.sender.call{value: refund}("");
+          require(success, "Refund failed");
+          revert TriggerableWithdrawalRefundFailed();
+        }
+
```
mkurayan, line 136: Small suggestion.

        1. Save the initial balance before processing withdrawal requests.
        uint256 prevBalance = address(this).balance - msg.value;

        2. After finishing processing and making a refund, check invariant
        assert(address(this).balance == prevBalance);

contracts/0.8.9/oracle/ValidatorsExitBusOracle.sol
```diff
@@ -109,6 +111,10 @@ contract ValidatorsExitBusOracle is BaseOracle, PausableUntil {
         _initialize(consensusContract, consensusVersion, lastProcessingRefSlot);
     }
 
+    function finalizeUpgrade_v2() external {
```
F4ever, line 117: This method should be part of VEB

contracts/0.8.9/oracle/ValidatorsExitBusOracle.sol
```diff
@@ -216,10 +219,13 @@ contract ValidatorsExitBusOracle is BaseOracle, PausableUntil {
     {
         _checkMsgSenderIsAllowedToSubmitData();
         _checkContractVersion(contractVersion);
+        bytes32 dataHash = keccak256(abi.encode(data.data));
```
F4ever: We should get `keccak256` from data + metadata (in our case it should be some type of data version or data fomat)

ToDo: @F4ever describe new data formats in docs.

F4ever: I think we could simplify here
```suggestion
        bytes32 dataHash = keccak256(data.data);
```

contracts/0.8.9/oracle/ValidatorsExitBusOracle.sol
```diff
@@ -439,6 +446,20 @@ contract ValidatorsExitBusOracle is BaseOracle, PausableUntil {
...
     }
 
+    function _storeOracleExitRequestHash(bytes32 exitRequestHash, ReportData calldata report, uint256 contractVersion) internal {
+        if (report.requestsCount == 0) {
+            return;
+        }
+
+        mapping(bytes32 => RequestStatus) storage hashes = _storageExitRequestsHashes();
+
+        RequestStatus storage request = hashes[exitRequestHash];
```
F4ever: Add event log that such record was created

contracts/0.8.9/oracle/ValidatorsExitBus.sol
F4ever: Contract version should be uint256. It's defined in contract `Versioned`.
Also I'd prefer do such optimization in last iteration. I don't have strong expiriens in solidity, but seems it would make logic more complex

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    function _setLocatorAddress(address addr) internal {
+        if (addr == address(0)) revert ZeroAddress();
+
+        LOCATOR_CONTRACT_POSITION.setStorageAddress(addr);
+    }
+
+    function triggerExitHashVerify(bytes calldata data, uint256[] calldata keyIndexes) external payable {
+        bytes32 dataHash = keccak256(abi.encode(data));
```
F4ever:
```suggestion
        bytes32 dataHash = keccak256(data);
```

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -0,0 +1,118 @@
...
+        RequestStatus storage requestStatus = _storageExitRequestsHashes()[dataHash];
```
F4ever: We should check that `dataHash` persists in mapping and revert with custom exception if no such record

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -0,0 +1,118 @@
...
+
+        address locatorAddr = LOCATOR_CONTRACT_POSITION.getStorageAddress();
+        address withdrawalVaultAddr = ILidoLocator(locatorAddr).withdrawalVault();
+        uint256 minFee = IWithdrawalVault(withdrawalVaultAddr).getWithdrawalRequestFee();
```
F4ever: Not minFee, but withdrawalFee

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -0,0 +1,118 @@
...
+        uint256 requestsFee = keyIndexes.length * minFee;
+
+        if (msg.value < requestsFee) {
+           revert FeeNotEnough(minFee, keyIndexes.length, msg.value);
```
F4ever: Maybe better:
`InsufficientPayment`, `InsufficientETHSent`, `InsufficientValueSent`

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -0,0 +1,118 @@
...
+
+        uint256 refund = msg.value - requestsFee;
+
+        uint256 lastDeliveredKeyIndex = requestStatus.deliveredItemsCount - 1;
+
+        bytes memory pubkeys = new bytes(keyIndexes.length * PUBLIC_KEY_LENGTH);
+
+        for (uint256 i = 0; i < keyIndexes.length; i++) {
+            if (keyIndexes[i] > lastDeliveredKeyIndex) {
+                revert KeyWasNotUnpacked(keyIndexes[i], lastDeliveredKeyIndex);
```
F4ever: We use `delivered` working above in `lastDeliveredKeyIndex`, so maybe `KeyWasNotDelivered`

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -0,0 +1,118 @@
...
+           revert FeeNotEnough(minFee, keyIndexes.length, msg.value);
+        }
+
+        uint256 refund = msg.value - requestsFee;
+
+        uint256 lastDeliveredKeyIndex = requestStatus.deliveredItemsCount - 1;
+
+        bytes memory pubkeys = new bytes(keyIndexes.length * PUBLIC_KEY_LENGTH);
+
+        for (uint256 i = 0; i < keyIndexes.length; i++) {
```
F4ever, line 97: I propose to add here check that 
`require(keyIndexes[i] < requestStatus.totalItemsCount)`

contracts/0.8.9/oracle/ValidatorsExitBus.sol
F4ever: Maybe better this way
https://github.com/lidofinance/core/pull/927/files#r1966406996

F4ever: Let's define what is 16.
```
MSB <---------------------------------------------------- LSB
| 24 bits: moduleId | 40 bits: nodeOpId | 64 bits: valIndex |
```
Or even define and replace it with constants.
Like `keyIndexes[i] * PACKED_REQUEST_LENGTH + PACKED_MODULE_ID_LENGHT + PACKED_NODE_OP_ID_LENGTH + PACKED_VALIDATOR_INDEX_SIZE`

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -0,0 +1,118 @@
...
+
+    /// @dev Errors
+    error KeyWasNotUnpacked(uint256 keyIndex, uint256 lastUnpackedKeyIndex);
+    error ZeroAddress();
+    error FeeNotEnough(uint256 minFeePerRequest, uint256 requestCount, uint256 msgValue);
+    error TriggerableWithdrawalRefundFailed();
+
+    // TODO: make type optimization
+    struct DeliveryHistory {
+      uint256 blockNumber;
```
F4ever: We should use timestamp here. It's crucial for Curated Staking modules to get exact timestamp to track exit process.
Perhaps we should store blockNumber also, but I don't see any usage for it yet.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
mkurayan: IMHO it will not overcomplicate implementation but save a lot of gas.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
Amuhar: changed

contracts/0.8.9/oracle/ValidatorsExitBus.sol
Amuhar: done, thanks

contracts/0.8.9/oracle/ValidatorsExitBus.sol
Amuhar: fixed

contracts/0.8.9/oracle/ValidatorsExitBus.sol
Amuhar: fixed

contracts/0.8.9/oracle/ValidatorsExitBus.sol
Amuhar: fixed

contracts/0.8.9/oracle/ValidatorsExitBus.sol
Amuhar: fixed

contracts/0.8.9/oracle/ValidatorsExitBusOracle.sol
Amuhar: added

contracts/0.8.9/oracle/ValidatorsExitBus.sol
Amuhar: thanks!

contracts/0.8.9/oracle/ValidatorsExitBus.sol
Amuhar: done

PR #929, opened 2025-01-23
[VAULTS]: Granular permissions for operations in Dashboard/Delegation

failingtwice:
This PR introduces granular permissions for each operation in Dashboard/Delegation to allow flexible role setups.


contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -206,33 +215,99 @@ contract Dashboard is AccessControlEnumerable {
...
+        }
+    }
+
     /**
      * @notice Transfers ownership of the staking vault to a new owner.
      * @param _newOwner Address of the new owner.
      */
-    function transferStVaultOwnership(address _newOwner) external virtual onlyRole(DEFAULT_ADMIN_ROLE) {
-        _transferStVaultOwnership(_newOwner);
+    function transferOwnership(address _newOwner) external virtual {
```
folkyatina: StVault was there for a reason. Without it, you may think that Dashboard has some ownership, which can be transferred.

TheDZhon: +1 to @folkyatina

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -206,33 +215,99 @@ contract Dashboard is AccessControlEnumerable {
...
+        _authTransferOwnership();
```
folkyatina: Why not a modifier from the start?

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -206,33 +215,99 @@ contract Dashboard is AccessControlEnumerable {
...
+
+        OwnableUpgradeable(address(stakingVault())).transferOwnership(_newOwner);
     }
 
     /**
      * @notice Disconnects the staking vault from the vault hub.
      */
-    function voluntaryDisconnect() external payable virtual onlyRole(DEFAULT_ADMIN_ROLE) fundAndProceed {
-        _voluntaryDisconnect();
+    function voluntaryDisconnect() external payable virtual onlyRole(VOLUNTARY_DISCONNECT_ROLE) fundAndProceed {
```
folkyatina: Do we need a virtual here and generally everywhere?

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -290,7 +365,7 @@ contract Dashboard is AccessControlEnumerable {
     function mintWstETH(
         address _recipient,
         uint256 _tokens
-    ) external payable virtual onlyRole(DEFAULT_ADMIN_ROLE) fundAndProceed {
+    ) external payable virtual onlyRole(MINT_ROLE) fundAndProceed {
```
folkyatina: Here we break permission model. Mint role can fund, because of `fundAndProceed`

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -400,7 +465,7 @@ contract Dashboard is AccessControlEnumerable {
      * @notice Rebalances the vault by transferring ether
      * @param _ether Amount of ether to rebalance
      */
-    function rebalanceVault(uint256 _ether) external payable virtual onlyRole(DEFAULT_ADMIN_ROLE) fundAndProceed {
+    function rebalanceVault(uint256 _ether) external payable virtual onlyRole(REBALANCE_ROLE) fundAndProceed {
```
folkyatina: rebalance role can fund

folkyatina (changes requested):
The rough idea is right, but I think we can make better. 

WDYT about having granular permissions on a separate layer in separate internal function. 

Like, 

mintWstETH -> _mintWithAuth -> _mint -> stakingVault.mint`

And with minimal method overriding

contracts/0.8.25/vaults/Dashboard.sol
failingtwice: I thought we rejected the "StVault" term

contracts/0.8.25/vaults/Dashboard.sol
failingtwice: Because then, we'll have to reimplement the function with a different modifier. In this current impl, we only need to override the auth function.

contracts/0.8.25/vaults/Dashboard.sol
failingtwice: Nope, will remove it.

contracts/0.8.25/vaults/Dashboard.sol
failingtwice: Let's get rid of `fundAndProceed` maybe?

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -38,24 +38,24 @@ interface IWstETH is IERC20, IERC20Permit {
...
-contract Dashboard is AccessControlEnumerable {
+contract Dashboard is Permissions {
     /// @notice Address of the implementation contract
     /// @dev Used to prevent initialization in the implementation
     address private immutable _SELF;
     /// @notice Total basis points for fee calculations; equals to 100%.
     uint256 internal constant TOTAL_BASIS_POINTS = 10000;
 
+    /// @notice Indicates whether the contract has been initialized
+    bool public initialized;
```
folkyatina: Should we group this member with the other storage one? Seems strange to have it between constants and immutables.

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -38,24 +38,24 @@ interface IWstETH is IERC20, IERC20Permit {
...
+
     /// @notice The stETH token contract
-    IStETH public immutable STETH;
```
folkyatina, line 49: Why do we change public to explicit methods? Public members has better indication of immutability by the upper case names.

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -70,51 +70,72 @@ contract Dashboard is AccessControlEnumerable {
...
     }
 
     // ==================== View Functions ====================
 
+    /// @notice The underlying `StakingVault` contract
+    function stakingVault() external view returns (address) {
+        return address(_stakingVault());
+    }
+
+    function stETH() external view returns (address) {
```
folkyatina: Seems better as a `public immutable`.

contracts/0.8.25/vaults/Dashboard.sol
loga4: let's move `initialized` closer to `vaultHub` var (right now it's between const and immutable)

contracts/0.8.25/utils/MassAccessControl.sol
```diff
@@ -0,0 +1,47 @@
...
+
+import {IZeroArgument} from "../interfaces/IZeroArgument.sol";
+
+/**
+ * @title MassAccessControl
+ * @author Lido
+ * @notice Mass-grants and revokes roles.
+ */
+abstract contract MassAccessControl is AccessControlEnumerable, IZeroArgument {
+    struct Ticket {
```
folkyatina: Assignment, maybe?

contracts/0.8.25/vaults/Permissions.sol
```diff
@@ -0,0 +1,100 @@
...
+    function _rebalanceVault(uint256 _ether) internal onlyRole(REBALANCE_ROLE) {
+        _stakingVault().rebalance(_ether);
+    }
+
+    function _requestValidatorExit(bytes calldata _pubkey) internal onlyRole(REQUEST_VALIDATOR_EXIT_ROLE) {
+        _stakingVault().requestValidatorExit(_pubkey);
+    }
+
+    function _voluntaryDisconnect() internal onlyRole(VOLUNTARY_DISCONNECT_ROLE) {
+        uint256 shares = _vaultHub().vaultSocket(address(_stakingVault())).sharesMinted;
```
folkyatina: Can we move all the combination logic upper to the Dashboard and leave here only calls to the vault and auth?

contracts/0.8.25/vaults/Permissions.sol
```diff
@@ -0,0 +1,100 @@
...
+     * @notice Permission for requesting validator exit from the StakingVault.
+     */
+    bytes32 public constant REQUEST_VALIDATOR_EXIT_ROLE = keccak256("StakingVault.Permissions.RequestValidatorExit");
+
+    /**
+     * @notice Permission for voluntary disconnecting the StakingVault.
+     */
+    bytes32 public constant VOLUNTARY_DISCONNECT_ROLE = keccak256("StakingVault.Permissions.VoluntaryDisconnect");
+
+    function _stakingVault() internal view virtual returns (IStakingVault);
```
folkyatina: Can we store the address in this contracts without this virtual methods overriding?

contracts/0.8.25/vaults/Permissions.sol
```diff
@@ -0,0 +1,100 @@
...
+
+    function _vaultHub() internal view virtual returns (VaultHub);
+
+    function _stETH() internal view virtual returns (IStETH);
+
+    function _votingCommittee() internal pure virtual returns (bytes32[] memory);
```
folkyatina: This is probably the only one that worth to be virtual.

folkyatina (changes requested):
Seems very close to perfection. 

Few things that I'd consider to improve:
- get rid of method overriding, where possible. It's really hinders the readability and I don't see any meaningful things we really win by using it. 
- also, I'm not sure that `ZeroArgument` and `MassACL` really deserves their own files/contracts. Structure complexity is not a joke, you know.

beage666 (approved):

beage666 (approved):

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -210,44 +209,49 @@ contract Dashboard is AccessControlEnumerable {
...
-        _transferStVaultOwnership(_newOwner);
+    function transferStakingVaultOwnership(address _newOwner) external {
+        _transferStakingVaultOwnership(_newOwner);
     }
 
     /**
      * @notice Disconnects the staking vault from the vault hub.
      */
-    function voluntaryDisconnect() external payable virtual onlyRole(DEFAULT_ADMIN_ROLE) fundAndProceed {
+    function voluntaryDisconnect() external payable fundAndProceed {
```
folkyatina, line 219: Did we want to rename `fundAndProceed` to `fundable`, isn't it?

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -210,44 +209,49 @@ contract Dashboard is AccessControlEnumerable {
...
      * @param _newOwner Address of the new owner.
      */
-    function transferStVaultOwnership(address _newOwner) external virtual onlyRole(DEFAULT_ADMIN_ROLE) {
-        _transferStVaultOwnership(_newOwner);
+    function transferStakingVaultOwnership(address _newOwner) external {
+        _transferStakingVaultOwnership(_newOwner);
     }
 
     /**
      * @notice Disconnects the staking vault from the vault hub.
```
folkyatina, line 217: Need to explain multi-permission nature of this method. Mean, you may need three different roles to use it.

contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -38,28 +38,38 @@ interface IWstETH is IERC20, IERC20Permit {
...
-contract Dashboard is AccessControlEnumerable {
-    /// @notice Address of the implementation contract
-    /// @dev Used to prevent initialization in the implementation
-    address private immutable _SELF;
-    /// @notice Total basis points for fee calculations; equals to 100%.
+contract Dashboard is Permissions {
+    /**
+     * @notice Struct containing an account and a role for granting/revoking roles.
+     */
+    struct RoleAssignment {
```
folkyatina, line 45: This mechanics fits nicely in Permissions.sol, I believe.

contracts/0.8.25/vaults/Delegation.sol
```diff
@@ -454,20 +249,40 @@ contract Delegation is Dashboard {
      * @dev Claims the curator/node operator fee amount.
      * @param _recipient The address to which the fee will be sent.
      * @param _fee The accrued fee amount.
+     * @dev Use `Permissions._unsafeWithdraw()` to avoid the `WITHDRAW_ROLE` check.
```
folkyatina, line 252: Why not just a `stackingVault().withdraw()` ?

folkyatina (approved):
Nothing that really blocks it from merging.

PR #930, opened 2025-01-27
feat: verify deposit guardian upon deposit

failingtwice:
This PR adds a guardian check upon deposit to beacon chain. The guardian is an account that signs on the validity deposit data (e.g. no frontruns)

### DIFF
--- a/contracts/0.8.25/vaults/StakingVault.sol
+++ b/contracts/0.8.25/vaults/StakingVault.sol
@@ -5,8 +5,11 @@
 pragma solidity 0.8.25;
 
 import {OwnableUpgradeable} from "contracts/openzeppelin/5.2/upgradeable/access/OwnableUpgradeable.sol";
+import {ERC165} from "@openzeppelin/contracts-v5.2/utils/introspection/ERC165.sol";
+import {IERC1271} from "@openzeppelin/contracts-v5.2/interfaces/IERC1271.sol";
 
 import {VaultHub} from "./VaultHub.sol";
+import {SignatureUtils} from "contracts/common/lib/SignatureUtils.sol";
 
 import {IDepositContract} from "../interfaces/IDepositContract.sol";
 import {IStakingVault} from "./interfaces/IStakingVault.sol";
@@ -67,6 +70,7 @@ contract StakingVault is IStakingVault, OwnableUpgradeable {
         uint128 locked;
         int128 inOutDelta;
         address nodeOperator;
+        address depositGuardian;
         bool beaconChainDepositsPaused;
     }
 
@@ -75,6 +79,8 @@ contract StakingVault is IStakingVault, OwnableUpgradeable {
      *         The implementation is petrified to this version
      */
     uint64 private constant _VERSION = 1;
+
+    bytes32 public constant DEPOSIT_GUARDIAN_MESSAGE_PREFIX = keccak256("StakingVault.DepositGuardianMessagePrefix");
 
     /**
      * @notice Address of `VaultHub`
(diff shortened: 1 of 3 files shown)


contracts/0.8.25/vaults/interfaces/IStakingVault.sol
```diff
@@ -27,27 +27,59 @@ interface IStakingVault {
         bytes32 depositDataRoot;
     }
 
+    struct GuardianSignature {
```
folkyatina: We need unstructured signature here, because PDG will consume something other than ECDSA as a signature, e.g. pubkey. So we need `bytes signature` here.

contracts/0.8.25/vaults/StakingVault.sol
```diff
@@ -313,16 +320,26 @@ contract StakingVault is IStakingVault, OwnableUpgradeable {
      * @param _deposits Array of deposit structs
      * @dev Includes a check to ensure StakingVault is balanced before making deposits
      */
-    function depositToBeaconChain(Deposit[] calldata _deposits) external {
+    function depositToBeaconChain(
+        Deposit[] calldata _deposits,
+        bytes32 _expectedGlobalDepositRoot,
+        GuardianSignature calldata _signature
```
folkyatina: We need just `bytes signature` here and looks like SignatureUtils is not fit for our needs

contracts/0.8.25/vaults/StakingVault.sol
```diff
@@ -331,9 +348,27 @@ contract StakingVault is IStakingVault, OwnableUpgradeable {
                 deposit.signature,
                 deposit.depositDataRoot
             );
+
             totalAmount += deposit.amount;
+            concatenatedDepositDataRoots = abi.encodePacked(concatenatedDepositDataRoots, deposit.depositDataRoot);
         }
 
+        if (
+            !SignatureUtils.isValidSignature(
```
DiRaiks: Why does signature verification occur after sending a deposit to the beacon chain? ```BEACON_CHAIN_DEPOSIT_CONTRACT.deposit{value: deposit.amount}(...args)```
Shouldn't it be the other way around?

contracts/0.8.25/vaults/StakingVault.sol
```diff
@@ -313,27 +318,55 @@ contract StakingVault is IStakingVault, OwnableUpgradeable {
...
         if (msg.sender != $.nodeOperator) revert NotAuthorized("depositToBeaconChain", msg.sender);
         if ($.beaconChainDepositsPaused) revert BeaconChainDepositsArePaused();
         if (!isBalanced()) revert Unbalanced();
 
         uint256 totalAmount = 0;
         uint256 numberOfDeposits = _deposits.length;
+        // XOR is a commutative operation, so the aggregate root will be the same regardless of the order of deposits
+        bytes32 depositDataBatchXorRoot;
+
         for (uint256 i = 0; i < numberOfDeposits; i++) {
```
folkyatina, line 377: I'd rather see here two loops and the signature validation before the deposit if we don't save the huge amount of gas here.

contracts/0.8.25/vaults/StakingVault.sol
```diff
@@ -617,6 +677,21 @@ contract StakingVault is IStakingVault, OwnableUpgradeable {
...
+
+    /**
+     * @notice Thrown when the guardian signature is invalid
+     */
+    error DepositGuardianSignatureInvalid();
+
+    /**
+     * @notice Thrown when the deposit guardian is not an EIP-1271 contract
+     */
+    error DepositGuardianContractDoesNotSupportEIP1271();
```
folkyatina, line 713: Not used?

PR #949, opened 2025-02-17
[VAULTS] Sideloading

failingtwice:
Sideloading is a mechanism that allows vaults to temporarily mint unbacked stETH shares for immediate use, with the requirement that sufficient backing must be added to the vault during the same transaction. This creates a flash-loan-like mechanism specific to stVaults.


TheDZhon:
image

contracts/0.8.25/vaults/FlashMinter.sol
```diff
@@ -0,0 +1,101 @@
...
+        address _vault,
+        address _recipient,
+        uint256 _amountOfShares,
+        bytes calldata _data
+    ) external whenResumed {
+        // basic checks & auth
+        if (_vault == address(0)) revert ZeroArgument("_vault");
+        if (_amountOfShares == 0) revert ZeroArgument("_amountOfShares");
+        // allow only registered recipients or the vault owner itself for flash minting
+        if (!isRegisteredFlashMintRecipient[_recipient] && _recipient != OwnableUpgradeable(_vault).owner())
```
folkyatina: The second part of the condition does not look right. It's basically enables any contract to flashmint.

folkyatina: Do we need a way to disable a whitelist, wdyt?

contracts/0.8.25/vaults/FlashMinter.sol
```diff
@@ -0,0 +1,101 @@
...
+
+        // call the recipient with the provided data
+        (bool success, ) = _recipient.call(_data);
+        if (!success) revert CallbackFailed(_recipient, _data);
+
+        // shares cannot exceed the max mintable amount
+        uint256 maxMintableShares = _maxMintableShares(_vault, reserveRatioBP, shareLimit);
+
+        if (vaultSharesAfterMint > maxMintableShares) {
+            revert InsufficientValuationToMint(_vault, IStakingVault(_vault).valuation());
```
folkyatina: Lets do a separate error for this

contracts/0.8.25/vaults/FlashMinter.sol
```diff
@@ -0,0 +1,101 @@
...
+
+        if (totalEtherLocked > IStakingVault(_vault).locked()) {
+            IStakingVault(_vault).lock(totalEtherLocked);
+        }
+
+        // mint the shares to the recipient
+        STETH.mintExternalShares(_recipient, _amountOfShares);
+
+        // call the recipient with the provided data
+        (bool success, ) = _recipient.call(_data);
```
folkyatina: Passing calldata as an arg feels like a critical, tbh. 

I think that's better to stick to some interface to limit potential impact surface. E.g. ERC-3156 have pretty decent interfaces we can partly reuse.

folkyatina (commented):
Think we need it to be much stricter: 
- no arbitrary calldata as an argument
- interface closer to flashloan for the recipient (without unnecessary bits)
- no callback to owners, only to registered "providers"

And maybe call it "leverage" and "leverageProvider" registry.

contracts/0.8.25/vaults/Sideloading.sol
```diff
@@ -0,0 +1,324 @@
...
+pragma solidity 0.8.25;
+
+import {VaultHub} from "./VaultHub.sol";
+
+import {ILido as IStETH} from "../interfaces/ILido.sol";
+import {IStakingVault} from "./interfaces/IStakingVault.sol";
+import {ISideloader} from "../interfaces/ISideloader.sol";
+
+/**
+ * @title Sideloading
```
TheDZhon, line 15: maybe add an inspiration note on flash loan

contracts/0.8.25/vaults/Sideloading.sol
```diff
@@ -0,0 +1,324 @@
...
+     *                                   |     & callback           |
+     *                                   |                          |
+     *                                   v                          v
+     *   +---------------+              +------------+            +---------+
+     *   |               |              |            |            |         |
+     *   |   e.g. DEX    |<------------ | Sideloader |----------->| stVault |
+     *   |               |   4. swap    |            |  5. fund   |         |
+     *   +---------------+      to ETH  +------------+            +---------+
+     *
+     * NB: Sideloading conservatively assumes that the sideloader will not be able to
```
TheDZhon: maybe better provide rationale upfront

contracts/0.8.25/vaults/Sideloading.sol
```diff
@@ -0,0 +1,324 @@
...
+        // update the shares minted BEFORE sideloading
+        socket.sharesMinted = uint96(totalSharesAfterSideload);
+
+        // update the locked amount BEFORE sideloading
+        if (totalEtherLocked > IStakingVault(_vault).locked()) {
+            IStakingVault(_vault).lock(totalEtherLocked);
+        }
+
+        // mint the shares to the sideloader
+        STETH.mintExternalShares(_sideloader, _amountOfShares);
```
TheDZhon: maybe extract (not critical)

TheDZhon: consider future operator grid and migration

TheDZhon (commented):
:eyes:

contracts/0.8.25/vaults/Sideloading.sol
failingtwice: will be done separately, once grid and migration is implemented

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -496,14 +493,12 @@ contract VaultHub is PausableUntilWithRoles {
...
 
-            if (socket.pendingDisconnect) continue; // we skip disconnected vaults
-
             uint256 treasuryFeeShares = _treasureFeeShares[i];
             if (treasuryFeeShares > 0) {
                 socket.sharesMinted += uint96(treasuryFeeShares);
             }
 
-            IStakingVault(socket.vault).report(_valuations[i], _inOutDeltas[i], _locked[i]);
+            IStakingVault(socket.vault).report(_valuations[i], _inOutDeltas[i], socket.pendingDisconnect ? 0 : _locked[i]);
```
folkyatina: Better to move the condition to `calculate` method.

TheDZhon:
Closing due to additional complexity atop of the flashloan approach

PR #957, opened 2025-02-20
Detach Accounting from VaultHub

vp4242:
Detach Accounting from VaultHub

### DIFF
--- a/contracts/0.4.24/Lido.sol
+++ b/contracts/0.4.24/Lido.sol
@@ -586,7 +586,7 @@ contract Lido is Versioned, StETHPermit, AragonApp {
      * @dev can be called only by accounting
      */
     function mintShares(address _recipient, uint256 _amountOfShares) public {
-        _auth(getLidoLocator().accounting());
+        _authBoth(getLidoLocator().accounting(), getLidoLocator().vaultHub());
         _whenNotStopped();
 
         _mintShares(_recipient, _amountOfShares);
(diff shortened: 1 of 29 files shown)


contracts/0.4.24/Lido.sol
```diff
@@ -586,7 +586,7 @@ contract Lido is Versioned, StETHPermit, AragonApp {
      * @dev can be called only by accounting
      */
     function mintShares(address _recipient, uint256 _amountOfShares) public {
-        _auth(getLidoLocator().accounting());
+        _authBoth(getLidoLocator().accounting(), getLidoLocator().vaultHub());
```
failingtwice: This function is only used once. Perhaps, it's better to do the check inline:
- save up on bytecode size
- save up a little on gas, probably (no extra `JUMPI`).

failingtwice (approved):
lgtm

tamtamchik (approved):

contracts/0.4.24/Lido.sol
```diff
@@ -586,7 +586,7 @@ contract Lido is Versioned, StETHPermit, AragonApp {
      * @dev can be called only by accounting
      */
     function mintShares(address _recipient, uint256 _amountOfShares) public {
-        _auth(getLidoLocator().accounting());
+        require(msg.sender == getLidoLocator().accounting() || msg.sender == getLidoLocator().vaultHub(), "APP_AUTH_FAILED");
```
folkyatina, line 589: Check, please, if it's possible to split this auth in two like in burnShares.

PR #967, opened 2025-03-01
Fix parameter naming inconsistency in OracleDaemonConfig error

crStiv:
Standardizes parameter naming in error declarations by adding underscore prefix to parameter names in ValueExists, EmptyValue, and ValueDoesntExist errors to match the style used in ValueIsSame error.


tamtamchik:
Sorry, but the change is irrelevant.

PR #970, opened 2025-03-05
test: vaults initial permissions test

cakeinpanic:
Added integraiton tests for roles in Delegation


test/integration/utils/test-utils.ts
```diff
@@ -0,0 +1,11 @@
+import { HardhatEthersSigner } from "@nomicfoundation/hardhat-ethers/signers";
+import { impersonate, randomAddress } from "lib";
+import { ether } from "../../../lib/units";
+
+export async function getRandomSigners(amount: number): Promise<HardhatEthersSigner[]> {
```
tamtamchik: Maybe move to `lib`? it is kinda utils folder

test/integration/utils/test-utils.ts
```diff
@@ -0,0 +1,11 @@
+import { HardhatEthersSigner } from "@nomicfoundation/hardhat-ethers/signers";
+import { impersonate, randomAddress } from "lib";
+import { ether } from "../../../lib/units";
```
tamtamchik: Please try to use bare imports instead of relative, to make it easier to refactor the code

test/integration/utils/test-utils.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export async function getRandomSigners(amount: number): Promise<HardhatEthersSigner[]> {
+  const signers = [];
+  for (let i = 0; i < amount; i++) {
+    signers.push(await impersonate(randomAddress(), ether("1")));
```
tamtamchik:
```suggestion
    signers.push(await impersonate(randomAddress(), ether("10000")));
```
It's ok to have some big amount, may be used for deposits and some funding

test/integration/vaults/roles.integration.ts
```diff
@@ -0,0 +1,370 @@
...
+        });
+
+        it("claimCuratorFee", async () => {
+          await testGrantingRole(testDelegation, "claimCuratorFee", await testDelegation.CURATOR_FEE_CLAIM_ROLE(), [
+            stranger,
+          ]);
+        });
+
+        // todo: find out why owner can't grant this role
+        it.skip("claimNodeOperatorFee", async () => {
```
tamtamchik: This is a specific role under the control of `nodeOperatorManager` (initially set in the `initialize` method of the Delegation contract, so the owner can't change it). Pass the actor to the testGrantingRole?

tamtamchik (changes requested):
Thank you! 
It looks good, although some linting love (see CI / Linting) and enabling skipped tests (need test code update) are required.

ps. strange that husky didn't lint staged automatically

test/integration/utils/test-utils.ts
cakeinpanic: totally agree

test/integration/vaults/roles.integration.ts
cakeinpanic: fixed

test/integration/utils/test-utils.ts
cakeinpanic: fixed

test/integration/utils/test-utils.ts
cakeinpanic: fixed!

package.json
tamtamchik, line 1: May you please revert those changes? Looks like they are temporary

test/integration/vaults/roles.integration.ts
```diff
@@ -0,0 +1,406 @@
...
+const VAULT_NODE_OPERATOR_FEE = 1_00n; // 3% node operator fee
+
+type Methods<T> = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [K in keyof T]: T[K] extends (...args: any) => any ? K : never; // gdfg
+}[keyof T];
+
+type DelegationMethods = Methods<Delegation>; // "foo" | "bar"
+
+describe("Scenario: Staking Vaults Delegation Roles", () => {
```
tamtamchik:
```suggestion
describe("Integration: Staking Vaults Delegation Roles Initial Setup", () => {
```
It's not like a scenario test, but more regular integration + updated title a bit

test/integration/vaults/roles.integration.ts
```diff
@@ -0,0 +1,406 @@
...
+  });
+
+  beforeEach(async () => {
+    snapshot = await Snapshot.take();
+  });
+
+  afterEach(async () => await Snapshot.restore(snapshot));
+
+  // initializing contracts with signers
+  describe("Full contract initialization", () => {
```
tamtamchik:
```suggestion
  describe("Vault created with all the roles", () => {
```

test/integration/vaults/roles.integration.ts
```diff
@@ -0,0 +1,406 @@
...
+          nodeOperatorFeeClaimers: [nodeOperatorFeeClaimers],
+        },
+        "0x",
+      );
+
+      const createVaultTxReceipt = (await deployTx.wait()) as ContractTransactionReceipt;
+      const createVaultEvents = ctx.getEvents(createVaultTxReceipt, "VaultCreated");
+
+      testDelegation = await ethers.getContractAt("Delegation", createVaultEvents[0].args?.owner);
+          });
```
tamtamchik:
```suggestion
    });
```

tamtamchik: Should be fixed automatically by husky... 🤔

test/integration/vaults/roles.integration.ts
```diff
@@ -0,0 +1,406 @@
...
+
+    describe("Only roles", () => {
```
tamtamchik:
```suggestion
    describe("Verify ACL for methods that require only role", () => {
```

test/integration/vaults/roles.integration.ts
```diff
@@ -0,0 +1,406 @@
...
+              failingUsers: allRoles.filter((r) => r !== assetRecoverer),
+            },
+            [ZeroAddress, 0, stranger],
+            await testDelegation.ASSET_RECOVERY_ROLE(),
+          );
+        });
+      });
+    });
+
+    describe("Only confirmed roles", () => {
```
tamtamchik:
```suggestion
    describe("Verify ACL for methods that require confirmations", () => {
```

test/integration/vaults/roles.integration.ts
```diff
@@ -0,0 +1,406 @@
...
+      expect(await testDelegation.connect(funder).withdrawableEther()).to.equal(0);
+    });
+
+    it("Allows to retrieve roles addresses", async () => {
+      expect(await testDelegation.getRoleMembers(await testDelegation.MINT_ROLE())).to.deep.equal([minter.address]);
+    });
+  });
+
+  // initializing contracts without signers
+  describe("Empty contract initialization", () => {
```
tamtamchik:
```suggestion
  describe("Vault created with no roles", () => {
```

test/integration/vaults/roles.integration.ts
```diff
@@ -0,0 +1,406 @@
...
+        "0x",
+      );
+
+      const createVaultTxReceipt = (await deployTx.wait()) as ContractTransactionReceipt;
+      const createVaultEvents = ctx.getEvents(createVaultTxReceipt, "VaultCreated");
+
+      testDelegation = await ethers.getContractAt("Delegation", createVaultEvents[0].args?.owner);
+    });
+
+    describe("Only roles", () => {
```
tamtamchik:
```suggestion
    describe("Verify ACL for methods that require only role", () => {
```

test/integration/vaults/roles.integration.ts
```diff
@@ -0,0 +1,406 @@
...
+          await testGrantingRole(
+            testDelegation,
+            "claimNodeOperatorFee",
+            await testDelegation.NODE_OPERATOR_FEE_CLAIM_ROLE(),
+            [stranger],
+            nodeOperatorManager
+          );
+        });
+      });
+    });
```
tamtamchik, line 578: ⚠️ I believe we are missing the Dashboard methods, and those need to be confirmed
I know that all they do is reject, but it is still necessary to test that rejection, wdyt?

tamtamchik: ⚠️  Also, I suggest testing that, in fact, all the roles that should be [] are [] using the getRoleMembers function. This is important to verify because we deploy contracts indirectly via the factory and it may do something with roles, so we want to verify that we get all the roles as proposed, and they are not being manipulated during the deploy

tamtamchik (changes requested):
Sorry for me being slowpoke and not reporting it before, but it looks like we need to add more tests to cover more cases. Please see the comments.

test/integration/vaults/roles.integration.ts
cakeinpanic: test for getRoleMembers - done - agree, it's important to test

cakeinpanic: regarding other dashboard tests – I'm adding them now, but some methods fail with another errors _before_ they check roles, and I'm not sure if it worth creating prerequisites in this testset

for example
```
  10) Integration: Staking Vaults Delegation Roles Initial Setup
       Vault created with all the roles
         Verify ACL for methods that require only role
           Dashboard methods
             withdraw:
     AssertionError: Expected transaction to be reverted with custom error 'AccessControlUnauthorizedAccount', but it reverted with custom error 'RequestedAmountExceedsUnreserved'

```


other methods
burnWstETHWithPermit, burnStETHWithPermit, burnSharesWithPermit, withdrawWETH, withdraw, fundWeth

tamtamchik (approved):
💚

PR #975, opened 2025-03-07
feat: preparations for the upgrade code

arwer13:
NB: the upgrade code isn't part of this PR

1. Don't read addresses from locator in Sanity Checker ctor and Delegation ctor. Otherwise we'd have to upgrade LidoLocator from Aragon voting twice, setting incorrect LidoLocator impl first.
2. Fix 3 tests which weren't designed properly to be run on forking after upgrade
3. Fix upgrade initialization for AccountingOracle and Lido


contracts/0.8.25/vaults/Dashboard.sol
```diff
@@ -75,15 +75,17 @@ contract Dashboard is Permissions {
     /**
      * @notice Constructor sets the stETH, WETH, and WSTETH token addresses.
      * @param _wETH Address of the weth token contract.
+     * @param _wstETH Address of the wsteth token contract.
      * @param _lidoLocator Address of the Lido locator contract.
      */
-    constructor(address _wETH, address _lidoLocator) {
+    constructor(address _wETH, address _wstETH, address _lidoLocator) {
```
tamtamchik, line 81: wstETH we do have in locator now

arwer13: it is in the new locator only which is to be set only upon Aragon voting execution

PR #984, opened 2025-03-14
test: add vault creation tests

cakeinpanic:
Could not debug and find out why some methods throwing specific errors
Market those tests with "skip" and added comments

Testing Vault creation flow https://www.notion.so/System-Scenarios-168bf633d0c98033ad7ed41465c55715?pvs=4#17abf633d0c9804ea74ee1e11545940c


test/integration/vaults/vault-creation.integration.ts
```diff
@@ -0,0 +1,236 @@
...
+    );
+    const createVaultTxReceipt = (await deployTx.wait()) as ContractTransactionReceipt;
+    const createVaultEvents = ctx.getEvents(createVaultTxReceipt, "VaultCreated");
+
+    expect(createVaultEvents.length).to.equal(1n);
+
+    stakingVault = await ethers.getContractAt("StakingVault", createVaultEvents[0].args?.vault);
+    delegation = await ethers.getContractAt("Delegation", createVaultEvents[0].args?.owner);
+
+    snapshot = await Snapshot.take();
```
tamtamchik: Snapshots should be taken before any on-chain actions (e.g., createVaultWithDelegation) to ensure that the blockchain can be restored to its original state without any remnants.

test/integration/vaults/vault-creation.integration.ts
```diff
@@ -0,0 +1,236 @@
...
+  });
+
+  after(async () => await Snapshot.restore(snapshot));
+
+  async function connectToHub() {
```
tamtamchik: How about adding lib/protocol/vaults with common methods used across tests (such as creating, connecting, and disconnecting the vault to protocol and so on)?

test/integration/vaults/vault-creation.integration.ts
```diff
@@ -0,0 +1,236 @@
...
+  it("Allows to pause/resume deposits to validators", async () => {
+    expect(await delegation.connect(depositPausers).pauseBeaconChainDeposits()).to.be.ok;
+    expect(await delegation.connect(depositResumers).resumeBeaconChainDeposits()).to.be.ok;
+  });
+
+  it("Allows to force / Ask Node Operator to withdraw funds from validator(s)", async () => {
+    const vaultOwnerAddress = await stakingVault.owner();
+    const vaultOwner: ContractRunner = await impersonate(vaultOwnerAddress, ether("10000"));
+    expect(await stakingVault.connect(vaultOwner).requestValidatorExit(SAMPLE_PUBKEY)).to.be.ok;
+    Tracing.enable();
```
tamtamchik: Leftover, maybe set up a husky task to block commit with `Tracing.enable()` ... 🤔

test/integration/vaults/vault-creation.integration.ts
```diff
@@ -0,0 +1,250 @@
...
+        vaultHub,
+        "NotConnectedToHub",
+      );
+      await disconnectFromHub();
+    });
+
+    it("to burn stEth", async () => {
+      const { vaultHub } = ctx.contracts;
+
+      // todo: why it fails with ALLOWANCE_EXCEEDED before NotConnectedToHub? How to fix it?
```
cakeinpanic: weird error I did not succeed to debug

test/integration/vaults/vault-creation.integration.ts
```diff
@@ -0,0 +1,250 @@
...
+      before(async () => {
+        await delegation.connect(funder).fund({ value: ether("1") });
+        await generateFeesToClaim();
+      });
+
+      it("to claim Curator's fee", async () => {
+        const { vaultHub } = ctx.contracts;
+        //await connectToHub();
+        // todo: expecting for method to be reverted because we are not connected to the hub, but it is passing fine
+        await expect(delegation.connect(curatorFeeClaimers).claimCuratorFee(stranger)).to.be.revertedWithCustomError(
```
cakeinpanic, line 220: Dima says that `claimCuratorFee` and `claimNodeOperatorFee` should not work until connected to hub, but here they pass fine

And if I manually disconnect from Hub or don't do `await stakingVault.connect(hubSigner).report(rewards, 0n, 0n);`,  it gives ZeroArgument

Does it work as it is supposed to work?

test/integration/vaults/roles.integration.ts
```diff
@@ -125,6 +125,33 @@ describe("Integration: Staking Vaults Delegation Roles Initial Setup", () => {
       testDelegation = await ethers.getContractAt("Delegation", createVaultEvents[0].args?.owner);
     });
 
+    it("Allows anyone to read public metrics of the vault", async () => {
```
X-astro, line 123: this case looks meaningless. 🤔

test/integration/vaults/vault-creation.integration.ts
X-astro:
```
    /**
     * @dev Burns stETH tokens from the sender backed by the vault
     * @param _amountOfStETH Amount of tokens to burn
     */
    function _burnStETH(uint256 _amountOfStETH) internal {
        uint256 _amountOfShares = STETH.getSharesByPooledEth(_amountOfStETH);
        STETH.transferSharesFrom(msg.sender, address(vaultHub), _amountOfShares);
        _burnShares(_amountOfShares);
    }
```
@cakeinpanic because it tries to transfer shares token before updating related shares state over contract.

X-astro (changes requested):

test/integration/vaults/roles.integration.ts
cakeinpanic: it may test that no one would add roles protection on public methods....

you think needs to be removed?

test/integration/vaults/vault-creation.integration.ts
tamtamchik: Yep, some approval should be made for burning, should be visible in happy path test

tamtamchik (commented):
Need to fix working with snapshots, otherwise, it'll break the invariant that tests introduce no different in the state between scenarios (on file level)

test/integration/vaults/vault-creation.integration.ts
cakeinpanic: fixed!

test/integration/vaults/vault-creation.integration.ts
```diff
@@ -0,0 +1,250 @@
...
+          delegation.connect(nodeOperatorFeeClaimers).claimNodeOperatorFee(stranger),
+        ).to.be.revertedWithCustomError(vaultHub, "NotConnectedToHub");
+
+        await disconnectFromHub();
+      });
+    });
+  });
+
+  it("NO Manager can spawn a validator using ETH from the Vault ", () => {
+    // todo: what method to use?
```
cakeinpanic: @tamtamchik please remind which method to use here

tamtamchik: In the happy path, there is a deposit part that can be taken from there. Although it is now huge (using PDG) and probably this should be refactored into helper methods inside the context 🤔

lib/protocol/vaults.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export async function connectToHub(
+  ctx: ProtocolContext,
+  stakingVault: StakingVault,
+  { reserveRatio, rebalanceThreshold }: { reserveRatio: bigint; rebalanceThreshold: bigint },
+) {
+  const { vaultHub } = ctx.contracts;
+  const agentSigner = await ctx.getSigner("agent");
+
+  const treasuryFeeBP = 5_00n; // 5% of the treasury fee
```
tamtamchik: Move to options?

test/integration/vaults/vault-creation.integration.ts
```diff
@@ -0,0 +1,215 @@
...
+  let snapshot: string;
+  let originalSnapshot: string;
+
+  before(async () => {
+    ctx = await getProtocolContext();
+
+    originalSnapshot = await Snapshot.take();
+
+    // ERC7002 pre-deployed contract mock (0x00000961Ef480Eb55e80D19ad83579A64c007002)
+    await deployWithdrawalsPreDeployedMock(1n);
```
tamtamchik: No need anymore, now pre-deployed by default

test/integration/vaults/vault-creation.integration.ts
```diff
@@ -0,0 +1,215 @@
...
+  afterEach(async () => await Snapshot.restore(snapshot));
+
+  after(async () => await Snapshot.restore(originalSnapshot));
+
+  it("Allows to fund an withdraw funds for dedicated roles", async () => {
+    expect(await delegation.connect(funder).fund({ value: 2n })).to.be.ok;
+    expect(await delegation.connect(owner).withdrawableEther()).to.equal(2n);
+    expect(await delegation.connect(withdrawer).withdraw(stranger, 2n)).to.be.ok;
+    expect(await delegation.connect(owner).withdrawableEther()).to.equal(0);
+  });
```
tamtamchik:
```suggestion
  it("Allows to fund an withdraw funds for dedicated roles", async () => {
    await expect(delegation.connect(funder).fund({ value: 2n }))
      .to.emit(stakingVault, "Funded")
      .withArgs(delegation, 2n);

    expect(await delegation.connect(owner).withdrawableEther()).to.equal(2n);
    expect(await delegation.connect(withdrawer).withdraw(stranger, 2n))
      .to.emit(stakingVault, "Withdrawn")
      .withArgs(delegation, 2n);

    expect(await delegation.connect(owner).withdrawableEther()).to.equal(0);
  });
```

tamtamchik: And so on, it is crucial to check that event emitted (and state changed) because this is an expected behaviour from the function call, events are part of the logic

tamtamchik (changes requested):
💭 Please add events checking.

test/integration/vaults/vault-creation.integration.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    await pdg.topUpNodeOperatorBalance(nodeOperatorManager, { value: ether("1") });
+
+    const vaultWC = await stakingVault.withdrawalCredentials();
+    const validator = generateValidator(vaultWC);
+    const predepositData = generatePredeposit(validator);
+
+    await expect(pdg.predeposit(stakingVault, [predepositData]))
+      .to.emit(stakingVault, "DepositedToBeaconChain")
+      .withArgs(ctx.contracts.predepositGuarantee.address, 1, 1000000000000000000n);
```
cakeinpanic, line 259: could not reverse-engineer why it comes to 1000000000000000000n here, but test is green :)
@tamtamchik

tamtamchik: 1n Ether = 1000000000000000000n wei, looks good

test/integration/vaults/vault-creation.integration.ts
```diff
@@ -0,0 +1,246 @@
...
+  after(async () => await Snapshot.restore(originalSnapshot));
+
+  async function generateFeesToClaim() {
+    const { vaultHub } = ctx.contracts;
+    const hubSigner = await impersonate(await vaultHub.getAddress(), ether("100"));
+    const rewards = ether("1");
+    await stakingVault.connect(hubSigner).report(rewards, 0n, 0n);
+  }
+
+  it("Allows to fund an withdraw funds for dedicated roles", async () => {
```
tamtamchik: Maybe add contexts like in the doc? Vault owner + subroles, NO + subroles, anyone... would make review easier. wdyt?

tamtamchik (changes requested):
~~Not sure why the `lockfile` is changed. Please revert.~~ It is ok

Could not find `transferStakingVaultOwnership` testing scenarios (`Vote to transfer the ownership of Staking Vault` / `Confirm to transfer the ownership of Staking Vault`) that is a vital part of tests.

tamtamchik (approved):
🍏 LGTM!

PR #990, opened 2025-03-19
feat: ssz porting from csm

mkurayan:
Porting SSZ Library from CSM to Core Repository

Both `Triggerable Withdrawals` and `Vaults` require SSZ for Merkle proof verification.
This merge request aims to streamline development efforts by extracting and centralizing the shared SSZ-related code, which was originally ported from CSM.

### DIFF
--- a/contracts/0.8.25/lib/GIndex.sol
+++ b/contracts/0.8.25/lib/GIndex.sol
@@ -1,0 +1,122 @@
+// SPDX-FileCopyrightText: 2025 Lido <info@lido.fi>
+// SPDX-License-Identifier: GPL-3.0
+
+/*
+ GIndex library from CSM
+ original: https://github.com/lidofinance/community-staking-module/blob/7071c2096983a7780a5f147963aaa5405c0badb1/src/lib/GIndex.sol
+*/
+
+pragma solidity 0.8.25;
+
+type GIndex is bytes32;
+
+using {isRoot, isParentOf, index, width, shr, shl, concat, unwrap, pow} for GIndex global;
+
+error IndexOutOfRange();
+
+/// @param gI Is a generalized index of a node in a tree.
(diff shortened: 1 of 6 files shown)


contracts/0.8.25/lib/GIndex.sol
```diff
@@ -0,0 +1,117 @@
+// SPDX-FileCopyrightText: 2025 Lido <info@lido.fi>
+// SPDX-License-Identifier: GPL-3.0
+
+pragma solidity 0.8.25;
+
```
Jeday, line 10: Let's add coment with link to orignal lib liker here:
https://github.com/lidofinance/core/blob/0a51796aba1fe3617c6bf49d20efed65a26342d2/contracts/0.8.25/lib/GIndex.sol#L4

We want to outline 2 things:
 - "borrowed" but from lido
 -  already audited and unchanged

contracts/0.8.25/lib/Types.sol
Jeday, line 1: In context of this repo, I think it's better to name this SSZTypes (or similar) to outline this file will contain types relevant to ssz.

Jeday (dismissed):

contracts/0.8.25/lib/GIndex.sol
Jeday: same for ssz.sol

mkurayan: Comments added, thank!

contracts/0.8.25/lib/Types.sol
mkurayan: Good point that `Types` is to abstract name, I think to rename it to `BeaconTypes`.
So the name convey that the file holds fundamental data structures relevant to the Beacon Chain.

PR #994, opened 2025-03-20
[VAULTS] prevent minting against node operator fee

failingtwice:

## Context

## Problem

## Solution

Your proposed solution


contracts/0.8.25/vaults/Permissions.sol
folkyatina, line 1: Maybe, it's worth to concentrate this feature in two contracts instead of three? 
- it looks pretty alien to other functions in Permissions
- we can have a method like `_mintSharesButNotForFees()` in Dashboard and put this constraint there
- or even merge Delegation straight into the Dashboard and call it a day
WDYT?

failingtwice:
dropped in favor of https://github.com/lidofinance/core/pull/1011

PR #998, opened 2025-03-23
chore: fix incorrect constant for zero address in ethers.js

0xwitty:
## Context  
In ethers.js, the correct constant for the zero address is `ethers.constants.AddressZero`.  

## Problem  
The code was incorrectly using `ethers.ZeroAddress`, which does not exist in ethers.js.  

## Solution  
I replaced `ethers.ZeroAddress` with `ethers.constants.AddressZero` to ensure the correct usage of the zero address constant in the code.


arwer13:
The proposed change is not valid.

PR #1013, opened 2025-04-02
feat: rebalance shortfall

ev-d:
- Add checking isVaultHealth in rebalanceShortfall, if isVaultHealth is true rebalanceShortfall returns 0
- Add protection against underflow in division
- Add last cases that return type(uint256).max
- Add tests for updated logic

## Context
VaultHub


contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -194,7 +194,13 @@ contract VaultHub is PausableUntilWithRoles {
...
         // X = (mintedStETH * TOTAL_BASIS_POINTS - vault.valuation() * maxMintableRatio) / reserveRatio
 
-        return (mintedStETH * TOTAL_BASIS_POINTS - valuation * maxMintableRatio) / reserveRatioBP;
+        // Protection against underflow in division
+        if (!isHealthy && mintedStETH * TOTAL_BASIS_POINTS > valuation * maxMintableRatio) {
+            return (mintedStETH * TOTAL_BASIS_POINTS - valuation * maxMintableRatio) / reserveRatioBP;
+        }
+
+        // return MAX_UINT_256
+        return type(uint256).max;
```
folkyatina: This is not reachable, because `!isHealthy` is the same with `mintedStETH * TOTAL_BASIS_POINTS > valuation * maxMintableRatio` basically.

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -194,7 +194,13 @@ contract VaultHub is PausableUntilWithRoles {
         // reserveRatio = TOTAL_BASIS_POINTS - maxMintableRatio
         // X = (mintedStETH * TOTAL_BASIS_POINTS - vault.valuation() * maxMintableRatio) / reserveRatio
 
-        return (mintedStETH * TOTAL_BASIS_POINTS - valuation * maxMintableRatio) / reserveRatioBP;
+        // Protection against underflow in division
+        if (!isHealthy && mintedStETH * TOTAL_BASIS_POINTS > valuation * maxMintableRatio) {
```
failingtwice: There seems to be no reason to check for `!isHealthy` since there is a `isHealthy` check before that.

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -177,11 +177,11 @@ contract VaultHub is PausableUntilWithRoles {
         uint256 reserveRatioBP = socket.reserveRatioBP;
         uint256 maxMintableRatio = (TOTAL_BASIS_POINTS - reserveRatioBP);
         uint256 valuation = IStakingVault(_vault).valuation();
+        bool isHealthy = isVaultHealthy(_vault);
```
TheDZhon: let's move it upper and have a short circuit approach (return early)

contracts/0.8.25/vaults/VaultHub.sol
TheDZhon: agree

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -177,11 +177,11 @@ contract VaultHub is PausableUntilWithRoles {
...
 
-        // to avoid revert below
-        if (mintedStETH * TOTAL_BASIS_POINTS < valuation * maxMintableRatio) {
-            // return MAX_UINT_256
-            return type(uint256).max;
+        // Health vault do not need to rebalance
+        if (isHealthy) {
```
TheDZhon: let's structure this func this way:
- if healthy early `return 0`
- if 'unrebalanceable' early return +inf (i.e., bad debt): `mintedStETH > valuation`
- calculate otherwise

TheDZhon (changes requested):
Let's polish it a bit further :pray:

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -202,8 +208,8 @@ contract VaultHub is PausableUntilWithRoles {
         uint256 maxMintableRatio = (TOTAL_BASIS_POINTS - reserveRatioBP);
         uint256 valuation = IStakingVault(_vault).valuation();
 
-        // to avoid revert below
-        if (mintedStETH * TOTAL_BASIS_POINTS < valuation * maxMintableRatio) {
+        // Protection against underflow in division
```
```suggestion
        // Impossible to rebalance a vault with deficit
```

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -194,6 +194,12 @@ contract VaultHub is PausableUntilWithRoles {
     /// @return amount to rebalance
     function rebalanceShortfall(address _vault) public view returns (uint256) {
```
TheDZhon, line 206: A small note to justify the current impl:

`mintedStETH * TOTAL_BASIS_POINTS ≤ valuation * maxMintableRatio` is always **true** if the vault is **healthy** (because `valuation * maxMintableRatio ≤ valuation * (1 - rebalanceThresholdBP)`), 

so we really don't need to check for the underflow here if we have an early return for `isVaultHealthy`:
```solidity
return (mintedStETH * TOTAL_BASIS_POINTS - valuation * maxMintableRatio) / reserveRatioBP;
```

TheDZhon (approved):
LGTM, just a small comment

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -202,8 +208,8 @@ contract VaultHub is PausableUntilWithRoles {
         uint256 maxMintableRatio = (TOTAL_BASIS_POINTS - reserveRatioBP);
         uint256 valuation = IStakingVault(_vault).valuation();
 
-        // to avoid revert below
-        if (mintedStETH * TOTAL_BASIS_POINTS < valuation * maxMintableRatio) {
+        // Protection against underflow in division
+        if (mintedStETH > valuation) {
```
```suggestion
        if (mintedStETH >= valuation) {
```

PR #1016, opened 2025-04-07
[VAULTS] common SSZ commit

Jeday:
## Context

Common commit for SSZ lib in  TW and Vaults
 - SSZ and GIndex tests


tamtamchik (approved):
LGTM

PR #1026, opened 2025-04-10
VEB delivering requests by parts

Amuhar:

## Context

## Problem

## Solution

Your proposed solution


contracts/0.8.9/lib/ReportExitLimitUtils.sol
```diff
@@ -0,0 +1,158 @@
...
+            (uint256(_data.prevExitRequestsLimit) << PREV_EXIT_REQUESTS_LIMIT_OFFSET) |
+            (uint256(_data.maxExitRequestsLimitGrowthBlocks) << MAX_EXIT_REQUESTS_LIMIT_GROWTH_BLOCKS_OFFSET) |
+            (uint256(_data.maxExitRequestsLimit) << MAX_EXIT_REQUESTS_LIMIT_OFFSET)
+        );
+    }
+}
+
+library ReportExitLimitUtils {
+
+    error Debug(uint256 limit, uint256 block, uint256 prev);
```
mkurayan: It appears this error was only intended for debugging purposes and can likely be removed now.

contracts/0.8.9/lib/ReportExitLimitUtils.sol
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+
+    /**
+     * @notice find a minimum of two numbers with a constant gas consumption
+     * @dev doesn't use branching logic inside
+     * @param _lhs left hand side value
+     * @param _rhs right hand side value
+     */
+    function _constGasMin(uint256 _lhs, uint256 _rhs) internal pure returns (uint256 min) {
```
mkurayan, line 136: Why so complicated? Can we keep it simple?
`return _lhs < _rhs ? _lhs : _rhs;`

instead of current approach
```
        uint256 lhsIsLess;
        assembly {
            lhsIsLess := lt(_lhs, _rhs) // lhsIsLess = (_lhs < _rhs) ? 1 : 0
        }
        min = (_lhs * lhsIsLess) + (_rhs * (1 - lhsIsLess));
```

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -137,28 +238,160 @@ contract ValidatorsExitBus is AccessControlEnumerable {
...
+          if (limit == 0) {
+            revert ExitRequestsLimit();
+          }
+
+          EXIT_REQUEST_LIMIT_POSITION.setStorageExitRequestLimit(exitRequestLimitData.updatePrevExitRequestsLimit(limit - 1));
+        }
+
+        IWithdrawalVault(withdrawalVaultAddr).addFullWithdrawalRequests{value:  withdrawalFee}(validator.validatorPubkey);
+
+        emit ValidatorExitRequest(validator.stakingModuleId, validator.nodeOperatorId, validator.validatorIndex, validator.validatorPubkey, _getTimestamp());
```
mkurayan: This may have been discussed before, but I was wondering: why do we need to emit a `ValidatorExitRequest` event after we’ve already added a withdrawalRequest via EIP-7002?

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -137,28 +238,160 @@ contract ValidatorsExitBus is AccessControlEnumerable {
         assert(address(this).balance == prevBalance);
     }
 
+    function triggerExitsDirectly(ValidatorExitData calldata validator) external payable whenResumed onlyRole(DIRECT_EXIT_HASH_ROLE) returns (uint256)  {
+        uint256 prevBalance = address(this).balance - msg.value;
+        address locatorAddr = address(LOCATOR);
```
mkurayan: **Suggested improvement:**

You’re performing unnecessary type conversions (`ILidoLocator -> address -> ILidoLocator`). Because `LOCATOR` is already defined as an `ILidoLocator`, you can directly call the method on `LOCATOR`:

```solidity
// Instead of:
address locatorAddr = address(LOCATOR);
address withdrawalVaultAddr = ILidoLocator(locatorAddr).withdrawalVault();

// Just do:
address withdrawalVaultAddr = LOCATOR.withdrawalVault();
```

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -48,41 +67,122 @@ contract ValidatorsExitBus is AccessControlEnumerable {
...
-        RequestStatus storage requestStatus = _storageExitRequestsHashes()[keccak256(abi.encode(request.data, request.dataFormat))];
         bytes calldata data = request.data;
+        RequestStatus storage requestStatus = _storageExitRequestsHashes()[keccak256(abi.encode(data, request.dataFormat))];
 
         if (requestStatus.contractVersion == 0) {
           revert ExitHashWasNotSubmitted();
         }
 
-        address locatorAddr = LOCATOR_CONTRACT_POSITION.getStorageAddress();
+        address locatorAddr = address(LOCATOR);
```
mkurayan: **Suggested improvement:**

You’re performing unnecessary type conversions (`ILidoLocator -> address -> ILidoLocator`). Because `LOCATOR` is already defined as an `ILidoLocator`, you can directly call the method on `LOCATOR`:

```solidity
// Instead of:
address locatorAddr = address(LOCATOR);
address withdrawalVaultAddr = ILidoLocator(locatorAddr).withdrawalVault();

// Just do:
address withdrawalVaultAddr = LOCATOR.withdrawalVault();
```

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -48,41 +67,122 @@ contract ValidatorsExitBus is AccessControlEnumerable {
...
     // and ensures that the events for the requests specified in the `keyIndexes` array have already been delivered.
-    function triggerExits(ExitRequestData calldata request, uint256[] calldata keyIndexes) external payable {
+    function triggerExits(ExitRequestData calldata request, uint256[] calldata keyIndexes) external payable whenResumed {
         uint256 prevBalance = address(this).balance - msg.value;
-        RequestStatus storage requestStatus = _storageExitRequestsHashes()[keccak256(abi.encode(request.data, request.dataFormat))];
         bytes calldata data = request.data;
+        RequestStatus storage requestStatus = _storageExitRequestsHashes()[keccak256(abi.encode(data, request.dataFormat))];
 
         if (requestStatus.contractVersion == 0) {
           revert ExitHashWasNotSubmitted();
```
mkurayan, line 182: Consider using an automated formatting tool for this file to ensure consistent indentation and spacing. There are several places where the spacing varies (two, three, or four spaces)

mkurayan: Consider using an automated formatting tool for ValidatorsExitBus.sol file to ensure consistent indentation and spacing. There are several places where the spacing varies (two, three, or four spaces)

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -48,41 +67,122 @@ contract ValidatorsExitBus is AccessControlEnumerable {
...
+    function submitReportHash(bytes32 exitReportHash) external whenResumed onlyRole(SUBMIT_REPORT_HASH_ROLE) {
+      uint256 contractVersion = getContractVersion();
+      _storeExitRequestHash(exitReportHash,  type(uint256).max, 0, contractVersion, DeliveryHistory(0,0));
+    }
 
-    function _setLocatorAddress(address addr) internal {
-        if (addr == address(0)) revert ZeroAddress();
+    function emitExitEvents(ExitRequestData calldata request, uint256 contractVersion) external whenResumed {
+        bytes calldata data = request.data;
+        _checkContractVersion(contractVersion);
```
mkurayan: It appears `contractVersion` is already stored in the exit request metadata and retrieved just a few lines below:

```solidity
struct RequestStatus {
    // Vebo contract version at the time of hash submission
    uint256 contractVersion;
}

RequestStatus storage requestStatus = _storageExitRequestsHashes()[keccak256(abi.encode(data, request.dataFormat))];
```

Given that `contractVersion` is included in `requestStatus`, why trust a permissionless method’s user input for the version? Instead, it would be safer to verify the `contractVersion` directly from the `requestStatus`. This avoids relying on external calls to pass the correct version.

contracts/0.8.9/lib/ReportExitLimitUtils.sol
loga4: Yeah, `return _lhs < _rhs ? _lhs : _rhs;` is simpler, but it introduces branching, so the compiler adds JUMP (jumpi) opcodes – gas cost depends on which path is taken.

With `constGasMin`, there’s no branching, just math. That means gas is always the same, which is better for predictability.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -137,28 +238,160 @@ contract ValidatorsExitBus is AccessControlEnumerable {
         assert(address(this).balance == prevBalance);
     }
 
+    function triggerExitsDirectly(ValidatorExitData calldata validator) external payable whenResumed onlyRole(DIRECT_EXIT_HASH_ROLE) returns (uint256)  {
```
mkurayan: We probably need to call `onValidatorExitTriggered` hook during direct exit call?

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -48,41 +67,122 @@ contract ValidatorsExitBus is AccessControlEnumerable {
...
+
+        requestStatus.deliverHistory.push(DeliveryHistory(deliveredItemsCount + toDeliver - 1, _getTimestamp()));
+        requestStatus.deliveredItemsCount += toDeliver;
     }
 
     /// @notice Triggers exits on the EL via the Withdrawal Vault contract after
     /// @dev This function verifies that the hash of the provided exit request data exists in storage
     // and ensures that the events for the requests specified in the `keyIndexes` array have already been delivered.
-    function triggerExits(ExitRequestData calldata request, uint256[] calldata keyIndexes) external payable {
+    function triggerExits(ExitRequestData calldata request, uint256[] calldata keyIndexes) external payable whenResumed {
```
mkurayan: We probably need to call `onValidatorExitTriggered` hook during trigger exit call?

Amuhar: yes, but in direct method

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -48,42 +75,121 @@ contract ValidatorsExitBus is AccessControlEnumerable {
...
         keccak256("lido.ValidatorsExitBus.reportHashes");
 
-    bytes32 private constant LOCATOR_CONTRACT_POSITION = keccak256("lido.ValidatorsExitBus.locatorContract");
+    constructor(address lidoLocator) {
+        LOCATOR = ILidoLocator(lidoLocator);
+    }
+
+    function submitReportHash(bytes32 exitReportHash) external whenResumed onlyRole(SUBMIT_REPORT_HASH_ROLE) {
+      uint256 contractVersion = getContractVersion();
+      _storeExitRequestHash(exitReportHash,  type(uint256).max, 0, contractVersion, DeliveryHistory(0,0));
```
mkurayan: It can be expensive to store `DeliveryHistory(0,0)` entries when they don’t provide any useful information. We can optimize the code by simply not storing those entries.

contracts/0.8.9/interfaces/IValidatorExitBus.sol
```diff
@@ -0,0 +1,47 @@
...
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity 0.8.9;
+
+interface IValidatorsExitBus {
+    struct ExitRequestData {
+        bytes data;
+        uint256 dataFormat;
+    }
+
+    struct DirectExitData {
```
mkurayan, line 11: It might be a good time to consider storage optimization for this structure.

We discussed the idea in this thread previosly.
https://github.com/lidofinance/core/pull/927#discussion_r1963690288

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -137,28 +244,218 @@ contract ValidatorsExitBus is AccessControlEnumerable {
...
+            // MSB <---------------------------------------------------------------------- LSB
+            // | 128 bits: zeros | 24 bits: moduleId | 40 bits: nodeOpId | 64 bits: valIndex |
+            uint64 valIndex = uint64(dataWithoutPubkey);
+            uint256 nodeOpId = uint40(dataWithoutPubkey >> 64);
+            uint256 moduleId = uint24(dataWithoutPubkey >> (64 + 40));
+
+            if (moduleId == 0) {
+                revert InvalidRequestsData();
+            }
+
```
mkurayan, line 485: In the original VEBO contract code (from which this was migrated), there was an additional sanity check ensuring that the reported data was sorted by operator and module. If the data was unsorted, the code would revert with a `NodeOpValidatorIndexMustIncrease` error. 

Why did we decide to remove this check?

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -48,42 +75,121 @@ contract ValidatorsExitBus is AccessControlEnumerable {
...
+        requestStatus.deliverHistory.push(DeliveryHistory(deliveredItemsCount + toDeliver - 1, _getTimestamp()));
+        requestStatus.deliveredItemsCount += toDeliver;
     }
 
     /// @notice Triggers exits on the EL via the Withdrawal Vault contract after
     /// @dev This function verifies that the hash of the provided exit request data exists in storage
     // and ensures that the events for the requests specified in the `keyIndexes` array have already been delivered.
-    function triggerExits(ExitRequestData calldata request, uint256[] calldata keyIndexes) external payable {
+    function triggerExits(ExitRequestData calldata request, uint256[] calldata keyIndexes) external payable whenResumed {
         uint256 prevBalance = address(this).balance - msg.value;
```
mkurayan: In the Withdrawal Vault contract, the check for preserving balance was relocated to a modifier. You might consider using a similar approach.

```
    /// @dev Ensures the contract’s ETH balance is unchanged.
    modifier preservesEthBalance() {
        uint256 balanceBeforeCall = address(this).balance - msg.value;
        _;
        assert(address(this).balance == balanceBeforeCall);
    }
```

contracts/0.8.9/oracle/ValidatorsExitBus.sol
Amuhar: i was mistaken, it should be in both methods

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -48,42 +75,121 @@ contract ValidatorsExitBus is AccessControlEnumerable {
...
-        RequestStatus storage requestStatus = _storageExitRequestsHashes()[keccak256(abi.encode(request.data, request.dataFormat))];
         bytes calldata data = request.data;
+        RequestStatus storage requestStatus = _storageExitRequestsHashes()[keccak256(abi.encode(data, request.dataFormat))];
 
         if (requestStatus.contractVersion == 0) {
           revert ExitHashWasNotSubmitted();
         }
```
mkurayan, line 203: Consider adding request data format validation, ensuring that the data is provided as a list—similar to what you did in `emitExitEvents`. My point is that both methods should either validate the format or not validate it at all, so the logic remains consistent.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -137,28 +244,218 @@ contract ValidatorsExitBus is AccessControlEnumerable {
         assert(address(this).balance == prevBalance);
     }
 
+    function triggerExitsDirectly(DirectExitData calldata exitData) external payable whenResumed onlyRole(DIRECT_EXIT_HASH_ROLE) returns (uint256)  {
```
mkurayan: We need to improve the permission system so that a module can only perform direct exits for its own node operators.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -22,22 +28,43 @@ contract ValidatorsExitBus is AccessControlEnumerable {
...
+        uint256 _exitRequestsLimitIncreasePerBlock
+    );
+
+    event DirectExitRequest(
+        uint256 indexed stakingModuleId,
+        uint256 indexed nodeOperatorId,
+        bytes validatorsPubkeys,
+        uint256 timestamp
+    );
     struct RequestStatus {
```
mkurayan, line 68: It might be a good time to consider storage optimization for this structure.

We discussed the idea in this thread previosly.
https://github.com/lidofinance/core/pull/927#discussion_r1963690288

contracts/0.8.9/oracle/ValidatorsExitBus.sol
Amuhar: Since we no longer need to return the last validators, and there are no sorting requirements anymore, validators can now be exited without any specific order—so this sorting is no longer needed.

contracts/0.8.9/lib/ReportExitLimitUtils.sol
krogla: saving on matches :)
in general the contract is very costly in terms of gas, so +10gas will not play a role at all

contracts/0.8.9/lib/ReportExitLimitUtils.sol
```diff
@@ -0,0 +1,155 @@
...
+
+library ReportExitLimitUtils {
+    /**
+    * @notice Calculate exit requests limit
+    * @dev using `_constGasMin` to make gas consumption independent of the current block number
+    */
+    function calculateCurrentExitRequestLimit(ExitRequestLimitData memory _data) internal view returns(uint256 limit) {
+        uint256 exitRequestLimitIncPerBlock;
+        if (_data.maxExitRequestsLimitGrowthBlocks != 0) {
+            exitRequestLimitIncPerBlock = _data.maxExitRequestsLimit / _data.maxExitRequestsLimitGrowthBlocks;
```
krogla, line 56: wouldn't it be easier to store just the value of `exitRequestLimitIncPerBlock`?
since in `setExitReportLimit` we first recalculate it to `maxExitRequestsLimitGrowthBlocks`, and then back here.
so we can just remove the whole `if` block

contracts/0.8.9/oracle/ValidatorsExitBus.sol
krogla: I agree, completely useless var and check, since `contractVersion` is not used anywhere else in the code, it can be removed at all

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -22,81 +27,182 @@ contract ValidatorsExitBus is AccessControlEnumerable {
...
+
+        RequestStatus storage requestStatus = _storageExitRequestsHashes()[
+            keccak256(abi.encode(data, request.dataFormat))
+        ];
+
+        if (requestStatus.contractVersion == 0) {
+            revert ExitHashWasNotSubmitted();
+        }
+
+        if (request.dataFormat != DATA_FORMAT_LIST) {
```
krogla, line 143: in the context of the comment about `contractVersion` above, is there a case when the `DATA_FORMAT_LIST` constant is changed without changing the contract version? i.e. it seems that the simultaneous presence of both checks is a bit redundant

contracts/0.8.9/oracle/ValidatorsExitBus.sol
krogla: +1

contracts/0.8.9/oracle/ValidatorsExitBus.sol
krogla: the sorting mechanism was there to exclude duplicates and repeated processing in the code. 
what will happen if a request has 5 identical keys in a row?

contracts/0.8.9/oracle/ValidatorsExitBus.sol
Amuhar: decided to not trigger ValidatorExitRequest event in this method

contracts/0.8.9/lib/ReportExitLimitUtils.sol
loga4: i just explained how it works)

PR #1039, opened 2025-04-17
fix: vault unlocked logic

dry914:
Change `vault.unlocked()` logic - if report is not fresh, use `address(this).balance` instead of `valuation()`

### DIFF
--- a/contracts/0.8.25/vaults/StakingVault.sol
+++ b/contracts/0.8.25/vaults/StakingVault.sol
@@ -292,10 +292,10 @@ contract StakingVault is IStakingVault, OwnableUpgradeable {
     /**
      * @notice Returns the unlocked amount of ether, which is the valuation minus the locked ether amount
      * @dev Unlocked amount is the total amount that can be withdrawn from `StakingVault`,
-     *      including ether currently being staked on validators
+     *      including ether currently being staked on validators only if the report is fresh
      */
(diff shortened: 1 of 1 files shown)


TheDZhon (changes requested):
:thinking:

contracts/0.8.25/vaults/StakingVault.sol
```diff
@@ -416,12 +416,6 @@ contract StakingVault is IStakingVault, OwnableUpgradeable {
         (bool success, ) = _recipient.call{value: _ether}("");
         if (!success) revert TransferFailed(_recipient, _ether);
 
-        if (isReportFresh()) {
-            if (valuation() < $.locked) revert ValuationBelowLockedAmount();
-        } else {
-            if (address(this).balance < $.locked) revert ValuationBelowLockedAmount();
-        }
```
TheDZhon, line 423: :warning: It might be dangerous as this check intentionally had been existing even before the lazy oracle was introduced (e.g., https://github.com/lidofinance/core/blob/45b27dde842316a18373606b432e2b19d14e7f0a/contracts/0.8.25/vaults/StakingVault.sol#L396)

dry914: I agree, I missed that. Thanks for pointing it out.

PR #1042, opened 2025-04-21
Disallow empty codehash for vaults

folkyatina:
EXTCODEHASH returns keccak(0x) if account exists but has no code sometimes, so we revert on adding it.


tamtamchik (approved):
👏

TheDZhon (approved):
LGTM

PR #1049, opened 2025-04-23
feat: tw nor implementation

eddort:


contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -1182,7 +1080,22 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
         bytes, // _publicKey
         uint256 _eligibleToExitInSec
     ) external view returns (bool) {
-         return _eligibleToExitInSec >= _getExitDeadlineThreshold();
+        return _eligibleToExitInSec >= exitDeadlineThreshold();
```
F4ever, line 1083: If validator was already marked as late, return false here

F4ever (approved):

PR #1051, opened 2025-04-24
fix: refund receipient added & exit type & extract val requests parsing

Amuhar:

## Context

## Problem

## Solution

Your proposed solution


contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -574,6 +594,66 @@ contract ValidatorsExitBus is IValidatorsExitBus, AccessControlEnumerable, Pausa
...
+        // dataWithoutPubkey format (128 bits total):
+        // MSB <-------------------- 128 bits --------------------> LSB
+        // |   128 bits: zeros  | 24 bits: moduleId | 40 bits: nodeOpId | 64 bits: valIndex |
+
+        validatorData.valIndex = uint64(dataWithoutPubkey);
+        validatorData.nodeOpId = uint40(dataWithoutPubkey >> 64);
+        validatorData.moduleId = uint24(dataWithoutPubkey >> (64 + 40));
+
+        bytes memory pubkey = new bytes(PUBLIC_KEY_LENGTH);
+        assembly {
```
mkurayan, line 631: If you not going to reuse the same chunk of memory as it was before, you can simplify this code and just use slice of calldata operator. `exitRequestData[start:end]`

But be aware of memory explosion. It might cost a lot.

PR #1052, opened 2025-04-25
[VAULTS] update role checks to include admin permissions for various…

failingtwice:
Any operation that requires a role clearance can now be performed by the role admin

## Context

Any operation requires an explicit role.

## Problem

It makes sense to give the role admin the permission by default because we can skip the step of explicitly granting role. This saves gas and does not reduce security.

## Solution

use a custom modification of `onlyRole` which includes an additional check for role admin.


test/integration/vaults/roles.integration.ts
```diff
@@ -652,7 +667,7 @@ describe("Integration: Staking Vaults Dashboard Roles Initial Setup", () => {
       }
     });
 
-    describe("Verify ACL for methods that require only role", () => {
+    describe.skip("Verify ACL for methods that require only role", () => {
```
tamtamchik, line 670: Maybe remove this test at all?

tamtamchik (approved):
LGTM

contracts/0.8.25/vaults/dashboard/Permissions.sol
```diff
@@ -228,11 +228,23 @@ abstract contract Permissions is AccessControlConfirmable {
      */
     function confirmingRoles() public pure virtual returns (bytes32[] memory);
 
+    /**
+     * @dev A custom modifier that checks if the caller has a role or the admin role for a given role.
+     * @param _role The role to check.
+     */
+    modifier onlyRoleMemberOrAdmin(bytes32 _role) {
+        if (hasRole(_role, msg.sender) || hasRole(getRoleAdmin(_role), msg.sender)) {
```
TheDZhon, line 236: the only question is order here: what would be more favorable due to the short-circuiting calculations:
- to assign granular roles and pay less gas
- to not assign granular roles and pay less gas if put other way around

TheDZhon (approved):
LGTM :clap:

contracts/0.8.25/vaults/dashboard/Permissions.sol
folkyatina: I'd prefer the favor the granular roles, just to incentivize more secure setup

PR #1053, opened 2025-04-25
Add more fields to vaults batch query

vp4242:
Add more fields to vaults batch query

### DIFF
--- a/contracts/0.8.25/vaults/VaultHub.sol
+++ b/contracts/0.8.25/vaults/VaultHub.sol
@@ -69,6 +69,11 @@ contract VaultHub is PausableUntilWithRoles {
         int256 inOutDelta;
         bytes32 withdrawalCredentials;
         uint256 liabilityShares;
+        uint96 shareLimit;
+        uint16 reserveRatioBP;
+        uint16 forcedRebalanceThresholdBP;
+        uint16 treasuryFeeBP;
+        bool pendingDisconnect;
     }
 
     // keccak256(abi.encode(uint256(keccak256("VaultHub")) - 1)) & ~bytes32(uint256(0xff))
@@ -190,7 +195,12 @@ contract VaultHub is PausableUntilWithRoles {
                 address(currentVault).balance,
                 currentVault.inOutDelta(),
                 currentVault.withdrawalCredentials(),
-                socket.liabilityShares
+                socket.liabilityShares,
+                socket.shareLimit,
+                socket.reserveRatioBP,
+                socket.forcedRebalanceThresholdBP,
+                socket.treasuryFeeBP,
+                socket.pendingDisconnect
             );
         }
         return batch;
--- a/contracts/0.8.25/vaults/VaultHub.sol
+++ b/contracts/0.8.25/vaults/VaultHub.sol
@@ -183,10 +183,10 @@ contract VaultHub is PausableUntilWithRoles {
     /// @param _offset offset of the vault in the batch (indexes start from 0)
     /// @param _limit limit of the batch
     /// @return batch of vaults info
-    function batchVaultsInfo(uint256 _offset, uint256 _limit) external view returns (VaultInfo[] memory) {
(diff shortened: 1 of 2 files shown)


contracts/0.8.25/vaults/VaultHub.sol
TheDZhon, line 1: It looks good in general, but have two questions from my side:
- why do we expose more fields?
- is it okay that we have a clash of name semantics with OperatorGrid and its returning structure? (see: https://github.com/lidofinance/core/blob/8d3f32a4e9e469eed1f9fc541b3c07d2f375829f/contracts/0.8.25/vaults/VaultHub.sol#L2820

TheDZhon (commented):
:eyes:

contracts/0.8.25/vaults/VaultHub.sol
vp4242: The link seems to be broken. What is it about? I don't see the similar struct for OperatorGrid

TheDZhon: https://github.com/lidofinance/core/blob/8d3f32a4e9e469eed1f9fc541b3c07d2f375829f/contracts/0.8.25/vaults/VaultHub.sol#L282 sorry, extra zero

vp4242: > why do we expose more fields?

More fields needed for fees calculations in Oracle.


Regarding OperatorGrid. My impression that OG supply the initial values that could be changed later on per vault basis. That effectively mean that better to take this values from vaults themselves. The other reason is that it's easier for Oracle to read all vaults related values from single function. The implementation could take the values from OG if it's needed.

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -190,7 +195,12 @@ contract VaultHub is PausableUntilWithRoles {
                 address(currentVault).balance,
                 currentVault.inOutDelta(),
                 currentVault.withdrawalCredentials(),
-                socket.liabilityShares
+                socket.liabilityShares,
```
TheDZhon, line 199: :brain:  Let's load socket to memory (can't comment on the line above), i.e.
```
- VaultSocket storage socket = $.sockets[i + 1 + _offset];
+ VaultSocket memory socket = $.sockets[i + 1 + _offset];
```

we're overspending gas atm

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -69,6 +69,11 @@ contract VaultHub is PausableUntilWithRoles {
         int256 inOutDelta;
         bytes32 withdrawalCredentials;
         uint256 liabilityShares;
+        uint96 shareLimit;
```
TheDZhon, line 72: not sure but might embed VaultSocket field right here instead (not a strong opinion though)

TheDZhon (changes requested):
:+1: A little nitpick

contracts/0.8.25/vaults/VaultHub.sol
vp4242: Ok. let's change it back :)

image

TheDZhon: Fair, that was good when we had only a single read from the socket, now the situation is quite opposite :)

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -182,15 +187,21 @@ contract VaultHub is PausableUntilWithRoles {
         VaultHubStorage storage $ = _getVaultHubStorage();
         uint256 limit = _offset + _limit > $.sockets.length - 1 ? $.sockets.length - 1 - _offset : _limit;
         VaultInfo[] memory batch = new VaultInfo[](limit);
+        uint256 startIndex = _offset + 1;
```
TheDZhon, line 190: can declare it inside return :thought_balloon: (might save bytecode a bit probably)

TheDZhon (commented):
:eyes:

PR #1058, opened 2025-04-29
Move StakingVault accounting to VaultHub 

failingtwice:
image

PR #1063, opened 2025-05-05
Lazy oracle freshness conditions update

### LINKED ISSUE #1059: stVaults: Accounting issue with outdated report check
https://github.com/lidofinance/core/blob/653a4e3720e48daf7e143ea7667d44778885b061/contracts/0.8.25/vaults/StakingVault.sol#L409-L413

The code contains a potential vuln, let's prohibit withdrawing with outdated report still allowing to withdraw when the vault is detached from its balance.

Tests are needed 👀

vp4242:
Lazy oracle freshness conditions update

### DIFF
--- a/contracts/0.8.25/vaults/StakingVault.sol
+++ b/contracts/0.8.25/vaults/StakingVault.sol
@@ -650,7 +650,8 @@ contract StakingVault is IStakingVault, OwnableUpgradeable {
     function isReportFresh() public view returns (bool) {
         ERC7201Storage storage $ = _getStorage();
         if (!$.vaultHubAuthorized) return true;
-        return block.timestamp - $.report.timestamp < VAULT_HUB.REPORT_FRESHNESS_DELTA();
+        (uint64 latestReportTimestamp,,) = VAULT_HUB.latestReportData();
+        return latestReportTimestamp == $.report.timestamp && block.timestamp - $.report.timestamp < VAULT_HUB.REPORT_FRESHNESS_DELTA();
     }
 
     function _getStorage() private pure returns (ERC7201Storage storage $) {
--- a/contracts/0.8.25/vaults/StakingVault.sol
+++ b/contracts/0.8.25/vaults/StakingVault.sol
@@ -651,7 +651,7 @@ contract StakingVault is IStakingVault, OwnableUpgradeable {
         ERC7201Storage storage $ = _getStorage();
         if (!$.vaultHubAuthorized) return true;
         (uint64 latestReportTimestamp,,) = VAULT_HUB.latestReportData();
-        return latestReportTimestamp == $.report.timestamp && block.timestamp - $.report.timestamp < VAULT_HUB.REPORT_FRESHNESS_DELTA();
+        return latestReportTimestamp == $.report.timestamp && block.timestamp - latestReportTimestamp < VAULT_HUB.REPORT_FRESHNESS_DELTA();
     }
 
     function _getStorage() private pure returns (ERC7201Storage storage $) {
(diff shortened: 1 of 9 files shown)


contracts/0.8.25/vaults/StakingVault.sol
```diff
@@ -652,12 +650,8 @@ contract StakingVault is IStakingVault, OwnableUpgradeable {
...
         ERC7201Storage storage $ = _getStorage();
         if (!$.vaultHubAuthorized) return true;
-        return block.timestamp - $.report.timestamp < VAULT_HUB.REPORT_FRESHNESS_DELTA();
-    }
-
-    function _checkFreshnessAndGetTotalValue() internal view returns (uint256) {
-        if (!isReportFresh()) revert ReportStaled();
-        return totalValue();
+        (uint64 latestReportTimestamp,,) = VAULT_HUB.latestReportData();
+        return latestReportTimestamp == $.report.timestamp && block.timestamp - $.report.timestamp < VAULT_HUB.REPORT_FRESHNESS_DELTA();
```
TheDZhon: gas optimization

```suggestion
        return latestReportTimestamp == $.report.timestamp && block.timestamp - latestReportTimestamp < VAULT_HUB.REPORT_FRESHNESS_DELTA();
```

contracts/0.8.25/vaults/StakingVault.sol
```diff
@@ -406,11 +406,9 @@ contract StakingVault is IStakingVault, OwnableUpgradeable {
         (bool success, ) = _recipient.call{value: _ether}("");
         if (!success) revert TransferFailed(_recipient, _ether);
 
-        if (isReportFresh()) {
-            if (totalValue() < $.locked) revert TotalValueBelowLockedAmount();
-        } else {
-            if (address(this).balance < $.locked) revert TotalValueBelowLockedAmount();
-        }
+        if (!isReportFresh()) revert ReportStaled();
```
TheDZhon, line 409: :warning: we shall allow withdrawing the whole balance if the vault is deauthorized regardless the report freshness and even unlocked amount probably, i.e., just a shortcut:
```solidity
function withdraw(address recipient, uint256 amount) external onlyOwner {    
    if (recipient == address(0)) revert ZeroArgument("_recipient");
    if (amount == 0) revert ZeroArgument("_ether");

    uint256 balance = address(this).balance;
    if (amount > balance) revert InsufficientBalance(balance);

    bool authorised = vaultHubAuthorized();
    ERC7201Storage storage $ = _getStorage();

    if (authorised) {
        if (!isReportFresh()) revert ReportStaled();

        uint256 unlockedAmt = unlocked();
        if (amount > unlockedAmt) revert InsufficientUnlocked(unlockedAmt);

        $.inOutDelta -= int128(int256(amount));
    }
    
    (bool ok, ) = recipient.call{value: amount}("");
    if (!ok) revert TransferFailed(recipient, amount);
    
    if (authorised && totalValue() < $.locked) revert TotalValueBelowLockedAmount();

    emit Withdrawn(msg.sender, recipient, amount);
}
 ```

TheDZhon (changes requested):
:clap: would suggest to handle the deauth case

contracts/0.8.25/vaults/StakingVault.sol
vp4242: isReportFresh() contains check for deauthorized VaultHub. Isn't this not enough?

TheDZhon: The issue I see is that `totalValue` might prohibit the vault owner to withdraw the funds arrived later than the lastly reported data

TheDZhon (changes requested):
1) Let's abandon changes for the `withdraw` function as @failingtwice will handle it as a part of the refactoring effort in #1058 
2) Let's merge the freshness part with a slight change suggested

:articulated_lorry:

PR #1105, opened 2025-05-26
feat(VaultHub): disable receive() by default

failingtwice:
Disables direct ether transfer to VaultHub. Enabled only on rebalances


failingtwice:
Not needed

PR #1126, opened 2025-05-29
Atomic exit requests unpack

mkurayan:


contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -599,68 +546,46 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
         emit ExitRequestsLimitSet(maxExitRequestsLimit, exitsPerFrame, frameDurationInSec);
     }
 
-    function _consumeLimit(uint256 requestsCount, function(uint256, uint256) internal pure returns(uint256) applyLimit) internal returns (uint256 requestsLimitedCount) {
+    function _consumeLimit(uint256 requestsCount) internal returns (uint256 requestsLimitedCount) {
```
mkurayan: _consumeLimit internal implementation might be simplified a little bit because we do not need return value any more.

Amuhar: fixed

F4ever (approved):

PR #1127, opened 2025-05-29
Rename error & try catch for notification

Amuhar:


mkurayan (approved):

contracts/0.8.9/TriggerableWithdrawalsGateway.sol
```diff
@@ -233,7 +232,7 @@ contract TriggerableWithdrawalsGateway is AccessControlEnumerable {
                     withdrawalRequestPaidFee,
                     exitType
              )
-            {} catch {
+            {} catch { // (bytes memory lowLevelRevertData)
```
loga4: yep, need to check lowLevelRevertDatafor empty data:
```suggestion
            {} catch (bytes memory lowLevelRevertData) {
                /// @dev This check is required to prevent incorrect gas estimation of the method.
                ///      Without it, Ethereum nodes that use binary search for gas estimation may
                ///      return an invalid value when the onValidatorExitTriggered() reverts because of the
                ///      "out of gas" error. Here we assume that the onValidatorExitTriggered() method doesn't
                ///      have reverts with empty error data except "out of gas".
                if (lowLevelRevertData.length == 0) revert UnrecoverableModuleError();
```

PR #1159, opened 2025-06-11
fix(NodeOperatorFee): revert NO fee change if vault is in quarantine

failingtwice:
Reverts NO operator fee change if the vault is in quarantine because the total value in the report may not reflect the adjustment.


TheDZhon (approved):
LGTM

PR #1162, opened 2025-06-12
fix: storage + remove connection condition in OG

loga4:
- fix storage location variables in OG and Confirmations
- remove `isVaultConnected` condition
- added `_requestedShareLimit` to event
- added dashboard check on `abandonDashboard`


TheDZhon (approved):
LGTM

PR #1204, opened 2025-06-24
fix: bump accounting oracle initial version to 3

eddort:


F4ever (approved):

contracts/0.8.9/oracle/AccountingOracle.sol
```diff
@@ -178,13 +179,18 @@ contract AccountingOracle is BaseOracle {
         _initialize(admin, consensusContract, consensusVersion, lastProcessingRefSlot);
 
         _updateContractVersion(2);
+        _updateContractVersion(3);
     }
 
     function finalizeUpgrade_v2(uint256 consensusVersion) external {
```
mkurayan, line 185: Consider to comment this or remove. In other contracts we usually we keep only latest finalize version.

mkurayan (approved):

PR #1215, opened 2025-07-02
Fix vault freshness check and repeating reports

### LINKED ISSUE #1190: [Bug]: vault creation during the report collection interval makes report not fresh
Tenderly example: https://www.tdly.co/shared/simulation/565e6119-b9f7-46d9-a662-92731001a648

If vault is created during report collection:
 - it's not included in IPFS report
 - isReportFresh: **false**

This results in DOS situation where until next oracle report is made no vault operations like mint/withdrawal are available because:
 - report cannot be applied 
 - operations require fresh report

### LINKED ISSUE #1200: [Bug]: Repetitive vault reports allowed at the same reference slot
### Summary

The `_applyVaultReport` function must check that this report has been already applied (https://github.com/lidofinance/core/blob/e37d22460231733fe560c125b0403be9d851307e/contracts/0.8.25/vaults/VaultHub.sol#L1026-L1048) 

### Expected Behavior

Revert when repetitive report attempted

### Potential Impact

Quarantine bypass/disruption probably (~MED)

### Steps to Reproduce

Send the same report >1 times through the `LazyOracle.updateVaultData`

### Possible Solutions

Add a missing check (existed before: https://github.com/lidofinance/core/pull/977/files)

```solidity
uint64 currentTimestamp = $.report.timestamp;
if (currentTimestamp >= _timestamp) revert ReportTooOld(currentTimestamp, _timestamp);

$.report.timestamp = _timestamp;
```

### Guidelines

### LINKED ISSUE #1292: [Bug]: A vault can submit the outdated report after reconnection
### Summary

Any vault can disconnect and then reconnect to the `VaultHub`. The disconnection process requires a 2-step approach:

1. Initiate disconnection.
2. Apply the latest report to finalise the disconnection. 

These steps can be completed in two consecutive calls. After this is done, the vault can also reconnect instantly. After the reconnection, the vault can apply the latest report right away. The new oracle report is submitted once a day. Since all previous actions can be completed in one transaction, the latest oracle report will still be the same as it was before we disconnected. I.e., representing the total value of the vault before the disconnection.

So let's imagine the following steps:

1. The vault has 100 ETH, the latest report also states there's 100 ETH.
2. The vault disconnects in one transaction as we described before. 
3. The disconnected vault withdraws 99 ETH, so there's now only 1 ETH there.
4. The vault reconnects with only 1 ETH. However, the latest report remains unchanged, stating that there is 100 ETH.
5. It's possible to apply the old report stating the vault has 100 ETH, while the vault only has 1 ETH. 

The 99 ETH difference is supposed to go to the quarantine, but from the other issue, we know how to avoid quarantines.

_**Note**_: In this flow, we applied the same report twice, using another issue. The first instance occurred during the disconnection, and the second instance occurred after the vault was reconnected. But it's not mandatory to do it. The attacker can also wait for the new oracle report, frontrun it, and disconnect right before, still using the previous report. That way, the new report will also have the total value of 100 ETH and can be used after reconnecting.

### Expected Behavior

The old report should not be submitted for a reconnected vault.

### Potential Impact

Stealing funds from Lido.

### Steps to Reproduce

...

### Possible Solutions

Ensure the "old" report cannot be applied after the vault is reconnected.

### Guidelines

folkyatina:
Fix #1190
Fix #1200 
Fix #1292


contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -1188,11 +1188,23 @@ contract VaultHub is PausableUntilWithRoles {
...
-            // if Accounting Oracle stop bringing the report, last report is fresh for 2 days
-            block.timestamp - latestReportTimestamp < REPORT_FRESHNESS_DELTA;
+        uint32 latestReportTimestamp32 = uint32(latestReportTimestamp);
+        uint32 vaultReportTimestamp = _record.report.timestamp;
+
+        // happy path. vault's report ts is the same as the latest AO report ts
+        if (latestReportTimestamp32 == vaultReportTimestamp) {
+            // check if AO has not stopped bringing the report
+            return block.timestamp - latestReportTimestamp < REPORT_FRESHNESS_DELTA;
+        }
```
```suggestion

```

:hocho:

contracts/common/lib/Math256.sol
```diff
@@ -41,4 +41,14 @@ library Math256 {
     function absDiff(uint256 a, uint256 b) internal pure returns (uint256) {
         return a > b ? a - b : b - a;
     }
+
+    /// @dev Returns the distance between two 32-bit numbers in counterclockwise direction
+    /// @param a the first number that's suppose to be larger than b
+    /// @param b the second number that's suppose to be smaller than a
+    /// @return the distance between a and b in counterclockwise direction modulo 2**32 + 1
```
TheDZhon:
```suggestion
    /// @return the distance between a and b in counterclockwise direction modulo 2**32
```

contracts/common/lib/Math256.sol
```diff
@@ -41,4 +41,14 @@ library Math256 {
...
+    /// @dev this function is used to check if a is larger than b taking into account the overflow of uint32
+    function ccwDistance32(uint32 a, uint32 b) internal pure returns (uint256) {
+        uint256 modulo = uint256(~uint32(0)) + 1;
```
TheDZhon:
```suggestion
        uint256 modulo = 1 << 32;
```

TheDZhon (approved):
LGTM, just a small change needed in comments.

TheDZhon (approved):
muy bien gracias :clap:

tamtamchik (approved):

PR #1216, opened 2025-07-02
Fix: double inOutDelta cache

### LINKED ISSUE #1198: [Bug]: incorrect handling of refSlot truncation to uint32
### Summary

The current version of the `RefSlotCache` library handles `refSlot` truncation to `uint32` incorrectly.

When updating the cache, it uses `!=` to compare `refSlot` values:
```
if (newStorage.refSlot != uint32(refSlot)) {
    ... // update cache
}
```
However, when retrieving the value for the last `refSlot`, it uses `>`:
```
if (uint32(refSlot) > _storage.refSlot) {
    return _storage.value;
} else {
    return _storage.valueOnRefSlot;
}
```
This leads to incorrect behavior when the refSlot exceeds `uint32.max`. The truncated `refSlot` wraps around and becomes less than the cached `refSlot`, causing the library to keep returning the outdated cached value — even after multiple ref slots have passed.

This incorrect behavior will persist until the cache is explicitly updated.

### Expected Behavior

The `getValueForLastRefSlot()` function should return `_storage.value` (instead of `_storage.valueOnRefSlot`) once **at least one ref slot** has passed.

### Potential Impact

This can result in incorrect calculations of `inOutDelta` and `badDebtToInternalize`.

### Steps to Reproduce

A [test case](https://github.com/lidofinance/core/pull/1197/files#diff-092cba572462a48b4e8aa40afed59dc214055ec4e77a275df5b535fe78640b4bR162-R178) is available in the fix branch that demonstrates the issue.

### Possible Solutions

Change the comparison in `getValueForLastRefSlot()` from:
`if (uint32(refSlot) > _storage.refSlot)`
to:
`if (uint32(refSlot) != _storage.refSlot)`

[Pr with fix](https://github.com/lidofinance/core/pull/1197).

### Guidelines

### LINKED ISSUE #1207: [Bug]: Usage of incorrect InOutDelta cache value
### Summary

When validating vault report parameters in `LazyOracle`, the following line is used to get the `inOutDelta` value at the time of the last `refSlot`:
```
inOutDeltaOnRefSlot = vaultHub.inOutDeltaAsOfLastRefSlot(_vault);
```

However, there's an edge case:
If a report is created by the oracle for `refSlot1`, but the lazy report for the vault arrives after `refSlot2` has started (and before the global report for `refSlot2` has been submitted), then the lazy report will pass validation, but the `inOutDeltaAsOfLastRefSlot` method will return the value as of `refSlot2`, not `refSlot1`.

### Expected Behavior

The `inOutDelta` used to validate a lazy report should always correspond to the `refSlot` the report is actually associated with.

### Potential Impact

In the edge case described, this leads to incorrect `inOutDelta` computation at the time of the report and may result in an unintended quarantine trigger.

### Steps to Reproduce

A test case demonstrating this issue will be available in the fix branch.

### Possible Solutions

Implement a two-step `inOutDelta` cache that safely stores values for the two most recent `refSlots`, ensuring correct resolution based on the actual report context.

### Guidelines

dry914:
Fix this issues:
- https://github.com/lidofinance/core/issues/1207
- https://github.com/lidofinance/core/issues/1198

---[refSlot1-----dailyReport1-----][refSlot2-----**fund**-----lazyReport1-----fund/withdraw-----dailyReport2----->

Note: This fix assumes that in the edge case between refSlot2 and the daily report for refSlot2, a report for refSlot1 can still be submitted and successfully processed.


contracts/0.8.25/vaults/LazyOracle.sol
```diff
@@ -207,7 +210,7 @@ contract LazyOracle is ILazyOracle, AccessControlEnumerableUpgradeable {
                 vaultAddress,
                 connection.vaultIndex,
                 address(vault).balance,
-                vaultHub.inOutDeltaAsOfLastRefSlot(vaultAddress),
+                record.inOutDelta.value,
```
dry914: Is it correct fix?

contracts/0.8.25/vaults/lib/RefSlotCache.sol
```diff
@@ -97,4 +99,22 @@ library RefSlotCache {
...
+    function getValueForRefSlot(
+        Int112WithRefSlotCache storage _storage,
+        uint32 _refSlot
+    ) internal view returns (int112) {
+        if (_refSlot > _storage.refSlot) {
+            return _storage.value;
+        } else if (_refSlot == _storage.refSlot) {
+            return _storage.valueOnRefSlot;
+        } else {
+            revert InOutDeltaCacheIsOverwritten();
```
dry914: ---[refSlot1-----dailyReport1-----][refSlot2-----**fund**-----lazyReport1-----fund/withdraw-----dailyReport2----->

Note: This fix assumes that in the edge case between refSlot2 and the daily report for refSlot2, a report for refSlot1 can still be submitted and successfully processed — but only if no fund occurred in between.
In other words, if the UI wants to perform a mint operation within this same window, it should first submit the report for refSlot1, and only then call fund.

contracts/0.8.25/vaults/VaultHub.sol
folkyatina, line 1: Let's add fresh report requirement for the fund method

folkyatina (commented):
Good as it is, but we add freshness requirement on fund to avoid deadlock

contracts/0.8.25/vaults/VaultHub.sol
dry914: `rebalance` and `settleObligations` functions also change `inOutDelta` cache and don't require fresh report

contracts/0.8.25/vaults/LazyOracle.sol
TheDZhon: :+1:

TheDZhon (commented):
:+1: 

Let's try to make the chained cache approach when we store:
- ref slot absolute `inOutDelta`
- diff between consequent ref slots

The idea is to `freeze` the prev ref slot cache values when new ref slot arrives

And let's shorten the freshness to (2 frames minus 1 hour)

kudzie3107 (approved):

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -758,16 +750,6 @@ contract VaultHub is PausableUntilWithRoles {
         emit BurnedSharesOnVault(_vault, _amountOfShares);
     }
 
-    /// @notice separate burn function for EOA vault owners; requires vaultHub to be approved to transfer stETH
-    /// @param _vault vault address
-    /// @param _amountOfShares amount of shares to transfer and burn
-    /// @dev msg.sender should be vault's owner
-    function transferAndBurnShares(address _vault, uint256 _amountOfShares) external {
```
dry914: can we remove it???

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -84,7 +84,7 @@ contract VaultHub is PausableUntilWithRoles {
         uint96 liabilityShares;
         // ### 3rd slot
```
dry914: 3-4 slots

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -155,7 +155,8 @@ contract VaultHub is PausableUntilWithRoles {
     uint256 public constant CONNECT_DEPOSIT = 1 ether;
 
     /// @notice The time delta for report freshness check
-    uint256 public constant REPORT_FRESHNESS_DELTA = 2 days;
+    /// @dev 2 frame minus 1 hour on mainnet (2 days - 1 hour)
```
dry914: revert to 2 frames

contracts/0.8.25/vaults/VaultHub.sol
dry914: use transferFrom in both methods

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -982,11 +964,14 @@ contract VaultHub is PausableUntilWithRoles {
             }),
             locked: uint128(CONNECT_DEPOSIT),
             liabilityShares: 0,
-            inOutDelta: RefSlotCache.Int112WithRefSlotCache({
-                value: int112(int256(vaultBalance)),
-                valueOnRefSlot: 0,
-                refSlot: 0
-            })
+            inOutDelta: [
+                RefSlotCache.Int112WithRefSlotCache({
```
dry914: move this logic to lib

contracts/0.8.25/vaults/VaultHub.sol
dry914: move this change to separate PR

TheDZhon (commented):
Let's take this approach, great work :clap: 

Though can't approve this particular implementation now and would suggest extensive test coverage :heart:

contracts/0.8.25/vaults/lib/RefSlotCache.sol
```diff
@@ -6,6 +6,8 @@ pragma solidity 0.8.25;
 
 import {IHashConsensus} from "contracts/common/interfaces/IHashConsensus.sol";
 
+uint256 constant CACHE_LENGTH = 2;
+
 library RefSlotCache {
```
TheDZhon, line 11: let's have two libs here: RefSlotCache and DoubleRefSlotCache

foundry.toml
```diff
@@ -35,3 +35,11 @@ fmt = { int_types = 'long' }
 
 # Skip contracts that are not supported by the IR compiler
 skip = ["TWVoteScript.sol"]
```
TheDZhon: can we lift this?

TheDZhon (approved):
LGTM, maybe some chores only

contracts/0.8.25/vaults/lib/RefSlotCache.sol
```diff
@@ -48,53 +83,74 @@ library RefSlotCache {
...
+        revert InOutDeltaCacheIsOverwritten();
     }
+
+    /// @dev There is a limitation on the refSlot value: it must be less than 2^32.
+    /// If it exceeds this limit, the refSlot will be truncated to 32 bits.
+    /// _activeCacheIndex may work incorrectly if one refSlot value is truncated and the other is not,
+    /// because the non-truncated value will always be greater than the truncated one,
+    /// causing incorrect activeIndex determination.
+    function _activeCacheIndex(Int112WithCache[DOUBLE_CACHE_LENGTH] memory _cache) private pure returns (uint256) {
+        return _cache[0].refSlot >= _cache[1].refSlot ? 0 : 1;
```
TheDZhon, line 153: let's add ccw check

foundry.toml
TheDZhon, line 1: probably should be in sync with the ad-hoc profile for VaultHub in the hardhat cfg

TheDZhon (approved):
:+1: A few considerations only left

contracts/0.8.25/vaults/lib/RefSlotCache.sol
dry914: I tried adding overflow handling for uint32 in the double inOutDelta cache logic — but it surfaced many edge cases, extra code, and tests.

I believe overcomplicating the code this way is not rational.

I have an alternative proposal: increase the type to uint48. With one slot per second, this would extend the lifespan from 135 years to 8,925,512 years. For the inOutDelta value, the type would decrease from int112 to int104.

contracts/0.8.25/vaults/lib/RefSlotCache.sol
```diff
@@ -48,53 +83,75 @@ library RefSlotCache {
...
+        if (_refSlot == _cache[previousCacheIndex].refSlot) {
+            return _cache[previousCacheIndex].valueOnRefSlot;
+        }
+
+        // 4. refSlot is less than prevRefSlot
+        revert InOutDeltaCacheIsOverwritten();
     }
+
+    /// @dev There is a limitation on the refSlot value: it must be less than 2^40.
+    /// If it exceeds this limit, the refSlot will be truncated to 40 bits.
```
TheDZhon:
```suggestion
    /// If it exceeds this limit, the refSlot will be truncated to 48 bits.
```

contracts/0.8.25/vaults/lib/RefSlotCache.sol
```diff
@@ -48,53 +83,75 @@ library RefSlotCache {
...
+    /// _activeCacheIndex may work incorrectly if one refSlot value is truncated and the other is not,
+    /// because the non-truncated value will always be greater than the truncated one,
+    /// causing incorrect activeIndex determination. However, 2^40 is a very large number,
```
TheDZhon:
```suggestion
    /// causing incorrect activeIndex determination. However, 2^48 is a very large number,
```

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -1617,7 +1606,7 @@ contract VaultHub is PausableUntilWithRoles {
     event MintedSharesOnVault(address indexed vault, uint256 amountOfShares, uint256 lockedAmount);
     event BurnedSharesOnVault(address indexed vault, uint256 amountOfShares);
     event VaultRebalanced(address indexed vault, uint256 sharesBurned, uint256 etherWithdrawn);
-    event VaultInOutDeltaUpdated(address indexed vault, int112 inOutDelta);
+    event VaultInOutDeltaUpdated(address indexed vault, int104 inOutDelta);
```
```suggestion
    event VaultInOutDeltaUpdated(address indexed vault, int256 inOutDelta);
```

contracts/0.8.25/vaults/VaultHub.sol
TheDZhon, line 1: Let's do the same with the record struct and get rid of ccw for timestamps there

cc: @folkyatina

TheDZhon (approved):
LGTM, yet some more work tbd :clap:

PR #1249, opened 2025-07-17
chore: update hardhat 2.24.3 -> 2.26.0

arwer13:


TheDZhon (approved):
LGTM, but would need to update the docker image we run in GH wfs

PR #1255, opened 2025-07-18
fix(test): make integrations green again

arwer13:
- update hardhat 2.26.0 → 2.26.1 (without hardhat-node docker update)
- increase memory for CI upgrade integration tests (not enough already): `  NODE_OPTIONS: --max_old_space_size=7200`
- fixes related to reaching STAKE_LIMIT 
- fixes related to the huge withdrawal queue on mainnet
- fix in core/happy-path incorrect shares transferred to treasury calculation
- skip obligation.integrations / Should settle redeptions and lido fees in correct order (gonna be fixed upon obligation rework)
- add SKIP_INTERFACES_CHECK to skip interfaces check upon compile


package.json
```diff
@@ -88,7 +90,7 @@
     "ethers": "6.14.4",
     "glob": "11.0.3",
     "globals": "15.15.0",
-    "hardhat": "2.26.0",
+    "hardhat": "2.26.1",
```
tamtamchik, line 93: ℹ️ We run on 2.26.0 in CI.

tamtamchik (approved):

package.json
arwer13: I see not problem to have hardhat-node on another "patch" version - it's just tedious to upgrade hardhat-node repo each time

PR #1277, opened 2025-07-21
fix(#1233, #1269): update pause on beacon chain deposits, remove no fee validation duplication

### LINKED ISSUE #1233: [Bug]: Pausing and resuming deposits with manual and automatic
if the deposits are paused because of the Hub pausing https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.8.25/vaults/VaultHub.sol#L1461 the owner cannot add a manual pause to it so even when the vault is healthy the resumeBeaconChainDeposits will not happen. unlike the opposite case where if the owner first paused the deposits, it continues only after the owner wanted ( + the vault is healthy). https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.8.25/vaults/VaultHub.sol#L778 this will revert. meaning if something bad happened and the owner wants to pause so the depositor will not deposit even when everything is sorted he must wait until the resume happened and only then he can pause manually.
same for resume, if the owner is the one that paused he cannot let the deposit continue when the vault is healthy. he must wait till the vault is healthy and only then he can call the manual resume. he cannot pre resume. https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.8.25/vaults/VaultHub.sol#L787 this will revert

### LINKED ISSUE #1269: [Enhancement]: Useless check
### Summary

_validateNodeOperatorFeeRate() is checked already in _setNodeOperatorFeeRate()

https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.8.25/vaults/dashboard/NodeOperatorFee.sol#L102

### Expected Behavior

do not check twice

### Potential Impact

gas lost

### Steps to Reproduce

 *

### Possible Solutions

remove the dupplicated check

### Guidelines

tamtamchik:
closes #1233, #1269

### DIFF
--- a/contracts/0.8.25/vaults/VaultHub.sol
+++ b/contracts/0.8.25/vaults/VaultHub.sol
@@ -509,6 +509,7 @@ contract VaultHub is PausableUntilWithRoles {
     /// @param _reportInOutDelta the inOutDelta of the vault
     /// @param _reportCumulativeLidoFees the cumulative Lido fees of the vault
     /// @param _reportLiabilityShares the liabilityShares of the vault
+    /// @param _reportSlashingReserve the slashingReserve of the vault
     function applyVaultReport(
         address _vault,
         uint256 _reportTimestamp,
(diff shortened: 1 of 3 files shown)


contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -780,7 +780,7 @@ contract VaultHub is PausableUntilWithRoles {
         VaultConnection storage connection = _checkConnectionAndOwner(_vault);
 
         connection.isBeaconDepositsManuallyPaused = true;
-        IStakingVault(_vault).pauseBeaconChainDeposits();
+        _pauseBeaconChainDepositsIfNotAlready(IStakingVault(_vault));
```
TheDZhon: we change semantics here :warning: 

probably shall revisit this and add an event for the flag itself

TheDZhon (commented):
:+1: 

on pause/resume — missing event, add a note about semantics
on fee check redundancy — LGTM

contracts/0.8.25/vaults/VaultHub.sol
tamtamchik: Fixed in a2219d2

failingtwice (approved):
lgtm

PR #1326, opened 2025-07-30
[VAULTS] Limit Reserve Ratio to 9,999

### LINKED ISSUE #1222: [Bug]: Division by zero at `reserveRatio = 10_000`
### Summary

`10_000` is a legal value for `reserveRatio`. However, when calculating lockedEther:
```solidity
uint256 lockedEther = Math256.max(
    (liability * TOTAL_BASIS_POINTS) / (TOTAL_BASIS_POINTS - _connection.reserveRatioBP),
    CONNECT_DEPOSIT
);
```
we divide a value by `TOTAL_BASIS_POINTS - reserveRatio`, which gives us `0` if RR is 10_000.

### Expected Behavior

Any legal reserve ratio should not throw an error.

### Potential Impact

Vault report never applies at RR 10_000

### Steps to Reproduce

1. Create a vault with reserveRatio 10_000
2. Bring report 

### Possible Solutions

1. Forbid RR 10_000
2. Check for zero before dividing

### Guidelines

failingtwice:
This PR decreases max RR value from 10,000 to 9,999 due to division by 0 error.

## Context and Problem

https://github.com/lidofinance/core/issues/1222

## Solution

Limit RR at 9,999


contracts/0.8.25/vaults/VaultHub.sol
TheDZhon, line 1: :brain: probably can remove checks in VaultHub for this purpose and rely only on the OperatorGrid internal validation bc we don't provide RR as a user input anyways.

TheDZhon (approved):
LGTM, though would suggest streamline it further :mag_right:

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -168,6 +168,8 @@ contract VaultHub is PausableUntilWithRoles {
     uint256 internal immutable PUBLIC_KEY_LENGTH = 48;
     /// @dev max value for fees in basis points - it's about 650%
     uint256 internal immutable MAX_FEE_BP = type(uint16).max;
+    /// @dev max value for reserve ratio in basis points - 9999
+    uint256 internal immutable MAX_RESERVE_RATIO_BP = 99_99;
```
folkyatina: Looks like it's not used anymore since checks are removed

folkyatina (approved):
LGTM except some dead code need to be deleted and better to have green tests

PR #1327, opened 2025-07-30
Require vault to be connected to update report data

### LINKED ISSUE #1271: [Undesired behavior]: Vault update can be called for a disconnected vault
The function `LazyOracle.updateVaultData` can be called for a disconnected vault. There is no check that the vault is connected. It is advised to have such a check to prevent vault records from being corrupted.

vp4242:
Require vault to be connected to update report data
Fixes https://github.com/lidofinance/core/issues/1271


TheDZhon (approved):
LGTM

Issue #1364, opened 2025-08-12
[Bug]: Incorrect Comment Regarding Validator Count Comparison


tamtamchik:
**Description**
This issue has been identified within the `_getInternalEther` function of the Lido contract.

The inline comment above the assertion incorrectly states that "clValidators can never be less than deposited ones." However, the assertion itself correctly checks that `depositedValidators >= clValidators`, which reflects the actual expected behavior: **deposited validators may exceed the number of Consensus Layer (CL)** validators due to validators being in a transient state (submitted to the Deposit contract but not yet activated on the CL).

**Recommendation**
We recommend updating the comment to correctly reflect the validator relationship. Suggested revision:
`// clValidators can never exceed depositedValidators.`

PR #1371, opened 2025-08-13
docs(lido): correct CL validators vs deposited validators comment in _getTransientBalance

MozirDmitriy:
Update comment to state “clValidators can never exceed depositedValidators,” aligning docs with the existing assertion assert(depositedValidators >= clValidators).
Rationale: deposited validators can be in a transient state (submitted to Deposit contract but not yet active on CL), so deposited can be greater than CL validators.
refs #1364


tamtamchik:

Issue #1267, opened 2025-07-20
[Bug]: bad debt existence when possible to fix with the predepositgerantee


tomerganor:
### Summary

if the funds were predeposited to the validator but it was frontruned the vault has baddebt but we can (and should) fix it with the predepositgerantee contract by calling https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.8.25/vaults/VaultHub.sol#L944
should not be possible to call https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.8.25/vaults/VaultHub.sol#L607 when this is the reason and there is funds in the PDG to fix this case.

### Expected Behavior

-

### Potential Impact

-

### Steps to Reproduce

-

### Guidelines

- [x] I agree to follow [Code of Conduct](https://github.com/lidofinance/core/blob/master/CODE_OF_CONDUCT.md).
- [x] I have read [Lido Contribution Guide](https://github.com/lidofinance/core/blob/master/CONTRIBUTING.md).

tamtamchik:
Seems we can acknowledge the issue, given that the compensation became permissionless after the initial report.

folkyatina:
Partially fixed by #1336

PR #1383, opened 2025-08-19
[VAULTS] Track unresolved validators in PDG to prevent pessimistic baddebt internalization

failingtwice:
Context
https://github.com/lidofinance/core/issues/1267


TheDZhon (commented):
Let's park it for a while — seems we can acknowledge the issue given the compensate became permissionless after initial report. :thinking:

PR #1387, opened 2025-08-21
[VAULTS] sync tier logic

### LINKED ISSUE #1361: [Bug]: Misleading Comment in the OperatorGrid.changeTier() Function
**Description**
This issue has been identified within the `changeTier` function of the `OperatorGrid` contract.

The function reverts if the vault has not yet been connected to `VaultHub`. The function checks whether the vault is connected using `vaultHub.isVaultConnected(_vault)` and then [calls](https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.8.25/vaults/OperatorGrid.sol#L454) `vaultHub.updateConnection()` regardless of the connection status. This will cause a revert if a vault is not yet connected to `VaultHub`, which is intentional.

The issue is classified as **Low** severity because there is a misleading comment before the call to `vaultHub.updateConnection`, which states that this call should be skipped if the vault is not connected.

**Recommendation**
We recommend modifying the comment to accurately reflect the implemented logic and to include a note about the potential revert if the vault is not connected.

### LINKED ISSUE #1380: [Bug]: It is not allowed to request share limit change without changing tier
Currently, it is not allowed 
https://github.com/lidofinance/core/blob/09efed8c6ed9df4afbff93c61a71c6854b9b9e79/contracts/0.8.25/vaults/OperatorGrid.sol#L427

### LINKED ISSUE #1388: [Feature Request]: Add function to update shareLimit of a vault
### Summary

It is proposed to add functionality that allows updating the `shareLimit` value of a vault by agreement between the vault owner and the node operator, without changing the tier or other parameters of the vault:
- new `shareLimit` must not exceed the maximum allowed for the vault’s tier/group
- if `shareLimit` is increased, the operator’s confirmation is required
- if `shareLimit` is decreased, the operator’s confirmation is not required

### Problem

.

### Implementation

.

### Guidelines

### LINKED ISSUE #1391: [Bug]: Missing fresh report requirements enable operations on stale vault data
#### Description

Multiple functions in `VaultHub` and `OperatorGrid` lack fresh Oracle report validation, allowing operations based on potentially outdated vault state. This creates several security vulnerabilities:

**Disconnection soft-locking:** 
`disconnect` and `voluntaryDisconnect` do not require a fresh oracle report. Disconnection is initiated via `_initiateDisconnection` using potentially stale `VaultRecord` data; `pendingDisconnect` is then set. Fees continue accruing until the next report. When a fresh report is applied during `pendingDisconnect`, `_settleObligations(..., NO_UNSETTLED_ALLOWED)` must fully settle fees. If the vault’s current balance on EL cannot cover accrued fees, report application reverts, leaving `pendingDisconnect` true, even though a vault might have sufficient value on CL. In this state, key actions are blocked by `_checkConnection`/`_checkConnectionAndOwner` (including `forceValidatorExit` and `triggerValidatorWithdrawals`), effectively soft‑locking the vault until external funding occurs.

**Off-chain Fee calculation:**
The `changeTier` function in `OperatorGrid` allows vaults to change their fee parameters (`infraFeeBP`, `liquidityFeeBP`, `reservationFeeBP`) without requiring a fresh report. This creates a potential attack vector where vaults can manipulate their fee obligations by changing tiers before reporting their cumulative value. The fee parameters are stored in storage but never used in fee calculations - the actual fees are reported externally through the `LazyOracle` system. A malicious vault owner could stay in the default tier with higher fees, accumulate significant value without reporting, then change to a higher tier with lower fees before reporting, resulting in retroactively lower fee obligations (this depends on the off-chain calculation, which is out of scope). The `_cumulativeLidoFees` value in case of the missing reports may be lower than it was before due to the change of the Vault Tier, as it can be unclear for the backend what fees to use without applied oracle reports as checkpoints. We recommend requiring fresh reports for tier changes by adding a check for `_isReportFresh(_vault)` in the `changeTier`/`updateConnection` function. 

**Stale health checks:**
The `triggerValidatorWithdrawals` and `forceValidatorExit` functions do not invoke the `_requireFreshReport` check to ensure the Oracle report is recent. Without freshness validation, a vault owner could execute partial or full withdrawals from the consensus layer using stale or outdated reports—potentially on an unhealthy vault—or call `forceValidatorExit` after the vault has become healthy. As a result, an account with `VALIDATOR_EXIT_ROLE` could trigger full validator exits unintentionally. 

Additionally, other functions also lack fresh report requirements, allowing operations based on stale data:
- `pauseBeaconChainDeposits(address _vault)` - Can pause deposits without a fresh report.
- `resumeBeaconChainDeposits(address _vault)` - Can resume deposits without a fresh report.
- `requestValidatorExit(address _vault, bytes calldata _pubkeys)` - Can request exits without fresh report.

#### Recommendation
We recommend implementing report freshness validation across all state-changing functions that rely on vault health metrics. Additionally, modify the disconnection logic to gracefully handle fee settlement failures by aborting disconnection rather than reverting, preventing vault soft-locking scenarios.

### Expected Behavior

The disconnection logic should abort disconnection when fee settlement fails due to insufficient EL balance, rather than reverting and leaving the vault in a soft-locked state.  Report freshness validation should be validated across all state-changing functions that rely on vault health metrics.

### Potential Impact

The vault gets soft locked, and can be disconnected only after direct funding.

### Steps to Reproduce

A vault with a 2048 ETH validator has minted stETH while maintaining health checks, with 1 ETH locked on EL and 2047 ETH on CL. The vault stops applying reports for a month while remaining healthy (no slashings, redemptions). During this period, it accrues significant Lido fees. When attempting to apply a report during disconnection, `_settleObligations(..., NO_UNSETTLED_ALLOWED)` requires settling more fees in ETH than a vault has available on EL. The function reverts, leaving `pendingDisconnect` true and blocking all vault operations until external funding occurs. The disconnection logic should abort disconnection when fee settlement fails due to insufficient EL balance, rather than reverting and leaving the vault in a soft-locked state.

### Possible Solutions

_No response_

### Guidelines

dry914:
Fix:
- https://github.com/lidofinance/core/issues/1380
- https://github.com/lidofinance/core/issues/1388
- https://github.com/lidofinance/core/issues/1361
- changeTier part of https://github.com/lidofinance/core/issues/1391


contracts/0.8.25/vaults/OperatorGrid.sol
```diff
@@ -472,6 +472,74 @@ contract OperatorGrid is AccessControlEnumerableUpgradeable, Confirmable2Address
         return true;
     }
 
+    /// @notice Sync vault params to the current tier params
+    /// @param _vault address of the vault
+    function syncVaultWithTier(address _vault) external {
```
TheDZhon: should be confirmable by a node operator :warning:

TheDZhon: can modify a bit `changeTier` instead to tackle with this issue

contracts/0.8.25/vaults/OperatorGrid.sol
```diff
@@ -425,37 +432,40 @@ contract OperatorGrid is AccessControlEnumerableUpgradeable, Confirmable2Address
 
         address nodeOperator = IStakingVault(_vault).nodeOperator();
 
-        uint256 vaultTierId = $.vaultTier[_vault];
-        if (vaultTierId == _requestedTierId) revert TierAlreadySet();
-
         Tier storage requestedTier = $.tiers[_requestedTierId];
-        if (nodeOperator != requestedTier.operator) revert TierNotInOperatorGroup();
+        // In sync mode, skip operator check since we're not changing tiers
+        if (!isSyncMode && nodeOperator != requestedTier.operator ) revert TierNotInOperatorGroup();
```
TheDZhon:
```suggestion
        if (!isSyncMode && nodeOperator != requestedTier.operator) revert TierNotInOperatorGroup();
```

contracts/0.8.25/vaults/OperatorGrid.sol
```diff
@@ -414,7 +416,12 @@ contract OperatorGrid is AccessControlEnumerableUpgradeable, Confirmable2Address
 
         ERC7201Storage storage $ = _getStorage();
         if (_requestedTierId >= $.tiers.length) revert TierNotExists();
-        if (_requestedTierId == DEFAULT_TIER_ID) revert CannotChangeToDefaultTier();
+
```
TheDZhon: let's introduce a dedicate fn to deal with sync mode (when we actually don't change the tier, but rather sync up with the up-to-date tier params)

for this we would also need new events and simplified set of checks (e.g., the same node operator assumption is intrinsic)

contracts/0.8.25/vaults/OperatorGrid.sol
```diff
@@ -472,6 +482,49 @@ contract OperatorGrid is AccessControlEnumerableUpgradeable, Confirmable2Address
...
+        if (_requestedShareLimit > tierShareLimit) revert RequestedShareLimitTooHigh(_requestedShareLimit, tierShareLimit);
+
+        VaultHub.VaultConnection memory vaultConnection = vaultHub.vaultConnection(_vault);
+        address vaultOwner = vaultConnection.owner;
+
+        if (_requestedShareLimit == vaultConnection.shareLimit) revert ShareLimitAlreadySet();
+
+        // nodeOperator confirmation is only needed if vault is not in the default tier and if the requested shareLimit is more than the current one
+        if (vaultTierId == DEFAULT_TIER_ID || _requestedShareLimit <= vaultConnection.shareLimit) {
+            if (msg.sender != vaultOwner) revert NotAuthorized("updateVaultShareLimit", msg.sender);
```
TheDZhon:
```suggestion
            if (msg.sender != vaultOwner) return false;
```

contracts/0.8.25/vaults/OperatorGrid.sol
```diff
@@ -472,6 +482,49 @@ contract OperatorGrid is AccessControlEnumerableUpgradeable, Confirmable2Address
...
+        } else {
+            address nodeOperator = IStakingVault(_vault).nodeOperator();
+            // store the caller's confirmation; only proceed if the required number of confirmations is met.
+            if (!_collectAndCheckConfirmations(msg.data, vaultOwner, nodeOperator)) return false;
+        }
+
+        vaultHub.updateConnection(
```
TheDZhon, line 544: under updateConnection we probably want to add a no to subscribe on the vaults' updates by a no operator validating them

contracts/0.8.25/vaults/OperatorGrid.sol
```diff
@@ -472,6 +482,49 @@ contract OperatorGrid is AccessControlEnumerableUpgradeable, Confirmable2Address
...
+
+        if (_requestedShareLimit > tierShareLimit) revert RequestedShareLimitTooHigh(_requestedShareLimit, tierShareLimit);
+
+        VaultHub.VaultConnection memory vaultConnection = vaultHub.vaultConnection(_vault);
+        address vaultOwner = vaultConnection.owner;
+
+        if (_requestedShareLimit == vaultConnection.shareLimit) revert ShareLimitAlreadySet();
+
+        // nodeOperator confirmation is only needed if vault is not in the default tier and if the requested shareLimit is more than the current one
+        if (vaultTierId == DEFAULT_TIER_ID || _requestedShareLimit <= vaultConnection.shareLimit) {
```
TheDZhon, line 537: prob we should document this particular shortcut to make it visible for client's code

TheDZhon (commented):
:clap: 

Let's iterate one more time and make it even better than it's now :+1:

contracts/0.8.25/vaults/dashboard/Dashboard.sol
```diff
@@ -557,15 +557,37 @@ contract Dashboard is NodeOperatorFee {
     }
 
     /**
-     * @notice Requests a change of tier on the OperatorGrid.
+     * @notice Requests a change of tier or sync of tier (if same tier) on the OperatorGrid.
```
folkyatina: update comment back

contracts/0.8.25/vaults/dashboard/Permissions.sol
```diff
@@ -365,6 +365,23 @@ abstract contract Permissions is AccessControlConfirmable {
...
+    function _syncTier() internal onlyRoleMemberOrAdmin(CHANGE_TIER_ROLE) returns (bool) {
+        return _operatorGrid().syncTier(address(_stakingVault()));
+    }
+
+    /**
+     * @dev Checks the CHANGE_TIER_ROLE and updates the share limit on the OperatorGrid.
+     * @param _requestedShareLimit The requested share limit.
+     * @return bool Whether the share limit update was confirmed.
+     */
+    function _updateVaultShareLimit(uint256 _requestedShareLimit) internal onlyRoleMemberOrAdmin(CHANGE_TIER_ROLE) returns (bool) {
```
folkyatina: 🤔 maybe rename the role because of this method

contracts/0.8.25/vaults/OperatorGrid.sol
```diff
@@ -481,6 +482,81 @@ contract OperatorGrid is AccessControlEnumerableUpgradeable, Confirmable2Address
...
+            _vault,
+            vaultConnection.shareLimit,
+            tier.reserveRatioBP,
+            tier.forcedRebalanceThresholdBP,
+            tier.infraFeeBP,
+            tier.liquidityFeeBP,
+            tier.reservationFeeBP
+        );
+
+        emit TierSynced(_vault);
```
folkyatina: 🤔 maybe not needed, because storage of OperatorGrid hasn't changed and we have event in `updateConnection`

contracts/0.8.25/vaults/OperatorGrid.sol
```diff
@@ -481,6 +482,81 @@ contract OperatorGrid is AccessControlEnumerableUpgradeable, Confirmable2Address
...
+    }
+
+    /// @notice Update vault share limit
+    /// @param _vault address of the vault
+    /// @param _requestedShareLimit share limit to set
+    /// @return bool Whether the update was confirmed.
+    function updateVaultShareLimit(address _vault, uint256 _requestedShareLimit) external returns (bool) {
+        (VaultHub vaultHub, VaultHub.VaultConnection memory vaultConnection, address vaultOwner, address nodeOperator) = _getVaultContextForConnectedVault(_vault);
+
+        ERC7201Storage storage $ = _getStorage();
```
folkyatina: Maybe put it under `_getVaultContextForConnectedVault` as well?

contracts/0.8.25/vaults/OperatorGrid.sol
```diff
@@ -372,7 +372,8 @@ contract OperatorGrid is AccessControlEnumerableUpgradeable, Confirmable2Address
     /// @param _requestedTierId id of the tier
     /// @param _requestedShareLimit share limit to set
     /// @return bool Whether the tier change was confirmed.
```
folkyatina:
```suggestion
    /// @return bool Whether the tier change was executed.
``` here and in other methods
Maybe worth explaining confirmations in more IQ-50 fashion.

File: contracts/0.8.25/vaults/OperatorGrid.sol [hunk 12]
```diff
@@ -481,6 +482,81 @@ contract OperatorGrid is AccessControlEnumerableUpgradeable, Confirmable2Address
...
+            tier.liquidityFeeBP,
+            tier.reservationFeeBP
+        );
+
+        emit TierSynced(_vault);
+
+        return true;
+    }
+
+    /// @notice Update vault share limit
```
folkyatina, line 522:
```suggestion
    /// @notice Update vault share limit for a connected vault
```

contracts/0.8.25/vaults/OperatorGrid.sol
```diff
@@ -481,6 +482,81 @@ contract OperatorGrid is AccessControlEnumerableUpgradeable, Confirmable2Address
         return true;
     }
 
+    /// @notice Syncs vault tier with current tier params
+    /// @param _vault address of the vault
+    /// @return bool Whether the sync was confirmed.
+    /// @dev Requires vault to be connected to VaultHub to finalize tier sync.
```
folkyatina:
```suggestion
    /// @dev Requires vault to be connected to VaultHub
```

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -1499,6 +1502,7 @@ contract VaultHub is PausableUntilWithRoles {
```
folkyatina: Looks like it's worth to extract `updateVaultFees` to OperatorGrid to group all business-related logic together.

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -1623,12 +1627,14 @@ contract VaultHub is PausableUntilWithRoles {
 
     event VaultConnectionUpdated(
         address indexed vault,
+        address indexed nodeOperator,
         uint256 shareLimit,
         uint256 reserveRatioBP,
         uint256 forcedRebalanceThresholdBP
     );
     event VaultFeesUpdated(
```
folkyatina, line 1492: Let's merge events as well

folkyatina (commented):
Some nitpicks and an idea to move updating fees to OperatorGrid as well

contracts/0.8.25/vaults/OperatorGrid.sol
```diff
@@ -496,6 +569,40 @@ contract OperatorGrid is AccessControlEnumerableUpgradeable, Confirmable2Address
...
+        uint256 _liquidityFeeBP,
+        uint256 _reservationFeeBP
+    ) external onlyRole(REGISTRY_ROLE) {
+        if (_vault == address(0)) revert ZeroArgument("_vault");
+
+        _requireLessThanBP(_infraFeeBP, MAX_FEE_BP);
+        _requireLessThanBP(_liquidityFeeBP, MAX_FEE_BP);
+        _requireLessThanBP(_reservationFeeBP, MAX_FEE_BP);
+
+        VaultHub vaultHub = _vaultHub();
```
TheDZhon, line 589: shall we use _getVaultContextForConnectedVault as above?

TheDZhon (approved):
LGTM, pls don't forget to update ET

contracts/0.8.25/vaults/OperatorGrid.sol
folkyatina: Looks redundant for this method, tbh

contracts/0.8.25/vaults/OperatorGrid.sol
```diff
@@ -696,6 +800,29 @@ contract OperatorGrid is AccessControlEnumerableUpgradeable, Confirmable2Address
...
+        if (!vaultHub.isVaultConnected(_vault)) revert VaultNotConnected();
+
+        vaultConnection = vaultHub.vaultConnection(_vault);
+        vaultOwner = vaultConnection.owner;
+        nodeOperator = IStakingVault(_vault).nodeOperator();
+
+        vaultTierId = _getStorage().vaultTier[_vault];
+    }
+
+    function _requireLessThanBP(uint256 _valueBP, uint256 _maxValueBP) internal pure {
```
TheDZhon:
```suggestion
    function _requireLessOrEqThanBP(uint256 _valueBP, uint256 _maxValueBP) internal pure {
```

TheDZhon: or `revertIfGreater`

TheDZhon (approved):
LGTM, a few comments only

PR #1400, opened 2025-08-26
[VAULTS] Patch: `PinnedBeaconProxy` can hide malicious implementation via constructor

### LINKED ISSUE #1390: [Bug]: `PinnedBeaconProxy` can hide malicious implementation via constructor
### Summary

The staking vaults are designed to be connected to the `VaultHub`. To connect, a vault must meet the following requirements:

1. The vault’s contract code must match `PinnedBeaconProxy` and include Lido’s official beacon address. `PinnedBeaconProxy` is a standard beacon proxy with one extra feature: it can `ossify()`, which allows the owner to fix the current implementation.

2. The vault must not be `isOssified()`. The `StakingVault` implementation is the one that reports `isOssified()`

If the implementation is pinned, it is stored in the proxy’s `PINNED_BEACON_STORAGE_SLOT`. Normally this slot is empty or holds one of the official `StakingVault` implementations. If an attacker manages to store a malicious implementation here, they gain full control of the vault. This would allow them to withdraw all collateral while liabilities remain, effectively stealing `stETH` from the Lido core protocol.

The attack works because the proxy’s `codehash` is only based on the runtime code, not the constructor. An attacker can deploy a proxy with the same whitelisted `codehash` but modify the constructor to preset `PINNED_BEACON_STORAGE_SLOT` with a malicious implementation. That implementation can simply return `false` in `isOssified()`, allowing it to pass the `VaultHub` checks. Once connected, the attacker could mint a large amount of `stETH` and then drain all collateral from the malicious vault.

### Expected Behavior

-

### Potential Impact

-

### Steps to Reproduce

-

### Possible Solutions

_No response_

### Guidelines

failingtwice:
Context: https://github.com/lidofinance/core/issues/1390

### DIFF
--- a/contracts/0.8.25/vaults/PinnedBeaconProxy.sol
+++ b/contracts/0.8.25/vaults/PinnedBeaconProxy.sol
@@ -4,7 +4,6 @@
 // See contracts/COMPILERS.md
 pragma solidity 0.8.25;
 
-import {IBeacon} from "@openzeppelin/contracts-v5.2/proxy/beacon/IBeacon.sol";
 import {BeaconProxy} from "@openzeppelin/contracts-v5.2/proxy/beacon/BeaconProxy.sol";
 import {PinnedBeaconUtils} from "./lib/PinnedBeaconUtils.sol";
 
@@ -25,19 +24,20 @@ import {PinnedBeaconUtils} from "./lib/PinnedBeaconUtils.sol";
 contract PinnedBeaconProxy is BeaconProxy {
     constructor(address beacon, bytes memory data) BeaconProxy(beacon, data) payable {}
 
+    function isOssified() external view returns (bool) {
+        return PinnedBeaconUtils.getPinnedImplementation() != address(0);
+    }
+
     function _implementation() internal view virtual override returns (address) {
         address pinnedImpl = PinnedBeaconUtils.getPinnedImplementation();
         if (pinnedImpl != address(0)) {
             return pinnedImpl;
         }
-        return IBeacon(_getBeacon()).implementation();
+        
+        return super._implementation();
     }
 
     function implementation() external view returns (address) {
         return _implementation();
-    }
-
-    function isOssified() external view returns (bool) {
-        return PinnedBeaconUtils.ossified();
     }
 }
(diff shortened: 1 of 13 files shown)


contracts/0.8.25/vaults/PinnedBeaconProxy.sol
```diff
@@ -36,4 +36,8 @@ contract PinnedBeaconProxy is BeaconProxy {
     function implementation() external view returns (address) {
         return _implementation();
     }
+
+    function isOssified() external view returns (bool) {
+        return PinnedBeaconUtils.ossified();
```
TheDZhon: may we use the same approach as in `_implementation`

contracts/0.8.25/vaults/PinnedBeaconProxy.sol
```diff
@@ -36,4 +36,8 @@ contract PinnedBeaconProxy is BeaconProxy {
     function implementation() external view returns (address) {
         return _implementation();
     }
+
```
TheDZhon: let's change the behavior of `ossify` as well ⇒ to not revert if someone spoofed the storage, but unconditionally override

TheDZhon: probably can add ossify to the scope of proxy?

TheDZhon: to sum up, two classes of attacks:
- the issue described (ability to mint stETH against 'unknown' impl)
- the issue about 'honeypot' for new deployments: can't even connect to vaulthub if impl storage already spoofed
- DAO can't dismantle 'ossify' from new impls

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -358,7 +359,7 @@ contract VaultHub is PausableUntilWithRoles {
 
         IStakingVault vault_ = IStakingVault(_vault);
         if (vault_.pendingOwner() != address(this)) revert VaultHubNotPendingOwner(_vault);
-        if (vault_.isOssified()) revert VaultOssified(_vault);
+        if (IPinnedBeaconProxy(address(vault_)).isOssified()) revert VaultOssified(_vault);
```
TheDZhon, line 362: :+1:

TheDZhon (commented):
:heart: :love_hotel: 

Prob some refactoring only and attack vectors analysis

contracts/0.8.25/vaults/PinnedBeaconProxy.sol
folkyatina: Looks like the 3rd one requires 'vault ownership' to be checked in the proxy, which is too heavy of a change to get in the last minute. See #1408 for an example.

folkyatina (commented):
LGTM, but need a bit of love:
- remove isOssified from StakingVault
- unify approaches to storage access in PinnedBeaconProxy

PR #1287, opened 2025-07-23
[VAULTS]: Audit fixes

folkyatina:
This PR contains all the fresh fixes for the issues found by vault audits as well as fresh fixes for triggerable withdrawals (e.g fix for https://research.lido.fi/t/security-disclosure-post-mortem-csverifier-weak-validation-of-the-historical-block-gindex-user-funds-remain-safe/10466)

Lido: 
- storage migrated in new slots to prevent misleading offchain tooling
- events are sorted out
- rebalancing use shares instead of ether

Accounting: 
- return `simulatedShareRate` to the report to fix WQ handling
- reorder calls on applying report to prevent invoking callbacks mid-state

LazyOracle: 
- Quarantine was fixed and rewritten for more clarity
- More sanity checks

StakingVault:
- fix ossification check
- prevent any other similar attack by limiting vaults by factory, not codehash

PDG:
- compensation is now permissionless and directly to the vault

VaultHub:
- Redemptions are now in shares and mostly redesigned
- Obligations include forced rebalances, redemptions and unpaid fees
- Bad debt socializing and internalizing improved
- Minimal reserve is introduced
- Safer disconnection
- Locked is now calculated dynamically from maxLiabilityShares and minimalReserve
- maxLiabilityShares is used for safer unlocking
- inOutDelta is cached for last two refslots now

OperatorGrid:
- added syncTier method to update vault params from tier
- added jail mechanics
- moved `updateShareLimit` and `updateFees` from VaultHub here

VaultFactory:
- having a registry of deployed vaults
- fix setting wrong nodeOperatorFee recipient

PR #1401, opened 2025-08-26
Add Dual Governance to protocol scratch deploy

sandstone-ag:
Add Dual Governance to protocol scratch deploy

## Context

After the deployment of Dual Governance on Lido mainnet we need to add the DG deployment to protocol scratch deploy routine.

## Problem

DAO-654
After the deployment of Dual Governance on Lido mainnet we need to add the DG deployment to protocol scratch deploy routine.

## Solution

Added DG to scratch deployment


scripts/scratch/steps/0160-deploy-dual-governance.ts
```diff
@@ -0,0 +1,376 @@
...
+            .max_sealable_withdrawal_blockers_count,
+        max_tiebreaker_activation_timeout:
+          networkState[Sk.dualGovernanceConfig].dual_governance.sanity_check_params.max_tiebreaker_activation_timeout,
+        min_tiebreaker_activation_timeout:
+          networkState[Sk.dualGovernanceConfig].dual_governance.sanity_check_params.min_tiebreaker_activation_timeout,
+        min_withdrawals_batch_size:
+          networkState[Sk.dualGovernanceConfig].dual_governance.sanity_check_params.min_withdrawals_batch_size,
+      },
+    },
+    dual_governance_config_provider: {
```
arwer13: Just
```suggestion
    dual_governance_config_provider: networkState[Sk.dualGovernanceConfig].dual_governance_config_provider
```
here and similar places, why not?

scripts/scratch/steps/0160-deploy-dual-governance.ts
```diff
@@ -0,0 +1,376 @@
...
+  try {
+    await runCommand("npm run test:regressions", DG_INSTALL_DIR);
+  } catch (error) {
+    // TODO: some of regression tests don't work at the moment, need to fix it.
+    log.error("DG regression tests run failed");
+    log(`${error}`);
+  }
+}
+
+async function runCommand(command: string, workingDirectory: string) {
```
arwer13: runCommand duplicates here and in dg-installation.ts . May move to `lib/subprocess.ts` or some other lib file

arwer13 (changes requested):
Thanks for the work!

I see these follow ups:
1. Rebase on up-to-date vaults branch https://github.com/lidofinance/core/pull/1287

2. Add integration which verifies dual governance upgrade pass (for any dummy one). Boilerplates for the voting script (contract) is in the vaults branch as well. At the moment it won't work because proper roles distribution (in LidoTemplate.sol) is missing.

3. Make sure no-DG setup is also operational. It is at the moment, but after the roles tinkering there gonna be two Aragon roles configuration. Actually, I'm not 100% sure we  optional DG installation is worth the effort, but likely it is. To simplify we can split LidoTemplate into LidoTemplateBase, LidoTemplate and LidoTemplateWithDG - it seems not much work. But for the integration tests and for the utility contracts/scripts which do an upgrade there must be two setups to support in this case. Wdyt @tamtamchik ?

tamtamchik:
> Make sure no-DG setup is also operational. It is at the moment, but after the roles tinkering there gonna be two Aragon roles configuration. Actually, I'm not 100% sure we  optional DG installation is worth the effort, but likely it is. To simplify we can split LidoTemplate into LidoTemplateBase, LidoTemplate and LidoTemplateWithDG - it seems not much work. But for the integration tests and for the utility contracts/scripts which do an upgrade there must be two setups to support in this case. Wdyt @tamtamchik ?

Agree, we need no-DG setup also for some devnets to simplify operational work when DG not related to the scope of changes.

PR #874, opened 2024-11-17
[EPIC] stVaults

tamtamchik:
## stVaults

New way of isolated staking, through separate vaults, with optional stETH liquidity

- Support minting externally-backed stETH in the Lido contract 
- Extract protocol accounting to separate contract (two-step, better precision)
- StakingVault + VaultHub
- Dashboard 
- VaultFactory
- OperatorGrid
- PredepositGuarantee
- LazyOracle

PR #1403, opened 2025-08-27
[VAULTS] various Lido fixes and polish

### LINKED ISSUE #1229: [Bug]: wrong order of update to external shares
### Summary

the pre and postRebaseTokenAmount are not correct here https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.4.24/StETH.sol#L550 and here https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.4.24/StETH.sol#L557
in the case where you burn external shares, since we first https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.4.24/Lido.sol#L701 set the external shares to decrease and this increase the amount of internal shares in a wrong way causing the calculation to be wrong.
example:
total share 2000
external shares 1500
internal shares 500
total assets 3000
ratio 1:6

if we burn 1000 external shares we first decrease the external shares to 500 
causing the internal shares to look like 1500 this will change the ratio from 1:6 to 1:2 and the emits will be wrong. 
now the pre balance is 2000 and post is 6000.
this pattern is scary.

### Expected Behavior

-

### Potential Impact

-

### Steps to Reproduce

burn external shares

### Possible Solutions

_No response_

### Guidelines

### LINKED ISSUE #1230: [Bug]: possible to transfer StETH to StETH contract
### Summary

it is possible to transfer shares to the StETH contract which is not allowed here https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.4.24/StETH.sol#L497 or here https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.4.24/Lido.sol#L640 (accounting will not mint to StETH) but is allowed (in the new code) here https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.4.24/Lido.sol#L681

### Expected Behavior

not allow this transfer

### Potential Impact

-

### Steps to Reproduce

external mint to the StETH address

### Possible Solutions

block this by checking in the external mint

### Guidelines

### LINKED ISSUE #1363: [Bug]: Inconsistent Accessor Usage for Packed Storage Slot
**Description**
Within the `_getLidoLocator()` function of the Lido contract, the contract retrieves the address stored in a packed storage slot using `getStorageAddress()`. However, the paired setter method `_setLidoLocator()` uses `setLowUint160`, suggesting that the lower 160 bits of the slot should be interpreted as an address.

This inconsistency may introduce confusion for maintainers and creates room for bugs in future upgrades if assumptions about the storage format diverge.

**Recommendation**
To ensure consistency and clarity, we recommend replacing:
```
LOCATOR_AND_MAX_EXTERNAL_RATIO_POSITION.getStorageAddress()
```
with:
```
LOCATOR_AND_MAX_EXTERNAL_RATIO_POSITION.getLowUint160()
```
This will align the getter and setter semantics, reduce confusion, and make the storage layout more predictable.

### LINKED ISSUE #1364: [Bug]: Incorrect Comment Regarding Validator Count Comparison

### LINKED ISSUE #1374: [Bug]: Staking Can Be Resumed While Protocol Is Stopped, Breaking Core Invariant
**Description**
The resumeStaking function in Lido allows resuming staking (accepting new deposits) even when the protocol is globally stopped (i.e., after calling stop). This breaks a core protocol invariant, explicitly stated in the code, that protocol pause must always imply staking pause. The _submit function, which handles deposits, relies on this invariant and does not check the protocol's stopped state, only the staking pause flag. As a result, an authorized actor with the STAKING_CONTROL_ROLE can call resumeStaking after stop, enabling deposits and minting of stETH when the protocol is not ready to process them.

The issue is classified as Medium severity because it exposes the protocol to operational mistakes and breaks the intended safety model.

**Recommendation**
We recommend adding an explicit check in resumeStaking to ensure the protocol is not globally stopped (e.g., require(!isStopped(), "PROTOCOL_STOPPED")). This will enforce the invariant and prevent staking from being enabled while the protocol is paused.

### LINKED ISSUE #1398: [Bug]: changes of the storage layout may brake integrations
`Lido.sol` storage optimizations may break external integrations that work directly with store

folkyatina:
Fixes #1229 
Fixes #1230 
Fixes #1363 
Fixes #1364
Fixes #1374 
Fixes #1398

Consistent and backward compatible events for minting and burning shares and other fixes for this review from @avsetsin https://github.com/lidofinance/core/pull/874#pullrequestreview-3115831184

### DIFF
--- a/contracts/0.4.24/Lido.sol
+++ b/contracts/0.4.24/Lido.sol
@@ -192,7 +192,7 @@ contract Lido is Versioned, StETHPermit, AragonApp {
     event ExternalSharesMinted(address indexed receiver, uint256 amountOfShares, uint256 amountOfStETH);
 
     // External shares burned for account
-    event ExternalSharesBurned(address indexed account, uint256 amountOfShares, uint256 stethAmount);
+    event ExternalSharesBurnt(uint256 amountOfShares);
 
     // Maximum ratio of external shares to total shares in basis points set
     event MaxExternalRatioBPSet(uint256 maxExternalRatioBP);
@@ -212,7 +212,7 @@ contract Lido is Versioned, StETHPermit, AragonApp {
      * @dev NB: by default, staking and the whole Lido pool are in paused state
      * @dev The contract's balance must be non-zero to mint initial shares of stETH
      */
-    function initialize(address _lidoLocator, address _eip712StETH) public payable onlyInit {
+    function initialize(address _lidoLocator, address _eip712StETH) external payable onlyInit {
         _bootstrapInitialHolder(); // stone in the elevator
 
         _setLidoLocator(_lidoLocator);
@@ -636,13 +636,11 @@ contract Lido is Versioned, StETHPermit, AragonApp {
      * @param _amountOfShares amount of shares to mint
      * @dev can be called only by accounting
      */
-    function mintShares(address _recipient, uint256 _amountOfShares) public {
+    function mintShares(address _recipient, uint256 _amountOfShares) external {
         _auth(_getLidoLocator().accounting());
         _whenNotStopped();
 
         _mintShares(_recipient, _amountOfShares);
-        // emit event after minting shares because we are always having the net new ether under the hood
-        // for vaults we have new locked ether and for fees we have a part of rewards
         _emitTransferAfterMintingShares(_recipient, _amountOfShares);
     }
 
@@ -651,14 +649,17 @@ contract Lido is Versioned, StETHPermit, AragonApp {
      * @param _amountOfShares amount of shares to burn
      * @dev can be called only by burner
      */
-    function burnShares(uint256 _amountOfShares) public {
+    function burnShares(uint256 _amountOfShares) external {
         _auth(_getLidoLocator().burner());
         _whenNotStopped();
-        _burnShares(msg.sender, _amountOfShares);
 
-        // historically there is no events for this kind of burning
-        // TODO: should burn events be emitted here?
-        // maybe TransferShare for cover burn and all events for withdrawal burn
+        uint256 preRebaseTokenAmount = getPooledEthByShares(_amountOfShares);
+        _burnShares(msg.sender, _amountOfShares);
+        uint256 postRebaseTokenAmount = getPooledEthByShares(_amountOfShares);
+
+        // Historically, Lido contract does not emit Transfer to zero address events
+        // for burning but emits SharesBurnt instead, so it's kept here for compatibility
+        _emitSharesBurnt(msg.sender, preRebaseTokenAmount, postRebaseTokenAmount, _amountOfShares);
     }
 
     /**
@@ -679,8 +680,6 @@ contract Lido is Versioned, StETHPermit, AragonApp {
         _setExternalShares(_getExternalShares() + _amountOfShares);
 
         _mintShares(_recipient, _amountOfShares);
-        // emit event after minting shares because we are always having the net new ether under the hood
-        // for vaults we have new locked ether and for fees we have a part of rewards
         _emitTransferAfterMintingShares(_recipient, _amountOfShares);
 
         emit ExternalSharesMinted(_recipient, _amountOfShares, getPooledEthByShares(_amountOfShares));
@@ -689,6 +688,7 @@ contract Lido is Versioned, StETHPermit, AragonApp {
     /**
      * @notice Burn external shares from the `msg.sender` address
      * @param _amountOfShares Amount of shares to burn
+     * @dev can be called only by VaultHub
      */
     function burnExternalShares(uint256 _amountOfShares) external {
         require(_amountOfShares != 0, "BURN_ZERO_AMOUNT_OF_SHARES");
@@ -703,8 +703,12 @@ contract Lido is Versioned, StETHPermit, AragonApp {
         _burnShares(msg.sender, _amountOfShares);
 
         uint256 stethAmount = getPooledEthByShares(_amountOfShares);
-        _emitTransferEvents(msg.sender, address(0), stethAmount, _amountOfShares);
-        emit ExternalSharesBurned(msg.sender, _amountOfShares, stethAmount);
+
+        // Historically, Lido contract does not emit Transfer to zero address events
+        // for burning but emits SharesBurnt instead, so it's kept here for compatibility
+        // we use the same `stethAmount` here as external shares burn does not change share rate
+        _emitSharesBurnt(msg.sender, stethAmount, stethAmount, _amountOfShares);
+        emit ExternalSharesBurnt(_amountOfShares);
     }
 
     /**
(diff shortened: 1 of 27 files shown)


contracts/0.4.24/Lido.sol
```diff
@@ -154,7 +159,7 @@ contract Lido is Versioned, StETHPermit, AragonApp {
         uint256 postBufferedEther
     );
 
-    // Emitted when token is rebased (total supply and/or total shares were changed)
+    // Emitted when the token is rebased (an oracle report is delivered)
```
TheDZhon:
```suggestion
    // Emitted when the token is rebased (an accounting oracle report is delivered)
```

contracts/0.4.24/Lido.sol
```diff
@@ -192,7 +197,7 @@ contract Lido is Versioned, StETHPermit, AragonApp {
     event ExternalSharesMinted(address indexed receiver, uint256 amountOfShares, uint256 amountOfStETH);
```
TheDZhon: it's a bit weird that `ExternalSharesBurnt` and `ExternalSharesMinted` have asymmetry in arguments

contracts/0.4.24/Lido.sol
```diff
@@ -154,7 +159,7 @@ contract Lido is Versioned, StETHPermit, AragonApp {
         uint256 postBufferedEther
     );
 
-    // Emitted when token is rebased (total supply and/or total shares were changed)
+    // Emitted when the token is rebased (an oracle report is delivered)
     event TokenRebased(
```
TheDZhon, line 163: `sharesMintedAsFees` ⇒ `sharesMintedAsProtocolFees`

contracts/0.4.24/Lido.sol
```diff
@@ -235,12 +240,44 @@ contract Lido is Versioned, StETHPermit, AragonApp {
     function finalizeUpgrade_v3(address _oldBurner, address[] _contractsWithBurnerAllowances) external {
         require(hasInitialized(), "NOT_INITIALIZED");
         _checkContractVersion(2);
+        _setContractVersion(3);
+
+        _migrateStorage();
+        _migrateBurner(_oldBurner, _contractsWithBurnerAllowances);
+    }
+
+    function _migrateStorage() internal {
```
TheDZhon:
```suggestion
    function _migrateStorage_v2_to_v3() internal {
```

contracts/0.4.24/Lido.sol
```diff
@@ -235,12 +240,44 @@ contract Lido is Versioned, StETHPermit, AragonApp {
...
+        );
+        CL_BALANCE_POSITION.setStorageUint256(0);
+        CL_VALIDATORS_POSITION.setStorageUint256(0);
+
+        bytes32 TOTAL_SHARES_POSITION = keccak256("lido.StETH.totalShares");
+        TOTAL_AND_EXTERNAL_SHARES_POSITION.setLowUint128(TOTAL_SHARES_POSITION.getStorageUint256());
+        TOTAL_SHARES_POSITION.setStorageUint256(0);
+    }
+
+    function _migrateBurner(address _oldBurner, address[] _contractsWithBurnerAllowances) internal {
```
TheDZhon:
```suggestion
    function _migrateBurner_v2_to_v3(address _oldBurner, address[] _contractsWithBurnerAllowances) internal {
```

contracts/0.4.24/Lido.sol
```diff
@@ -850,8 +873,18 @@ contract Lido is Versioned, StETHPermit, AragonApp {
     }
 
     /**
-     * @notice Emit the `TokenRebase` event
-     * @dev It's here for back compatibility reasons
+     * @notice Emits the `TokenRebase` and `InternalShareRateUpdated` events
```
TheDZhon, line 887: why do we need `InternalShareRateUpdated` when we already have `TokenRebased` and share rate with or without external vaults should be the same before and after?

TheDZhon (approved):
Overall looks good, a few nitpicks only

contracts/0.4.24/Lido.sol
folkyatina: For events, names of the args are included in abi, I'd rather keep as it was.

contracts/0.4.24/Lido.sol
folkyatina: To calculate gross share rate if we need it

contracts/0.4.24/Lido.sol
```diff
@@ -235,18 +242,59 @@ contract Lido is Versioned, StETHPermit, AragonApp {
...
+            CL_BALANCE_POSITION.getStorageUint256(),
+            CL_VALIDATORS_POSITION.getStorageUint256()
+        );
+        CL_BALANCE_POSITION.setStorageUint256(0);
+        CL_VALIDATORS_POSITION.setStorageUint256(0);
+
+        bytes32 TOTAL_SHARES_POSITION = keccak256("lido.StETH.totalShares");
+        uint256 totalShares = TOTAL_SHARES_POSITION.getStorageUint256();
+        assert(totalShares > 0); // sanity check
+        TOTAL_AND_EXTERNAL_SHARES_POSITION.setLowUint128(totalShares);
```
TheDZhon, line 281: would expect to have a dedicated named setter here tbh

contracts/0.4.24/StETH.sol
```diff
@@ -471,7 +466,7 @@ contract StETH is IERC20, Pausable {
      * @return the total amount of shares in existence.
      */
     function _getTotalShares() internal view returns (uint256) {
```
TheDZhon, line 468: probably better add _setTotalShares() too

TheDZhon (approved):
LGTM, just a single comment on the compressed storage usage.

contracts/0.4.24/StETH.sol
folkyatina: Not sure about that. During da-to-day general activity, there is no need in such a setter and moreover it can be pretty dangerous. So, as we are keeping this as a one time experience, I'd rather abstain.

PR #1404, opened 2025-08-27
[VAULTS] add simulatedShareRate to oracle report

### LINKED ISSUE #1220: [Bug]: full withdrawal of shares amount equal to internal shares cause DOS to core protocol
### Summary

since the external shares are withdraw able we can do a full internal shares withdraw that was not possible till now  given the initial mint to 0xdead address.
now we can mint (external mint with stVaults) extra shares and request to withdraw all of them. after this withdraw the internal shares is 0, internal shares == total_shares-external_shares
all the calculations (of shares to assets and vise versa) will revert in this case from div by 0 revert.
https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.4.24/Lido.sol#L1001

### Expected Behavior

in case of trying to do a full withdraw of all internal shares it is expected that the the withdrawal queue will block this or the core could recover using the last ratio or something similar.
note that blocking the number of external shares you can mint will not solve this since as long as there is a more external shares than the initial mint to 0xdead this vector is possible.

### Potential Impact

in this case the system will be DOSed.

### Steps to Reproduce

mint some external shares and do a full withdraw of the internal shares amount with the withdrawal queue.

### Possible Solutions

_No response_

### Guidelines

### LINKED ISSUE #1257: [Bug]: rebase smoothing should work with internal ether and shares
### Summary

https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.8.9/Accounting.sol#L194-L195 should be the internal shares and internal ETH
otherwise the limiter will not work correctly, which will allow more rebase than allowed by the positive rebase rate
the reason is that using the total shares will cause the increase in ETH and decrease in shares to appear smaller than it really is, letting the protocol to burn more than it should.

### Expected Behavior

-

### Potential Impact

rate will increase more than it should

### Steps to Reproduce

-

### Possible Solutions

_No response_

### Guidelines

### LINKED ISSUE #1359: [Bug]: Withdrawals finalization might leave redundant ether as a resulf of `_withdrawalsShareRate` change between the ref slot and on-chain calculation during the accounting report submission
### Summary

`_withdrawalsShareRate` gets calculated on-chain as a part of the accounting oracle-driven token rebase execution. Previously this param was calculated by the oracle and reported through a sanity checker, see: https://github.com/lidofinance/core/blob/master/contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol#L802-L833

### Expected Behavior

Withdrawals finalization should work coherently (providing the exact ether amount to finalize each withdrawal request with the appropriate share rate)

### Potential Impact

- Some of the oracle reports might send extra unclaimable ether to the WQ

### Steps to Reproduce

**Preconditions**

1. Set protocol state at reference slot **S**: `totalPooledEther = 100 ETH`, `totalShares = 100`.
2. Have a pending withdrawals batch totaling **20 shares** ready to be finalized.
    - 10 shares submitted with shareRate 1.0
    - 10 shares with shareRate 1.3
4. Prepare an oracle report for **S** with:
   * **Report(positive rebase):** `+5 ETH` and instruction to finalize the 20-share batch.
   * withdrawal batch will contain both request in one batch as simulatedShareRate on the Oracle will be 1.05 

**Race: mutate state after the ref slot**
4\. Before submitting the oracle report transaction, have an external actor deposit **100 ETH** (pre-rebase 1:1), resulting in `totalPooledEther = 200`, `totalShares = 200`.

**Submit and observe**
5\. Submit the oracle accounting report tx for **S**
6\. Observe `_withdrawalsShareRate` computed **on-chain at tx time**:

`(200 + 5) / 200 = 1.025` (ref-slot expectation was `105/100 = 1.05`).

7. Finalization amounts differ from what was intended at the ref slot:

   *  20 shares close as one batch but with maxShareRate = *1.025* which leads to 20.5 ETH being sent to WQ
   *  10 shares will be claimed as nominal for 10 ETH
   *  10 shares will be claimed as discounted for 10,25 ETH 

Which leaves 0.05 ETH stuck on WQ

**Expected vs actual**
8\. **Expected:** Withdrawals finalize using the share rate derived from the **reference slot S** (stable w\.r.t. intervening deposits) OR the appropriate ether amount according to the actual share rate.
9\. **Actual:** `_withdrawalsShareRate` is derived from **current state at tx time**, which can diverge from ref-slot math when balances change between S and the report transaction while leaving the same amount of ether to be sent to the WQ, leading to inconsistent/insufficient finalization amounts (especially in the negative rebase case).

### Possible Solutions

- Rollback to the previous scheme

### Guidelines

### LINKED ISSUE #1362: [Bug]: Function Visibility Can Be Changed
**Description**
The `simulateOracleReport` function of the `Accounting` contract is declared as `public view`, but it can be safely exposed externally.

Recommendation
Change the visibility of `simulateOracleReport` from public to `external` to:
- Reduce unnecessary gas overhead when called externally
- Clarify its intended interface usage for off-chain tooling or integrations

### LINKED ISSUE #1365: [Bug]: wrong order of calls during the Oracle report processing can lead to wrong Transfer events parameters
### Summary

In `_applyOracleReportContext` in Accounting.sol the https://github.com/lidofinance/core/blob/72fa8a5878a8fc27863efa0f9019344806c12f0f/contracts/0.8.9/Accounting.sol#L314 
where the fees are transferred goes before 
https://github.com/lidofinance/core/blob/72fa8a5878a8fc27863efa0f9019344806c12f0f/contracts/0.8.9/Accounting.sol#L317

where the values that contributes to the new totalPooledEther and totalShares are updates, so all transfer events will be emitted with wrong shareRate and, moreover, 

https://github.com/lidofinance/core/blob/72fa8a5878a8fc27863efa0f9019344806c12f0f/contracts/0.8.9/Accounting.sol#L410

passes the execution flow further to the modules having inconsistent protocol state

### Expected Behavior

Update all the state in Lido contract fist and make transfers after that

### Potential Impact

Broken indexers for Transfer events and maybe some jumpstart for reentancy if some module will be added that is  not safe enough

### Steps to Reproduce

-

### Possible Solutions

_No response_

### Guidelines

dry914:
Fix:
- https://github.com/lidofinance/core/issues/1359
- https://github.com/lidofinance/core/issues/1365
- limited fix for https://github.com/lidofinance/core/issues/1220
- https://github.com/lidofinance/core/issues/1257
- https://github.com/lidofinance/core/issues/1362

### DIFF
--- a/contracts/0.8.9/Accounting.sol
+++ b/contracts/0.8.9/Accounting.sol
@@ -129,19 +129,11 @@ contract Accounting {
         Contracts memory contracts = _loadOracleReportContracts();
         if (msg.sender != contracts.accountingOracle) revert NotAuthorized("handleOracleReport", msg.sender);
 
-        (PreReportState memory pre, CalculatedValues memory update) = _calculateOracleReportContext(contracts, _report);
+        PreReportState memory pre = _snapshotPreReportState(contracts);
+
+        CalculatedValues memory update = _simulateOracleReport(contracts, pre, _report, _report.simulatedShareRate);
 
         _applyOracleReportContext(contracts, _report, pre, update);
-    }
-
-    /// @dev prepare all the required data to process the report
-    function _calculateOracleReportContext(
-        Contracts memory _contracts,
-        ReportValues calldata _report
-    ) internal view returns (PreReportState memory pre, CalculatedValues memory update) {
-        pre = _snapshotPreReportState(_contracts);
-
-        update = _simulateOracleReport(_contracts, pre, _report, _report.simulatedShareRate);
     }
 
     /// @dev reads the current state of the protocol to the memory
@@ -329,17 +321,6 @@ contract Accounting {
             _update.postInternalEther,
             _update.sharesToMintAsFees
         );
-
-        // Sanity check for the provided simulated share rate
-        if (_report.withdrawalFinalizationBatches.length != 0) {
-            _contracts.oracleReportSanityChecker.checkSimulatedShareRate(
-                _update.postTotalPooledEther,
-                _update.postTotalShares,
-                _update.etherToFinalizeWQ,
-                _update.totalSharesToBurn,
-                _report.simulatedShareRate
-            );
-        }
     }
 
     /// @dev checks the provided oracle data internally and against the sanity checker contract
@@ -370,6 +351,17 @@ contract Accounting {
             _contracts.oracleReportSanityChecker.checkWithdrawalQueueOracleReport(
                 _report.withdrawalFinalizationBatches[_report.withdrawalFinalizationBatches.length - 1],
                 _report.timestamp
+            );
+        }
+
+        // Sanity check for the provided simulated share rate
+        if (_report.withdrawalFinalizationBatches.length != 0) {
+            _contracts.oracleReportSanityChecker.checkSimulatedShareRate(
+                _update.postTotalPooledEther,
+                _update.postTotalShares,
+                _update.etherToFinalizeWQ,
+                _update.sharesToBurnForWithdrawals,
+                _report.simulatedShareRate
             );
         }
     }
--- a/contracts/0.8.9/Accounting.sol
+++ b/contracts/0.8.9/Accounting.sol
@@ -108,17 +108,13 @@ contract Accounting {
 
     /// @notice calculates all the state changes that is required to apply the report
     /// @param _report report values
-    /// @param _withdrawalShareRate maximum share rate used for withdrawal finalization
-    ///                             if _withdrawalShareRate == 0, no withdrawals are
-    ///                             simulated
     function simulateOracleReport(
-        ReportValues calldata _report,
-        uint256 _withdrawalShareRate
+        ReportValues calldata _report
     ) public view returns (CalculatedValues memory update) {
         Contracts memory contracts = _loadOracleReportContracts();
         PreReportState memory pre = _snapshotPreReportState(contracts);
 
-        return _simulateOracleReport(contracts, pre, _report, _withdrawalShareRate);
+        return _simulateOracleReport(contracts, pre, _report);
     }
 
     /// @notice Updates accounting stats, collects EL rewards and distributes collected rewards
@@ -130,7 +126,7 @@ contract Accounting {
 
         PreReportState memory pre = _snapshotPreReportState(contracts);
 
-        CalculatedValues memory update = _simulateOracleReport(contracts, pre, _report, _report.simulatedShareRate);
+        CalculatedValues memory update = _simulateOracleReport(contracts, pre, _report);
 
         _applyOracleReportContext(contracts, _report, pre, update);
     }
@@ -150,20 +146,16 @@ contract Accounting {
     function _simulateOracleReport(
         Contracts memory _contracts,
         PreReportState memory _pre,
-        ReportValues calldata _report,
-        uint256 _withdrawalsShareRate
+        ReportValues calldata _report
     ) internal view returns (CalculatedValues memory update) {
         update.preTotalShares = _pre.totalShares;
         update.preTotalPooledEther = _pre.totalPooledEther;
 
-        if (_withdrawalsShareRate != 0) {
-            // Get the ether to lock for withdrawal queue and shares to move to Burner to finalize requests
-            (update.etherToFinalizeWQ, update.sharesToFinalizeWQ) = _calculateWithdrawals(
-                _contracts,
-                _report,
-                _withdrawalsShareRate
-            );
-        }
+        // Get the ether to lock for withdrawal queue and shares to move to Burner to finalize requests
+        (update.etherToFinalizeWQ, update.sharesToFinalizeWQ) = _calculateWithdrawals(
+            _contracts,
+            _report
+        );
 
         // Principal CL balance is the sum of the current CL balance and
         // validator deposits during this report
@@ -219,13 +211,12 @@ contract Accounting {
     /// @dev return amount to lock on withdrawal queue and shares to burn depending on the finalization batch parameters
     function _calculateWithdrawals(
         Contracts memory _contracts,
-        ReportValues calldata _report,
-        uint256 _simulatedShareRate
+        ReportValues calldata _report
     ) internal view returns (uint256 etherToLock, uint256 sharesToBurn) {
         if (_report.withdrawalFinalizationBatches.length != 0 && !_contracts.withdrawalQueue.isPaused()) {
             (etherToLock, sharesToBurn) = _contracts.withdrawalQueue.prefinalize(
                 _report.withdrawalFinalizationBatches,
-                _simulatedShareRate
+                _report.simulatedShareRate
             );
         }
     }
--- a/contracts/0.8.9/Accounting.sol
+++ b/contracts/0.8.9/Accounting.sol
@@ -129,27 +129,19 @@ contract Accounting {
         Contracts memory contracts = _loadOracleReportContracts();
         if (msg.sender != contracts.accountingOracle) revert NotAuthorized("handleOracleReport", msg.sender);
 
-        (
-            PreReportState memory pre,
-            CalculatedValues memory update,
-            uint256 withdrawalsShareRate
-        ) = _calculateOracleReportContext(contracts, _report);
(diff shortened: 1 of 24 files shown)


contracts/0.8.9/Accounting.sol
```diff
@@ -129,27 +129,19 @@ contract Accounting {
...
-            uint256 withdrawalsShareRate
-        ) = _calculateOracleReportContext(contracts, _report);
+        (PreReportState memory pre, CalculatedValues memory update) = _calculateOracleReportContext(contracts, _report);
 
-        _applyOracleReportContext(contracts, _report, pre, update, withdrawalsShareRate);
+        _applyOracleReportContext(contracts, _report, pre, update);
     }
 
     /// @dev prepare all the required data to process the report
     function _calculateOracleReportContext(
```
dry914: inline it

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -747,6 +789,55 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
             revert IncorrectRequestFinalization(statuses[0].timestamp);
     }
 
+    function _checkSimulatedShareRate(
+        LimitsList memory _limitsList,
+        uint256 _noWithdrawalsPostTotalPooledEther,
+        uint256 _noWithdrawalsPostTotalShares,
+        uint256 _simulatedShareRate
+    ) internal pure {
+        uint256 actualShareRate = (
```
dry914, line 799: calculate it in accounting

contracts/0.8.9/Accounting.sol
```diff
@@ -338,6 +329,17 @@ contract Accounting {
             _update.postInternalEther,
             _update.sharesToMintAsFees
         );
+
+        // Sanity check for the provided simulated share rate
+        if (_report.withdrawalFinalizationBatches.length != 0) {
```
dry914: move it to all sanity checks

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -747,6 +789,55 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+        // 2) user-submitted ether & minted `stETH` don't exceed the current staking rate limit
+        // (see Lido.getCurrentStakeLimit())
+        //
+        // can conclude that `simulatedShareRateDeviationBPLimit` (L) should be set as follows:
+        // L = (2 * SRL) * max(CLD, MPR),
+        // where:
+        // - CLD is consensus layer one-off balance decrease (as BP),
+        // - MPR is max positive token rebase (as BP),
+        // - SRL is staking rate limit normalized by TVL (`maxStakeLimit / totalPooledEther`)
+        //   totalPooledEther should be chosen as a reasonable lower bound of the protocol TVL
```
TheDZhon, line 827: it's no longer true for SRL as vaults can mint and burn stETH :warning:

TheDZhon (changes requested):
:warning:

contracts/0.8.9/Accounting.sol
```diff
@@ -109,17 +108,13 @@ contract Accounting {
 
     /// @notice calculates all the state changes that is required to apply the report
     /// @param _report report values
-    /// @param _withdrawalShareRate maximum share rate used for withdrawal finalization
-    ///                             if _withdrawalShareRate == 0, no withdrawals are
-    ///                             simulated
     function simulateOracleReport(
```
TheDZhon, line 115: let's describe the oracle flow briefly here:
1. simulate with zero withdrawals
2. take returned values
3. calculate withdrawals
4. pass a real report

or smth

contracts/0.8.9/Accounting.sol
```diff
@@ -167,22 +146,16 @@ contract Accounting {
     function _simulateOracleReport(
         Contracts memory _contracts,
         PreReportState memory _pre,
-        ReportValues calldata _report,
-        uint256 _withdrawalsShareRate
+        ReportValues calldata _report
     ) internal view returns (CalculatedValues memory update) {
         update.preTotalShares = _pre.totalShares;
```
TheDZhon, line 152: do we need this? bc `_pre` seems to be available here and there

contracts/0.8.9/Accounting.sol
```diff
@@ -167,22 +146,16 @@ contract Accounting {
     function _simulateOracleReport(
```
TheDZhon, line 147: the comment above is obsolete ⇒ would rather need to state that batches length is zero

contracts/0.8.9/Accounting.sol
```diff
@@ -198,8 +171,8 @@ contract Accounting {
             update.sharesToBurnForWithdrawals,
             update.totalSharesToBurn // shares to burn from Burner balance
         ) = _contracts.oracleReportSanityChecker.smoothenTokenRebase(
-            _pre.totalPooledEther,
-            _pre.totalShares,
+            _pre.totalPooledEther - _pre.externalEther,
+            _pre.totalShares - _pre.externalShares,
```
TheDZhon: would need to explicitly say that compat doc isn't changed

contracts/0.8.9/Accounting.sol
```diff
@@ -220,35 +193,81 @@ contract Accounting {
             - update.etherToFinalizeWQ; // withdrawals
```
TheDZhon:
```suggestion
            - update.etherToFinalizeWQ;
```

TheDZhon (commented):
:eyes:

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
folkyatina: Looks like it's still true if we are using internal ether and shares as for the smoothing

contracts/0.8.9/Accounting.sol
```diff
@@ -365,13 +411,39 @@ contract Accounting {
...
+
+        for (uint256 i; i < length; ++i) {
+            uint256 moduleShares = sharesToMint[i];
+            if (moduleShares > 0) {
+                LIDO.mintShares(recipients[i], moduleShares);
+            }
+        }
+
+        uint256 treasuryShares = _feeDistribution.treasurySharesToMint;
+        if (treasuryShares > 0) { // an edge case when all fees goes to modules
```
TheDZhon:
```suggestion
        if (treasuryShares > 0) { // zero is an edge case when all fees goes to modules
```

contracts/0.8.9/Accounting.sol
```diff
@@ -365,13 +411,39 @@ contract Accounting {
         );
 
         if (_report.withdrawalFinalizationBatches.length > 0) {
+            _contracts.oracleReportSanityChecker.checkSimulatedShareRate(
```
TheDZhon, line 414: let's mention why internal and not total

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -747,6 +789,55 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
             revert IncorrectRequestFinalization(statuses[0].timestamp);
     }
 
+    function _checkSimulatedShareRate(
```
TheDZhon, line 792: probably can rename params to internal ones (and for smoothing)

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -747,6 +789,55 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+        uint256 _simulatedShareRate
+    ) internal pure {
+        uint256 actualShareRate = (
+            _noWithdrawalsPostTotalPooledEther * SHARE_RATE_PRECISION_E27
+        ) / _noWithdrawalsPostTotalShares;
+
+        if (actualShareRate == 0) {
+            // can't finalize anything if the actual share rate is zero
+            revert ActualShareRateIsZero();
+        }
```
TheDZhon:
```suggestion
        assert(actualShareRate > 0);
```

and remove the `ActualShareRateIsZero` definition

TheDZhon (approved):
LGTM

PR #1407, opened 2025-08-29
fix: error on stepsFile

zkpepe:
## Context

`runMigrations` is called with `stepsFile`, which we check exists right before.

## Problem

TS still thinks `stepsFile` might be `undefined`, causing a type error.

## Solution

added a non-null assertion (`!`) to `stepsFile` when calling `runMigrations` to clear the type error.


tamtamchik:

PR #1408, opened 2025-08-29
⚠️⚠️⚠️ move Ownable2Step to BeaconProxy [WIP]

failingtwice:
To move all ossification processes into BeaconProxy, we need to move Ownable to the proxy to be able to authenticate ossify call.


failingtwice:
Closed due to a big scope of changes for little benefit.

PR #1409, opened 2025-09-01
fix(tw-deploy): Fix slots for verifier

hweawer:
Fix the first supported slot for TW deploy of verifier.

## Context

Fix the first supported slot for TW deploy of verifier.

## Problem

EIP 7600 was activated on hoodi by default.

### DIFF
--- a/scripts/triggerable-withdrawals/tw-deploy.ts
+++ b/scripts/triggerable-withdrawals/tw-deploy.ts
@@ -112,8 +112,8 @@ async function main(): Promise<void> {
   const BLOCK_ROOT_IN_SUMMARY_PREV_GINDEX = "0x000000000000000000000000000000000000000000000000000000000040000d";
   const BLOCK_ROOT_IN_SUMMARY_CURR_GINDEX = BLOCK_ROOT_IN_SUMMARY_PREV_GINDEX;
 
-  const FIRST_SUPPORTED_SLOT = 1;
-  const PIVOT_SLOT = 1;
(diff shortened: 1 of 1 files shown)


F4ever (approved):

scripts/triggerable-withdrawals/tw-deploy.ts
```diff
@@ -112,8 +112,8 @@ async function main(): Promise<void> {
   const BLOCK_ROOT_IN_SUMMARY_PREV_GINDEX = "0x000000000000000000000000000000000000000000000000000000000040000d";
   const BLOCK_ROOT_IN_SUMMARY_CURR_GINDEX = BLOCK_ROOT_IN_SUMMARY_PREV_GINDEX;
 
-  const FIRST_SUPPORTED_SLOT = 1;
-  const PIVOT_SLOT = 1;
+  const FIRST_SUPPORTED_SLOT = 364032 * SLOTS_PER_EPOCH; // https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7600.md#activation
+  const PIVOT_SLOT = FIRST_SUPPORTED_SLOT;
```
tamtamchik, line 116: Why do we need PIVOT_SLOT then?

tamtamchik (approved):
LGTM in general, but why do we need same vars twice?

scripts/triggerable-withdrawals/tw-deploy.ts
hweawer: PIVOT_SLOT is a parameter which can differ from FIRST_SUPPORTED_SLOT during upgrades of the contract, for the initial deployment they can be same

PR #1423, opened 2025-09-05
Merge develop with TW into testnet-2

arwer13:
Also make migration process.exit if it fails instead of trying to continue for long with the already invalid setup.


contracts/upgrade/TWVoteScript.sol
TheDZhon, line 1: why do we need this contract?

deployed-mainnet.json
TheDZhon, line 1: let's revisit after TW ugprade

contracts/common/interfaces/ICSModule.sol
TheDZhon, line 1: do we need this file?

TheDZhon (commented):
:eyes:

PR #1427, opened 2025-09-07
fix: script syntax and quote network variable

hexcow:
## Context

the deployment script is used to run Hardhat tasks with a specified network.

## Problem

the script had `set -e +u`, which is invalid, and `$NETWORK` wasn’t quoted, causing potential issues if it contains spaces or special characters.

## Solution

changed `set -e +u` → `set -e` and wrapped `$NETWORK` in quotes in the Hardhat command. Everything else stays the same.


tamtamchik:

PR #1435, opened 2025-09-10
fix: pause acceptor vault's deposits / require fresh report on rebalance

### LINKED ISSUE #1391: [Bug]: Missing fresh report requirements enable operations on stale vault data

### LINKED ISSUE #1471: [Bug]: should check pause deposits for a vault that receive badDebt
### Summary

they should not be able to deposit eth in this state 

### Expected Behavior

-

### Potential Impact

-

### Steps to Reproduce

-

### Possible Solutions

_No response_

### Guidelines

tamtamchik:
- Pause the vault acceptor in case of obligations.
- Added comments to functions that need a fresh report, and included a request for a fresh report in:
   - rebalance
   - resumeBeaconChainDeposits
- Fixed comment for permissionless forceValidatorExit.

### DIFF
--- a/contracts/0.8.25/vaults/VaultHub.sol
+++ b/contracts/0.8.25/vaults/VaultHub.sol
@@ -386,12 +386,9 @@ contract VaultHub is PausableUntilWithRoles {
     /// @param _liabilitySharesTarget maximum amount of liabilityShares that will be preserved, the rest will be
     ///         marked as redemptionShares. If value is greater than liabilityShares, redemptionShares are set to 0
     /// @dev NB: Mechanism to be triggered when Lido Core TVL <= stVaults TVL
-    /// @dev requires the fresh report
     function setLiabilitySharesTarget(address _vault, uint256 _liabilitySharesTarget) external onlyRole(REDEMPTION_MASTER_ROLE) {
         VaultConnection storage connection = _checkConnection(_vault);
         VaultRecord storage record = _vaultRecord(_vault);
-
-        _requireFreshReport(_vault, record);
 
         uint256 liabilityShares_ = record.liabilityShares;
         uint256 redemptionShares = liabilityShares_ > _liabilitySharesTarget ? liabilityShares_ - _liabilitySharesTarget : 0;
@@ -765,13 +762,11 @@ contract VaultHub is PausableUntilWithRoles {
     /// @param _amountOfShares amount of shares to burn
     /// @dev msg.sender should be vault's owner
     /// @dev this function is designed to be used by the smart contract, for EOA see `transferAndBurnShares`
-    /// @dev requires the fresh report
     function burnShares(address _vault, uint256 _amountOfShares) public whenResumed {
         _requireNotZero(_amountOfShares);
         _checkConnectionAndOwner(_vault);
 
         VaultRecord storage record = _vaultRecord(_vault);
-        _requireFreshReport(_vault, record);
 
         _decreaseLiability(_vault, record, _amountOfShares);
 
@@ -786,7 +781,6 @@ contract VaultHub is PausableUntilWithRoles {
     /// @param _vault vault address
     /// @param _amountOfShares amount of shares to transfer and burn
     /// @dev msg.sender should be vault's owner
-    /// @dev requires the fresh report
     function transferAndBurnShares(address _vault, uint256 _amountOfShares) external {
         LIDO.transferSharesFrom(msg.sender, address(this), _amountOfShares);
 
--- a/contracts/0.8.25/vaults/VaultHub.sol
+++ b/contracts/0.8.25/vaults/VaultHub.sol
@@ -589,6 +589,8 @@ contract VaultHub is PausableUntilWithRoles {
                 _overrideOperatorLimits: true
             });
 
+            _updateBeaconChainDepositsPause(_vaultAcceptor, acceptorRecord, acceptorConnection);
+
             emit BadDebtSocialized(_badDebtVault, _vaultAcceptor, badDebtSharesToAccept);
         }
 
--- a/contracts/0.8.25/vaults/VaultHub.sol
+++ b/contracts/0.8.25/vaults/VaultHub.sol
@@ -713,7 +713,10 @@ contract VaultHub is PausableUntilWithRoles {
         _requireNotZero(_shares);
         _checkConnectionAndOwner(_vault);
 
-        _rebalance(_vault, _vaultRecord(_vault), _shares);
+        VaultRecord storage record = _vaultRecord(_vault);
+        _requireFreshReport(_vault, record);
+
+        _rebalance(_vault, record, _shares);
     }
 
     /// @notice mint StETH shares backed by vault external balance to the receiver address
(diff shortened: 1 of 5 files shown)


contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -713,7 +715,10 @@ contract VaultHub is PausableUntilWithRoles {
         _requireNotZero(_shares);
         _checkConnectionAndOwner(_vault);
 
-        _rebalance(_vault, _vaultRecord(_vault), _shares);
+        VaultRecord storage record = _vaultRecord(_vault);
+        _requireFreshReport(_vault, record);
+
+        _rebalance(_vault, record, _shares);
```
TheDZhon, line 728: let's put `_requireFreshReport` inside `_rebalance`

TheDZhon: or even find a more common place (withdraw semantics)

TheDZhon (commented):
LGTM, except only for some refactoring to make the code more local.

contracts/0.8.25/vaults/VaultHub.sol
tamtamchik: Moving inside causes duplication in many areas where we need to verify the total value beforehand, so I will stick with the current approach, but I will enforce report freshness in all external functions where the total value is needed (e.g., resume beacon chain deposits).

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -385,11 +385,14 @@ contract VaultHub is PausableUntilWithRoles {
...
     /// @param _liabilitySharesTarget maximum amount of liabilityShares that will be preserved, the rest will be
     ///         marked as redemptionShares. If value is greater than liabilityShares, redemptionShares are set to 0
-    /// @dev NB: Mechanism to be triggered when Lido Core TVL <= stVaults TVL.
+    /// @dev NB: Mechanism to be triggered when Lido Core TVL <= stVaults TVL
+    /// @dev requires the fresh report
     function setLiabilitySharesTarget(address _vault, uint256 _liabilitySharesTarget) external onlyRole(REDEMPTION_MASTER_ROLE) {
         VaultConnection storage connection = _checkConnection(_vault);
         VaultRecord storage record = _vaultRecord(_vault);
 
+        _requireFreshReport(_vault, record);
```
TheDZhon: why though?

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -749,11 +765,13 @@ contract VaultHub is PausableUntilWithRoles {
     /// @param _amountOfShares amount of shares to burn
     /// @dev msg.sender should be vault's owner
     /// @dev this function is designed to be used by the smart contract, for EOA see `transferAndBurnShares`
+    /// @dev requires the fresh report
     function burnShares(address _vault, uint256 _amountOfShares) public whenResumed {
         _requireNotZero(_amountOfShares);
         _checkConnectionAndOwner(_vault);
 
         VaultRecord storage record = _vaultRecord(_vault);
+        _requireFreshReport(_vault, record);
```
TheDZhon: feels extra, anyways we require it to initiate disconnect

TheDZhon (commented):
:eyes:

contracts/0.8.25/vaults/VaultHub.sol
tamtamchik: I thought about pausing the vault in an unhealthy state or obligations, but it indeed does not make sense since bringing the report will pause it anyway.

contracts/0.8.25/vaults/VaultHub.sol
tamtamchik: As above, bringing the report by itself will pause the vault (if required), so there is no need to force it here, as long as some external monitoring exists.

PR #1445, opened 2025-09-15
SRv3: accounting

eddort:


krogla (approved):
not tested deeply, but lgtm

PR #1451, opened 2025-09-17
fix: quar calculation /  updateVaultShareLimit shortcut / totalValue underflow

### LINKED ISSUE #1342: [Bug]: making a vault get stuck via gifting
### Summary

withdrawing more than the value of the vault using gifting when slashing happens, this will cause the oracle report to not work https://github.com/lidofinance/core/blob/22cab0f0372015f2d2fce8bede64e98beae28571/contracts/0.8.25/vaults/LazyOracle.sol#L311 and eth gets stuck unless the owner fund the missing eth to the vault

### Expected Behavior

-

### Potential Impact

-

### Steps to Reproduce

-

### Possible Solutions

_No response_

### Guidelines

dry914:
Fixes:
1. https://github.com/lidofinance/core/issues/1342
2. remove `updateVaultShareLimit` shortcut
3. light refactoring of `_processTotalValue` calculation. Comment from Certora:

> in _processTotalValue why do we need both conditons here _reportedTotalValue <= quarantineThreshold || totalValueIncrease <= maxIncreaseWithRewards line 502
> i think that when the first condition is true the second is true so not sure we need the first one.


contracts/0.8.25/vaults/LazyOracle.sol
```diff
@@ -499,7 +499,9 @@ contract LazyOracle is ILazyOracle, AccessControlEnumerableUpgradeable {
             uint256 maxIncreaseWithRewards = quarantinedValue +
                 (onchainTotalValueOnRefSlot + quarantinedValue) * $.maxRewardRatioBP / TOTAL_BASIS_POINTS;
 
-            if (_reportedTotalValue <= quarantineThreshold || totalValueIncrease <= maxIncreaseWithRewards) {
+            // NB: No need to check "_reportedTotalValue <= quarantineThreshold" separately since
+            // when that condition is true, "totalValueIncrease <= maxIncreaseWithRewards" is always true as well
+            if (totalValueIncrease <= maxIncreaseWithRewards) {
```
TheDZhon: can we reformulate the first to be here `_reportedTotalValue`, not losing behavior cohesiveness

test/integration/vaults/lazyOracle.integration.ts
```diff
@@ -642,6 +642,51 @@ describe("Integration: LazyOracle", () => {
...
+      // totalValueWithoutQuarantine + currentInOutDelta - inOutDeltaOnRefSlot
+      // = 998 + 1 ETH - 1000 ETH
+      await expect(
+        reportVaultDataWithProof(ctx, stakingVault, {
+          totalValue: slashedTotalValue,
+          waitForNextRefSlot: false,
+        }),
+      ).to.be.revertedWithCustomError(lazyOracle, "UnderflowInTotalValueCalculation");
+
+      // but it works after waiting for next refSlot
```
TheDZhon, line 688: if attacker continues to repeat this, the freshness condition would prevent withdrawals

TheDZhon (commented):
:+1:

TheDZhon (approved):
LGTM

PR #1468, opened 2025-09-25
[VAULTS] Improve PDG UX

folkyatina:
Better UX for proveWCActivateAndTopUpValidators allowing it to skip stages instead of reverting


TheDZhon (approved):
GG 👍

PR #1493, opened 2025-10-02
fix(Lido): make staking pause/resume revert if paused/resumed

### LINKED ISSUE #758: Prevent possible `pauseStaking` and `resumeStaking` repetitive calls
Now it's possible to call one of the `pauseStaking` or `resumeStaking` methods more than once in a row.
Even though it doesn't pose any risks or inconsistencies beyond emitting extra events, the better approach would be completely preventing repetitive calls. The latter approach is used everywhere with other 'pausable' entities across the codebase.

folkyatina:
Fix #758

### DIFF
--- a/contracts/0.4.24/Lido.sol
+++ b/contracts/0.4.24/Lido.sol
@@ -316,6 +316,7 @@ contract Lido is Versioned, StETHPermit, AragonApp {
      */
     function pauseStaking() external {
         _auth(STAKING_PAUSE_ROLE);
+        require(!STAKING_STATE_POSITION.getStorageStakeLimitStruct().isStakingPaused(), "ALREADY_PAUSED");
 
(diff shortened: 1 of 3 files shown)


contracts/0.4.24/Lido.sol
```diff
@@ -316,6 +316,7 @@ contract Lido is Versioned, StETHPermit, AragonApp {
      */
     function pauseStaking() external {
         _auth(STAKING_PAUSE_ROLE);
+        require(!STAKING_STATE_POSITION.getStorageStakeLimitStruct().isStakingPaused(), "ALREADY_PAUSED");
```
TheDZhon: why not make isStakingPaused to be `public` and call it here and there?

TheDZhon (approved):
LGTM!

Issue #1305, opened 2025-07-24
[Bug]: Shortfall value might not be enough for vault to become healthy


shoham-certora:
### Summary

When using `VaultHub.rebalanceShortfall` value for rebalancing an unhealthy vault (e.g. in `VaultHub.forceRebalance`), the vault may still remain unhealthy due to rounding issues. Particularly in cases where the vault's `reserveRatioBP` is equal to its `forcedRebalanceThresholdBP`.

### Expected Behavior

After calling `VaultHub.forceRebalance(vault)` or `VaultHub.rebalance(vault, rebalanceShortfall(vault)` for an unhealthy vault, it is expected that the vault will be healthy (i.e. `isVaultHealthy(vault)` to return true).

### Potential Impact

The vault will remain unhealthy, and will require additional rebalancing calls.

### Steps to Reproduce

[...]

PR #1494, opened 2025-10-03
feat: shortfall rework

failingtwice:

## Context
#1305 
 
---

Math

- LS = liability shares
- TVS = total value in shares (floor)
- B = 10000 (basis points base)
- T = thresholdBP
- M = B - T = mintRatio in bp.
- X = shortfall (misleading name btw)

---
Vault is healthy if:

`LS / TVS <= M / B`

or

`LS * B <= TVS * M`

---

Rebalance does the following

`LS - X`

rebalance(roundUp(X))

and

`TVS - Y`

Normally, `X == Y`

BUT

because rebalance rounds UP, in the worst case scenario

`Y = X + 1`

thus, we can derive the post-rebalance condition:

`(LS - X) * B <= (TVS - X - 1) * M`

OR (using distributive property)

`LS * B - X * B <= TVS * M - X * M - M` 

add `X * M` to both sides

`LS * B - X * B + X * M <= TVS * M - M` 

subtract `LS * B`

`- X * B + X * M <= TVS * M - M - LS * B` 

or factor X

`X (M - B) <= TVS * M - M - LS * B`

because

`T = M - B`

`X * (-T) <= TVS * M - M - LS * B`

divide both sides by -T and flip the sign

`X >= (TVS * M - M - LS * B) / (-T) `

move minus to numerator

`X >= -(TVS * M - M - LS * B) / T `

factor -1

`X >= (-TVS * M + M + L * B) / T `

rearrange and get the final inequality

**`X >= (L * B - TVS * M + M) / T `**

AND after all of that we need to use ceiling division because the final inequality also includes division

**`X >= (L * B - TVS * M + M) + (T - 1) /  T`**

---

### DIFF
--- a/contracts/0.8.25/vaults/VaultHub.sol
+++ b/contracts/0.8.25/vaults/VaultHub.sol
@@ -1167,46 +1167,34 @@ contract VaultHub is PausableUntilWithRoles {
         VaultConnection storage _connection,
         VaultRecord storage _record
     ) internal view returns (uint256) {
-        uint256 totalValue_ = _totalValue(_record);
-        uint256 liabilityShares_ = _record.liabilityShares;
-
-        bool isHealthy = !_isThresholdBreached(
-            totalValue_,
-            liabilityShares_,
-            _connection.forcedRebalanceThresholdBP
+        return _minimumRebalanceShares(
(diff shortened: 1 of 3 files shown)


failingtwice:
closed in favor of #1497

PR #1496, opened 2025-10-06
fix: upgrade after TW went live on mainnet

arwer13:


contracts/upgrade/V3VoteScript.sol
```diff
@@ -70,20 +70,50 @@ contract V3VoteScript is OmnibusBase {
...
-            description: "3. Set Lido implementation in Kernel",
+            description: "4. Set Lido implementation in Kernel",
             call: _forwardCall(
                 TEMPLATE.AGENT(),
                 TEMPLATE.KERNEL(),
                 abi.encodeCall(IKernel.setApp, (IKernel(TEMPLATE.KERNEL()).APP_BASES_NAMESPACE(), params.lidoAppId, TEMPLATE.NEW_LIDO_IMPL()))
             )
         });
 
+        // Revoke APP_MANAGER_ROLE from the AGENT on Kernel ACL
```
folkyatina, line 98: Do we have a policy on revoking all the roles now?

folkyatina (approved):
LGTM

tamtamchik (approved):

contracts/upgrade/V3VoteScript.sol
arwer13: I though we do, don't we?
And more roles were revoked on DG even

PR #1498, opened 2025-10-06
fix disconnect process

### LINKED ISSUE #1476: [Bug]: imposible to disconnect vault with more fees than the vault value even when done by the admin role
### Summary

if vault value is 0 (or less than fees) and there are fees plus non zero balance (more than value) in the vault it cannot be dissconnected. that is not the expected behaviour with the _forceFullFeesSettlement flag

after the report comes to fix the total value we can donate 1 wei each time and prevent this

### Expected Behavior

-

### Potential Impact

-

### Steps to Reproduce

-

### Possible Solutions

_No response_

### Guidelines

tamtamchik:
Fixes #1476


contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -1057,8 +1057,14 @@ contract VaultHub is PausableUntilWithRoles {
                 revert NoUnsettledLidoFeesShouldBeLeft(_vault, unsettledLidoFees);
             }
 
-            uint256 withdrawable = Math256.min(unsettledLidoFees, _vaultBalance);
-            _settleLidoFees(_vault, _record, _connection, withdrawable);
+            uint256 withdrawable = Math256.min(
```
TheDZhon: ⚠️ Here we can omit settling some fees which doesn't make math break even.

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -1052,13 +1052,15 @@ contract VaultHub is PausableUntilWithRoles {
...
-            uint256 _vaultBalance = _availableBalance(_vault);
-            if (_vaultBalance < unsettledLidoFees && _forceFullFeesSettlement) {
+            uint256 balance = Math256.min(_availableBalance(_vault), _totalValue(_record));
+            if (balance < unsettledLidoFees && _forceFullFeesSettlement) {
                 revert NoUnsettledLidoFeesShouldBeLeft(_vault, unsettledLidoFees);
             }
 
-            uint256 withdrawable = Math256.min(unsettledLidoFees, _vaultBalance);
-            _settleLidoFees(_vault, _record, _connection, withdrawable);
+            uint256 withdrawable = Math256.min(balance, unsettledLidoFees);
```
TheDZhon: looks a bit confusing from the first glance, prob would be good to branch on the `_forceFullFeesSettlement` at the first place 🙏

TheDZhon (approved):
LGTM, but can add some more expressiveness

TheDZhon (approved):
LGTM

PR #1500, opened 2025-10-07
fix: overflow in node operator fee

folkyatina:
what if in the addFeeExemption the NO gives a max_int128 _exemptedAmount then the settledGrowth is max_int128. if there is any slashing which the NO can cause, the onwer cannot dissconnect since he cannot _calculateFee() (udnerflow in unsettledGrowth calculation) and cannot pay it. this locks the owner

### DIFF
--- a/contracts/0.8.25/vaults/dashboard/NodeOperatorFee.sol
+++ b/contracts/0.8.25/vaults/dashboard/NodeOperatorFee.sol
@@ -318,16 +318,18 @@ contract NodeOperatorFee is Permissions {
      * @dev fee exemption can only be positive
      */
     function _addFeeExemption(uint256 _amount) internal {
-        _correctSettledGrowth(settledGrowth + _amount.toInt256());
+        if (_amount > type(uint104).max) revert UnexpectedFeeExemptionAmount();
+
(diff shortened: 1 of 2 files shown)


contracts/0.8.25/vaults/dashboard/NodeOperatorFee.sol
```diff
@@ -318,16 +318,18 @@ contract NodeOperatorFee is Permissions {
      * @dev fee exemption can only be positive
      */
     function _addFeeExemption(uint256 _amount) internal {
-        _correctSettledGrowth(settledGrowth + _amount.toInt256());
+        if (_amount > type(uint104).max) revert UnexpectedFeeExemptionAmount();
```
TheDZhon, line 321: just wondering why exactly uint104 🤔

TheDZhon (approved):
GG

contracts/0.8.25/vaults/dashboard/NodeOperatorFee.sol
folkyatina: It's just arbitrary relatively sane value for ether limit that we using in several places.

Issue #1381, opened 2025-08-18
[Bug]: `RoleMemberConfirmed`: inconsistent naming — `member` / `role` fields are effectively `address` / `address(256)`


DiRaiks:
### Summary

In the current implementation of the `RoleMemberConfirmed` event:  

```solidity
event RoleMemberConfirmed(
    address indexed member,
    bytes32 indexed role,
    uint256 confirmTimestamp,
    uint256 expiryTimestamp,
    bytes data
);
```
- The `member` field contains the address of the confirmer (`msg.sender`).
- The `role` field, however, is inconsistent:
  - In some contracts (e.g., `Dashboard`), it correctly stores the `keccak256` hash of a role.
  - In others (e.g., `OperatorGrid.changeTier`), it stores an **address encoded as bytes32** (e.g., `vaultOwner` or `nodeOperator`).

### Problem
- The naming of the event arguments (`member` and `role`) suggests semantic consistency, but in practice both fields may contain addresses.
- This inconsistency makes it impossible for tooling to reliably collect confirmations by role.
  - For example, any off-chain indexer trying to fetch confirmations for `ROLE_X` will miss those from `OperatorGrid`, since there `role` is an address instead of a role identifier.
- The mismatch between field names and actual stored values is misleading and error-prone for consumers of the logs.

### Affected Code (links)
- [OperatorGrid.sol#L399 — changeTier](https://github.com/lidofinance/core/blob/v3.0.0-audits/contracts/0.8.25/vaults/OperatorGrid.sol#L399) — passes addresses into _collectAndCheckConfirmations.
- [Dashboard.sol#L568 — setNodeOperatorFeeRate](https://github.com/lidofinance/core/blob/v3.0.0-audits/contracts/0.8.25/vaults/dashboard/Dashboard.sol#L568) — passes proper confirmingRoles() into _collectAndCheckConfirmations.
- [Permissions.sol#L350 — _transferVaultOwnership](https://github.com/lidofinance/core/blob/v3.0.0-audits/contracts/0.8.25/vaults/dashboard/Permissions.sol#L350) — another correct example with roles.

### Expected Behavior

- `role` should always contain a role identifier (`bytes32 keccak256("ROLE_NAME")`), consistent across all contracts.
- Alternatively, if addresses are intended, the event’s parameter names should be updated to accurately reflect their contents.

### Potential Impact

- Event consumers and monitoring tools cannot reliably query confirmations **by role**.
[...]

PR #1509, opened 2025-10-14
feat: rename event field

failingtwice:
## Context
#1381 

## Solution
rename field from `role` to `roleOrAddress`

image


TheDZhon (approved):
LGTM

PR #1513, opened 2025-10-17
feat: add vault factory chaining for upgrade

folkyatina:


TheDZhon (approved):
GG, though would suggest to add some tests with chained addrs (not just trivial zero inits)

PR #1515, opened 2025-10-19
fix: fixes after the review

folkyatina:
https://github.com/lidofinance/core/pull/874#pullrequestreview-3343452474

### DIFF
--- a/contracts/0.8.25/vaults/dashboard/Permissions.sol
+++ b/contracts/0.8.25/vaults/dashboard/Permissions.sol
@@ -93,12 +93,6 @@ abstract contract Permissions is AccessControlConfirmable {
     /// @dev 0x25482e7dc9e29f6da5bd70b6d19d17bbf44021da51ba0664a9f430c94a09c674
     bytes32 public constant VAULT_CONFIGURATION_ROLE = keccak256("vaults.Permissions.VaultConfiguration");
 
-    /**
-     * @notice Address of the implementation contract
-     * @dev Used to prevent initialization in the implementation
-     */
-    address private immutable _SELF;
-
     VaultHub public immutable VAULT_HUB;
     ILidoLocator public immutable LIDO_LOCATOR;
 
@@ -111,7 +105,8 @@ abstract contract Permissions is AccessControlConfirmable {
         _requireNotZero(_vaultHub);
         _requireNotZero(_lidoLocator);
 
-        _SELF = address(this);
+        initialized = true;
+
         // @dev vaultHub is cached as immutable to save gas for main operations
         VAULT_HUB = VaultHub(payable(_vaultHub));
         LIDO_LOCATOR = ILidoLocator(_lidoLocator);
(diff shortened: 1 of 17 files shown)


contracts/0.8.25/vaults/dashboard/Permissions.sol
```diff
@@ -93,12 +93,6 @@ abstract contract Permissions is AccessControlConfirmable {
     /// @dev 0x25482e7dc9e29f6da5bd70b6d19d17bbf44021da51ba0664a9f430c94a09c674
     bytes32 public constant VAULT_CONFIGURATION_ROLE = keccak256("vaults.Permissions.VaultConfiguration");
 
-    /**
-     * @notice Address of the implementation contract
-     * @dev Used to prevent initialization in the implementation
-     */
-    address private immutable _SELF;
```
TheDZhon, line 100: cool — it makes bytecode more verifiable btw

contracts/0.8.25/vaults/dashboard/Permissions.sol
```diff
@@ -111,7 +105,8 @@ abstract contract Permissions is AccessControlConfirmable {
         _requireNotZero(_vaultHub);
         _requireNotZero(_lidoLocator);
 
-        _SELF = address(this);
+        initialized = true;
```
TheDZhon, line 108: It's good that starting from OZ 5.x contracts line, they don't make a deployer account to be the DEFAULT_ADMIN_ROLE holder by default anymore 🧠

contracts/common/lib/TriggerableWithdrawals.sol
```diff
@@ -155,15 +155,15 @@ library TriggerableWithdrawals {
...
     }
 
-    function _copyPubkeyToMemory(bytes calldata pubkeys, bytes memory target, uint256 keyIndex) private pure {
+    function _copyPubkeyToMemory(bytes memory target, bytes calldata pubkeys, uint256 keyIndex) private pure {
         assembly {
             calldatacopy(add(target, 32), add(pubkeys.offset, mul(keyIndex, PUBLIC_KEY_LENGTH)), PUBLIC_KEY_LENGTH)
         }
     }
 
     function _copyAmountToMemory(bytes memory target, uint64 amount) private pure {
```
TheDZhon: it's a bit dangerous one as it can override the last part of a pubkey written before (not the case currently, but still), so _the new fn is not order-agnostic_, I would suggest to make it unbreakable again:

```solidity
function _copyAmountWithPubkeyToMemory(
    bytes memory target,
    uint64 amount,
    bytes calldata pubkeys,
    uint256 keyIndex
) private pure {
    assembly {
        // Write the amount first:
        // mstore at [56..88) → uint64 lands in [80..88), zeroes [56..80)
        mstore(add(target, 56), amount)

        // Then write the 48-byte pubkey into [32..80), overwriting the zeros above.
        calldatacopy(
            add(target, 32),
            add(pubkeys.offset, mul(keyIndex, PUBLIC_KEY_LENGTH)),
            PUBLIC_KEY_LENGTH
        )
    }
}
```

and get rid of `_copyAmountToMemory` by calling `_copyAmountWithPubkeyToMemory` instead of two funcs in a row

TheDZhon (approved):
LGTM, however would harden the assembly part a bit

PR #1520, opened 2025-10-21
[VAULTS] Hoodi tests

folkyatina:


lib/protocol/helpers/nor-sdvt.ts
```diff
@@ -138,9 +139,21 @@ export const norSdvtAddNodeOperator = async (
 
   const operatorId = await module.getNodeOperatorsCount();
 
-  const managerSigner = await ctx.getSigner("agent");
+  const role = await module.MANAGE_NODE_OPERATOR_ROLE();
+  const managerSigner = await impersonate(await acl.getPermissionManager(module.address, role), ether("100"));
```
tamtamchik, line 143: Do we not have permissions on Hoodi for MANAGE_NODE_OPERATOR_ROLE? Should we add them to sync with the mainnet?

tamtamchik: Same questions for other roles

test/integration/core/accounting.integration.ts
```diff
@@ -20,6 +20,8 @@ describe("Integration: Accounting", () => {
   before(async () => {
     ctx = await getProtocolContext();
 
+    await report(ctx);
```
tamtamchik, line 24: ⚠️ This report pollutes the state of the fork, as it precedes the snapshot and thus never reverted

tamtamchik (commented):
One concern is polluting the fork state. Additionally, perhaps it's better to remove 'sepolia' from the repo entirely, but may be done separately.

failingtwice (approved):

PR #1523, opened 2025-10-22
[VAULTS] fix for node operator fee disbursement on disconnect

folkyatina:


contracts/0.8.25/vaults/dashboard/Dashboard.sol
```diff
@@ -252,9 +252,16 @@ contract Dashboard is NodeOperatorFee {
...
+     * @dev node operator fees can be lost if the fee recipient is set incorrectly
      */
     function voluntaryDisconnect() external {
-        disburseFee();
+        // try catch prevents blocking disconnection by setting wrong fee receiver (e.g. to non-receiving contract)
+        // but also it can prevent some clients from doing correct gas estimation
+        // so it's recommended to invoke disburseFee() first if having problems with gas estimation
+        (uint256 fee,,) = _calculateFee();
+        if (fee > 0) {
+            try this.disburseFee() {} catch {}
```
krogla: a way to make gas estimation work correctly, but only if the called method doesn't have reverts with empty error data except “out of gas”

```solidity
try this.disburseFee() {} catch (bytes memory data) {
   if (data.length == 0) revert UnrecoverableError();
}
```

TheDZhon (approved):
LGTM

PR #1530, opened 2025-10-27
fix: total minting capacity

tamtamchik:


contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -1469,9 +1469,16 @@ contract VaultHub is PausableUntilWithRoles {
     /// @notice Calculates the max lockable value of the vault
     /// @param _record The record of the vault
```
TheDZhon, line 1470: `_deltaValue` natspec is missing

TheDZhon (approved):
LGTM, but a small nitpick

PR #1531, opened 2025-10-28
[VAULTS] update settledGrowth even if fees are 0

folkyatina:


contracts/0.8.25/vaults/dashboard/NodeOperatorFee.sol
```diff
@@ -293,8 +293,12 @@ contract NodeOperatorFee is Permissions {
     }
 
     function _disburseFee(uint256 fee, int256 growth, address _recipient) internal {
```
TheDZhon, line 295:
```suggestion
    function _disburseFee(uint256 fee, int256 growth, address _recipient) internal {
        _setSettledGrowth(growth);
        if (fee != 0) {
            _doWithdraw(_recipient, fee);
            emit FeeDisbursed(msg.sender, fee, _recipient);
        }
    }
```

Probably behavior here will be a bit more consistent, bc the previously suggested fix do the following:
```
fee > 0, growth == settledGrowth ⇒ revert
fee = 0, growth == settledGrowth ⇒ no-op
```

contracts/0.8.25/vaults/dashboard/NodeOperatorFee.sol
```diff
@@ -457,4 +460,9 @@ contract NodeOperatorFee is Permissions {
      * @dev Error emitted when the vault is quarantined.
      */
     error VaultQuarantined();
+
+    /**
+     * @dev Error emitted when you try to disburse fees but there is no unsettled growth
+     */
+    error NoUnsettledGrowth();
```
TheDZhon: unused

TheDZhon (changes requested):
👀

contracts/0.8.25/vaults/dashboard/NodeOperatorFee.sol
folkyatina: `if (growth > settledGrowth) ` is important here. Because we don't want to reduce it (it's possible on reconnect scenarios).

Actually, previous version is consistently no-op because if growth == settledGrowth => fee == 0, but not vice versa

TheDZhon: agree

PR #1532, opened 2025-10-28
fix(Lido): better check for maxExternalRatioBP in finalizeUpgrade_v3

folkyatina:

## Context

## Problem

## Solution

Your proposed solution


TheDZhon (approved):
GG

PR #1536, opened 2025-10-28
fix(LazyOracle): correct condition for maxLiabilityShares validation

folkyatina:
What I don't understand is why, when applying a new report for the vault, in the LazyOracle::_handleSanityChecks function, the last check reverts if the record.maxLiabilityShares  is higher than the reported _maxLiabilityShares . This comparison seems incorrect because the on-chain value (record.maxLiabilityShares) can legitimately exceed the reported value if shares are minted just after the oracle fetches the value.

### DIFF
--- a/contracts/0.8.25/vaults/LazyOracle.sol
+++ b/contracts/0.8.25/vaults/LazyOracle.sol
@@ -442,7 +442,9 @@ contract LazyOracle is ILazyOracle, AccessControlEnumerableUpgradeable {
             revert CumulativeLidoFeesTooLarge(_cumulativeLidoFees - previousCumulativeLidoFees, maxLidoFees);
         }
 
-        // 5. _maxLiabilityShares is greater or equal than _liabilityShares and current `maxLiabilityShares`
+        // 5. _maxLiabilityShares must be greater or equal than _liabilityShares
+        // _maxLiabilityShares must be less or equal than the currently tracked on-chain record.maxLiabilityShares
+        // (the latter can increase after the ref slot reported)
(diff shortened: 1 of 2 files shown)


contracts/0.8.25/vaults/LazyOracle.sol
```diff
@@ -443,7 +443,7 @@ contract LazyOracle is ILazyOracle, AccessControlEnumerableUpgradeable {
         }
 
         // 5. _maxLiabilityShares is greater or equal than _liabilityShares and current `maxLiabilityShares`
-        if (_maxLiabilityShares < _liabilityShares || _maxLiabilityShares < record.maxLiabilityShares) {
+        if (_maxLiabilityShares < _liabilityShares || _maxLiabilityShares > record.maxLiabilityShares) {
```
TheDZhon, line 448: would suggest to have a test for this fix tbh

contracts/0.8.25/vaults/LazyOracle.sol
```diff
@@ -443,7 +443,7 @@ contract LazyOracle is ILazyOracle, AccessControlEnumerableUpgradeable {
         }
 
         // 5. _maxLiabilityShares is greater or equal than _liabilityShares and current `maxLiabilityShares`
```
TheDZhon: this comment reads weird, should be smth like:
- _maxLiabilityShares must be greater or equal than _liabilityShares
- _maxLiabilityShares must be less or equal than the currently tracked on-chain `record.maxLiabilityShares` (the latter can increase after the ref slot reported)

TheDZhon (approved):
LGTM, but some food for thought

PR #1538, opened 2025-10-30
[VAULTS] Fix/no-fee-revamp-2

folkyatina:
- distribute and disable NO fees on disconnect or transferOwnership
-  charge NO fees over quarantined value also (like Lido fees)


contracts/0.8.25/vaults/LazyOracle.sol
```diff
@@ -196,11 +196,16 @@ contract LazyOracle is ILazyOracle, AccessControlEnumerableUpgradeable {
         return _storage().maxLidoFeeRatePerSecond;
     }
 
+    /// @notice returns the amount of total value that is pending in the quarantine for the given vault
+    function quarantineValue(address _vault) external view returns (uint256) {
+        return _storage().vaultQuarantines[_vault].pendingTotalValueIncrease;
```
dry914: I want to emphasize that the value of the `quarantine.pendingTotalValueIncrease` variable is not always equal to the difference between the total value reported by the oracle and the current total value in the VaultHub.

In particular, if there is an active quarantine already in place, and an additional CL deposit or consolidation occurs, this additional amount will not be immediately included in `pendingTotalValueIncrease`.
It will only be accounted for after the current quarantine is completed.

contracts/0.8.25/vaults/dashboard/Dashboard.sol
TheDZhon: outdated comment

contracts/0.8.25/vaults/dashboard/Dashboard.sol
```diff
@@ -289,7 +304,6 @@ contract Dashboard is NodeOperatorFee {
     function abandonDashboard(address _newOwner) external {
         if (VAULT_HUB.isVaultConnected(address(_stakingVault()))) revert ConnectedToVaultHub();
         if (_newOwner == address(this)) revert DashboardNotAllowed();
-        if (settledGrowth != 0) _setSettledGrowth(0);
```
TheDZhon, line 292: shall we `_stopFeeAccrual` here instead in case of re-entering to the same dashboard?

contracts/0.8.25/vaults/dashboard/Dashboard.sol
```diff
@@ -298,6 +312,7 @@ contract Dashboard is NodeOperatorFee {
     /**
      * @notice Accepts the ownership over the StakingVault and connects to VaultHub. Can be called to reconnect
      *         to the hub after voluntaryDisconnect()
+     * @dev correct the settled growth before reconnecting in order to enable the fee accrual for the node operator
```
TheDZhon: we can probably revert if the fee accrual is stopped and detect via uint104 max value reading

contracts/0.8.25/vaults/dashboard/Dashboard.sol
```diff
@@ -308,6 +323,7 @@ contract Dashboard is NodeOperatorFee {
     /**
      * @notice Connects to VaultHub, transferring underlying StakingVault ownership to VaultHub.
      * @param _currentSettledGrowth The current settled growth value to verify against the stored one
+     * @dev correct the settled growth before reconnecting in order to enable the fee accrual for the node operator
```
TheDZhon: same as above

contracts/0.8.25/vaults/dashboard/NodeOperatorFee.sol
```diff
@@ -222,7 +227,7 @@ contract NodeOperatorFee is Permissions {
         if (latestCorrectionTimestamp >= _lazyOracle().latestReportTimestamp()) revert CorrectionAfterReport();
 
         // If the vault is quarantined, the total value is reduced and may not reflect the exemption
```
TheDZhon: but we are now calculating fees over quarantined amount, can it be relaxed or better stay strict?

contracts/0.8.25/vaults/LazyOracle.sol
TheDZhon: @dry914 good feedback! probably a cleaner solution would be just to have the latest `_reportedTotalValue` stored and have a simple getter, that way we can use it for calculating fees and be 1:1 consistent with Lido fees

cc: @folkyatina

TheDZhon (commented):
👀

contracts/0.8.25/vaults/LazyOracle.sol
folkyatina: But then we just store another value to another slot that in 99% cases are equal to report.totalValue and making bulky reports even more bulkier

contracts/0.8.25/vaults/LazyOracle.sol
```diff
@@ -95,13 +95,15 @@ contract LazyOracle is ILazyOracle, AccessControlEnumerableUpgradeable {
     struct Quarantine {
         uint128 pendingTotalValueIncrease;
         uint64 startTimestamp;
+        uint128 totalValueReminder;
```
failingtwice: *remainder

contracts/0.8.25/vaults/dashboard/Dashboard.sol
folkyatina: Actually, it's complicated. You better set settled growth after the connection to be sure how inOutDelta will look like

contracts/0.8.25/vaults/LazyOracle.sol
```diff
@@ -196,20 +198,35 @@ contract LazyOracle is ILazyOracle, AccessControlEnumerableUpgradeable {
...
     /// @notice returns the quarantine info for the vault
     /// @param _vault the address of the vault
     /// @dev returns zeroed structure if there is no active quarantine
     function vaultQuarantine(address _vault) external view returns (QuarantineInfo memory) {
-        Quarantine storage q = _storage().vaultQuarantines[_vault];
-        if (q.pendingTotalValueIncrease == 0) {
-            return QuarantineInfo(false, 0, 0, 0);
+        Quarantine memory q = _storage().vaultQuarantines[_vault];
+
+        bool isQuarantineActive = q.pendingTotalValueIncrease == 0;
```
dry914: `isQuarantineInactive` or `isQuarantineActive = q.pendingTotalValueIncrease != 0` ?

contracts/0.8.25/vaults/LazyOracle.sol
```diff
@@ -196,20 +198,35 @@ contract LazyOracle is ILazyOracle, AccessControlEnumerableUpgradeable {
         return _storage().maxLidoFeeRatePerSecond;
     }
 
+    /// @notice returns the amount of total value that is pending in the quarantine for the given vault
+    function quarantineValue(address _vault) external view returns (uint256) {
+        Quarantine memory q = _storage().vaultQuarantines[_vault];
+        uint256 pendingValue = q.pendingTotalValueIncrease;
+        if (pendingValue > 0) {
+            // saving one SLOAD if pendingValue
```
dry914:
```suggestion
            // saving one SLOAD if pendingValue is zero
```

contracts/0.8.25/vaults/LazyOracle.sol
```diff
@@ -581,25 +600,35 @@ contract LazyOracle is ILazyOracle, AccessControlEnumerableUpgradeable {
...
 
     function _startNewQuarantine(
-        address _vault,
         Quarantine storage _quarantine,
         uint256 _amountToQuarantine,
         uint256 _currentTimestamp
     ) internal {
         _quarantine.pendingTotalValueIncrease = uint128(_amountToQuarantine);
         _quarantine.startTimestamp = uint64(_currentTimestamp);
-        emit QuarantineActivated(_vault, _amountToQuarantine);
```
dry914, line 591: Why was the event moved out of internal function?
Now it’s in two places, and its argument gets recalculated twice each time.

For example:

> _startNewQuarantine(quarantine, **totalValueIncrease - quarantinedValue**, _reportTimestamp);
> emit QuarantineActivated(_vault, **totalValueIncrease - quarantinedValue**);

contracts/0.8.25/vaults/LazyOracle.sol
```diff
@@ -544,6 +557,9 @@ contract LazyOracle is ILazyOracle, AccessControlEnumerableUpgradeable {
                 return _reportedTotalValue;
             } else {
                 // Transition: QUARANTINE_ACTIVE → QUARANTINE_ACTIVE (maintain quarantine)
+                uint256 reminder = _reportedTotalValue - (onchainTotalValueOnRefSlot + quarantinedValue);
```
dry914: I suspect there might be an unintended revert due to underflow in this case:

1. On the first report:
    onChainTotalValue = 10 ETH, reportedTotalValue = 100 ETH, quarantine = 90 ETH
2. On the second report (e.g., after a slashing event):
    onChainTotalValue = 10 ETH, reportedTotalValue = 99.999 ETH, quarantine = 90 ETH
    → this leads to: 99.999 - (10 + 90)
    → underflow and revert.

contracts/0.8.25/vaults/LazyOracle.sol
folkyatina: It's insignificant spending of gas, but it looks more readable when each branch of the quarantine state machine has its own event in the end.

contracts/0.8.25/vaults/LazyOracle.sol
```diff
@@ -95,13 +95,15 @@ contract LazyOracle is ILazyOracle, AccessControlEnumerableUpgradeable {
     struct Quarantine {
         uint128 pendingTotalValueIncrease;
         uint64 startTimestamp;
+        uint128 totalValueRemainder;
```
TheDZhon, line 98: could have been a single slot though, but can disregard at this stage

```solidity
Quarantine {
        uint104 pendingTotalValueIncrease;
        uint104 totalValueRemainder;
        uint48 startTimestamp;
}
```

contracts/0.8.25/vaults/LazyOracle.sol
```diff
@@ -95,13 +95,15 @@ contract LazyOracle is ILazyOracle, AccessControlEnumerableUpgradeable {
...
     }
 
     struct QuarantineInfo {
         bool isActive;
         uint256 pendingTotalValueIncrease;
         uint256 startTimestamp;
         uint256 endTimestamp;
+        uint256 totalValueRemainder;
```
TheDZhon, line 106: weird placing at the end of the struct tbh

TheDZhon (approved):
GG

contracts/0.8.25/vaults/LazyOracle.sol
folkyatina: It's a trade-off with breaking the storage on the testnet. BTW, we don't really use it in 99% of cases though.

PR #1545, opened 2025-11-04
feat: remove transfer to vault

tamtamchik:
An addition that is not intended to be included in the scope of V3. 
Details https://research.lido.fi/t/post-mortem-nor-curated-module-sdvt-recovery-lever-weakness/10282


TheDZhon (approved):
LGTM (should be delivered later with SR 3.0 and CM 2.0)

PR #1546, opened 2025-11-05
[VAULT] extend a batch getter in LazyOracle

folkyatina:
batchValidatorStatus instead of batchValidatorStages


TheDZhon (approved):
GG, though what is weird is that we don't change any test with this code update 🌚

PR #1547, opened 2025-11-06
feat: allow update fees on unhealthy vault

tamtamchik:
Allow fee updates for unhealthy vaults


contracts/0.8.25/vaults/VaultHub.sol
TheDZhon: this comment is no longer holds completely

TheDZhon: moreover I feel it should say that it reverts if the vault's minting capacity will be exceeded with new parameters which is:
- not about healthiness
- not about _current_ params

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -453,16 +453,22 @@ contract VaultHub is PausableUntilWithRoles {
...
+        _requireConnected(connection, _vault);
 
+        VaultRecord storage record = _vaultRecord(_vault);
         _requireFreshReport(_vault, record);
 
-        uint256 totalValue_ = _totalValue(record);
-        uint256 liabilityShares_ = record.liabilityShares;
+        if (
+            _reserveRatioBP != connection.reserveRatioBP || 
+            _forcedRebalanceThresholdBP != connection.forcedRebalanceThresholdBP
```
TheDZhon, line 464: not sure about frt condition here bc it doesn't participate in the nested calc

TheDZhon (commented):
👀

contracts/0.8.25/vaults/VaultHub.sol
folkyatina: if we don't check it on frt change, you will be able to move healthy vault to the other tier with different frt but same rr and make it unhealthy.

TheDZhon: agree

contracts/0.8.25/vaults/VaultHub.sol
tamtamchik: Fixed in 692bd37

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -432,14 +432,14 @@ contract VaultHub is PausableUntilWithRoles {
...
     /// @notice updates the vault's connection parameters
-    /// @dev Reverts if the vault is not healthy as of latest report
     /// @param _vault vault address
     /// @param _shareLimit new share limit
     /// @param _reserveRatioBP new reserve ratio
     /// @param _forcedRebalanceThresholdBP new forced rebalance threshold
     /// @param _infraFeeBP new infra fee
     /// @param _liquidityFeeBP new liquidity fee
     /// @param _reservationFeeBP new reservation fee
+    /// @dev reverts if the vault's minting capacity will be exceeded with new parameters
```
```suggestion
    /// @dev reverts if the vault's minting capacity will be exceeded with new reserve parameters
```

TheDZhon (approved):
LGTM

PR #1548, opened 2025-11-06
[VAULT]: Shortfall strikes back 

folkyatina:
https://github.com/lidofinance/core/issues/1305#issuecomment-3496664405

### DIFF
--- a/contracts/0.8.25/vaults/VaultHub.sol
+++ b/contracts/0.8.25/vaults/VaultHub.sol
@@ -1218,9 +1218,6 @@ contract VaultHub is PausableUntilWithRoles {
             return 0;
         }
 
-        // if not healthy and low in debt, please rebalance the whole amount
-        if (liabilityShares_ <= 100) return liabilityShares_;
-
         uint256 reserveRatioBP = _connection.reserveRatioBP;
         uint256 maxMintableRatio = (TOTAL_BASIS_POINTS - reserveRatioBP);
         uint256 liability = _getPooledEthBySharesRoundUp(liabilityShares_);
@@ -1229,6 +1226,9 @@ contract VaultHub is PausableUntilWithRoles {
         if (liability > totalValue_) {
             return type(uint256).max;
         }
+
+        // if not healthy and low in debt, please rebalance the whole amount
+        if (liabilityShares_ <= 100) return liabilityShares_;
 
         // Solve the equation for X:
         // L - liability, TV - totalValue
--- a/contracts/0.8.25/vaults/VaultHub.sol
+++ b/contracts/0.8.25/vaults/VaultHub.sol
@@ -1218,6 +1218,9 @@ contract VaultHub is PausableUntilWithRoles {
             return 0;
         }
 
+        // if not healthy and low in debt, please rebalance the whole amount
+        if (liabilityShares_ <= 100) return liabilityShares_;
+
         uint256 reserveRatioBP = _connection.reserveRatioBP;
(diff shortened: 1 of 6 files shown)


contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -1242,10 +1242,11 @@ contract VaultHub is PausableUntilWithRoles {
         // X = (L * 100 - TV * MR) / (100 - MR)
         // RR = 100 - MR
         // X = (L * 100 - TV * MR) / RR
-        uint256 shortfallEth = (liability * TOTAL_BASIS_POINTS - totalValue_ * maxMintableRatio) / reserveRatioBP;
+        uint256 shortfallEth = Math256.ceilDiv(liability * TOTAL_BASIS_POINTS - totalValue_ * maxMintableRatio,
+            reserveRatioBP);
 
         // Add 10 extra shares to avoid dealing with rounding/precision issues
```
```suggestion
        // Add 100 extra shares to avoid dealing with rounding/precision issues
```

TheDZhon (approved):
GG 🚶‍♂️

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -1242,10 +1242,11 @@ contract VaultHub is PausableUntilWithRoles {
         // X = (L * 100 - TV * MR) / (100 - MR)
         // RR = 100 - MR
         // X = (L * 100 - TV * MR) / RR
-        uint256 shortfallEth = (liability * TOTAL_BASIS_POINTS - totalValue_ * maxMintableRatio) / reserveRatioBP;
+        uint256 shortfallEth = Math256.ceilDiv(liability * TOTAL_BASIS_POINTS - totalValue_ * maxMintableRatio,
```
failingtwice, line 1248: Ceil div gives additional 1 wei which is not very important, since we add 100 safety buffer anyway

contracts/0.8.25/vaults/VaultHub.sol
failingtwice:
```suggestion
        // Add 100 extra shares to avoid dealing with rounding/precision issues
```

contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -1242,10 +1242,11 @@ contract VaultHub is PausableUntilWithRoles {
...
+            reserveRatioBP);
 
         // Add 10 extra shares to avoid dealing with rounding/precision issues
-        uint256 shortfallShares = _getSharesByPooledEth(shortfallEth) + 10;
+        uint256 shortfallShares = _getSharesByPooledEth(shortfallEth) + 100;
```
failingtwice, line 1252: I have an intuition that 100 will not be enough either. Something tells me that issue arises when `liabilityShares` and `totalValue` are less than `TOTAL_BASIS_POINTS`. But I don't have hard arguments to support that.

contracts/0.8.25/vaults/VaultHub.sol
TheDZhon: it's acceptable imo

TheDZhon (approved):
LGTM, if we want to get harder though, the following might be done:

- early return vault.liabilityShares for every unhealthy vault with liabilityShares <= 100
- force wider (FRT - RR) gap (like 10BP)
- while loop 😶‍🌫️

TheDZhon (approved):
🔫

TheDZhon:
cc: @dry914 pls forgive us :(

TheDZhon:
image

PR #1551, opened 2025-11-10
feat: alerting harness contract

tamtamchik:


contracts/harness/AlertingHarness.sol
```diff
@@ -0,0 +1,308 @@
...
+pragma solidity 0.8.25;
+
+import {ILidoLocator} from "contracts/common/interfaces/ILidoLocator.sol";
+import {IStakingVault} from "contracts/0.8.25/vaults/interfaces/IStakingVault.sol";
+
+import {VaultHub} from "contracts/0.8.25/vaults/VaultHub.sol";
+import {LazyOracle} from "contracts/0.8.25/vaults/LazyOracle.sol";
+import {PredepositGuarantee} from "contracts/0.8.25/vaults/predeposit_guarantee/PredepositGuarantee.sol";
+
+
```
TheDZhon: Should provide a clear disclaimer that the contract is not a core protocol part and tooling only

contracts/harness/AlertingHarness.sol
```diff
@@ -0,0 +1,308 @@
...
+/// @title AlertingHarness
+/// @notice provides a harness for gathering vault-related data
+contract AlertingHarness {
+    /// @notice reference to the Lido locator contract used to resolve protocol contract addresses
+    ILidoLocator public immutable LIDO_LOCATOR;
+
+    /// @notice version of the contract
+    string public constant VERSION = "1.0.0";
```
TheDZhon: why do we need this though?

contracts/harness/AlertingHarness.sol
```diff
@@ -0,0 +1,308 @@
...
+    function _lazyOracle() internal view returns (LazyOracle) {
+        return LazyOracle(LIDO_LOCATOR.lazyOracle());
+    }
+
+    /// @notice helper to resolve the current PredepositGuarantee contract from the locator
+    /// @return contract instance of PredepositGuarantee
+    function _predepositGuarantee() internal view returns (PredepositGuarantee) {
+        return PredepositGuarantee(LIDO_LOCATOR.predepositGuarantee());
+    }
+}
```
TheDZhon: no new line at EOF 🌔

contracts/harness/AlertingHarness.sol
```diff
@@ -0,0 +1,308 @@
...
+/// @notice provides a harness for gathering vault-related data
+contract AlertingHarness {
+    /// @notice reference to the Lido locator contract used to resolve protocol contract addresses
+    ILidoLocator public immutable LIDO_LOCATOR;
+
+    /// @notice version of the contract
+    string public constant VERSION = "1.0.0";
+
+    /// @notice structure containing relevant data for a single staking vault
+    /// @param stakingVault the address of the staking vault
```
TheDZhon: The structure doesn't have this field

contracts/harness/AlertingHarness.sol
```diff
@@ -0,0 +1,308 @@
...
+    /// @param stakingVaultStagedBalance the staged balance of the staking vault
+    /// @param stakingVaultAvailableBalance the available balance of the staking vault
+    struct StakingVaultData {
+        address stakingVaultNodeOperator;
+        address stakingVaultDepositor;
+        address stakingVaultOwner;
+        address stakingVaultPendingOwner;
+        uint256 stakingVaultStagedBalance;
+        uint256 stakingVaultAvailableBalance;
+        bool stakingVaultBeaconChainDepositsPaused;
```
TheDZhon: missing natspec

contracts/tooling/AlertingHarness.sol
```diff
@@ -0,0 +1,308 @@
...
+        uint256 _offset,
+        uint256 _limit
+    ) external view returns (VaultConnectionData[] memory batch) {
+        (VaultHub vaultHub, uint256 batchSize) = _getBatchSize(_offset, _limit);
+
+        if (batchSize == 0) return new VaultConnectionData[](0);
+
+        batch = new VaultConnectionData[](batchSize);
+        for (uint256 i = 0; i < batchSize; ++i) {
+            address vault = vaultHub.vaultByIndex(_offset + i + 1);
```
TheDZhon, line 148: having two conventions for indexing might be a bit misleading (outer is 0-based, inner is 1-based)

contracts/harness/AlertingHarness.sol
```diff
@@ -0,0 +1,308 @@
...
+    /// @return vaultHub the VaultHub contract instance
+    /// @return batchSize actual batch size to use (0 if offset out of range)
+    function _getBatchSize(
+        uint256 _offset,
+        uint256 _limit
+    ) private view returns (VaultHub vaultHub, uint256 batchSize) {
+        vaultHub = _vaultHub();
+        uint256 vaultsCount = vaultHub.vaultsCount();
+
+        if (_offset > vaultsCount) return (vaultHub, 0);
```
TheDZhon:
```suggestion
        if (_offset >= vaultsCount) return (vaultHub, 0);
```

contracts/harness/AlertingHarness.sol
```diff
@@ -0,0 +1,308 @@
...
+        address vault;
+        LazyOracle.QuarantineInfo vaultQuarantineInfo;
+    }
+
+    struct VaultPendingActivationsData {
+        address vault;
+        uint256 vaultPendingActivationsCount;
+    }
+
+    struct VaultStakingVaultData {
```
TheDZhon: Vault Vault?

TheDZhon (commented):
👀

contracts/harness/AlertingHarness.sol
tamtamchik: we don't, had an idea of adding some checks for a version inside the alerting code, but can be omitted for now

contracts/tooling/AlertingHarness.sol
tamtamchik: Works exactly as in VaultHub, so all outer clients may use single way [0, vaultsCount)

contracts/harness/AlertingHarness.sol
tamtamchik: Tried to come up with a better naming

contracts/tooling/AlertingHarness.sol
failingtwice: To avoid confusion, consider not overriding the offset index and leave the responsibility for providing correct indexes on the user. I think it's okay that the user can set 0 as offset and get the dummy vault entry at index 0.

contracts/tooling/AlertingHarness.sol
```diff
@@ -0,0 +1,310 @@
...
+import {ILidoLocator} from "contracts/common/interfaces/ILidoLocator.sol";
+import {IStakingVault} from "contracts/0.8.25/vaults/interfaces/IStakingVault.sol";
+
+import {VaultHub} from "contracts/0.8.25/vaults/VaultHub.sol";
+import {LazyOracle} from "contracts/0.8.25/vaults/LazyOracle.sol";
+import {PredepositGuarantee} from "contracts/0.8.25/vaults/predeposit_guarantee/PredepositGuarantee.sol";
+
+/// @title AlertingHarness
+/// @dev this contract is NOT a part of the Lido core protocol logic, it is only used for tooling purposes
+contract AlertingHarness {
```
failingtwice, line 22: Might be worth adding `vaultsCount()` function to avoid querying it directly on VaultHub

failingtwice (dismissed):
minor non-blocking nitpicks

contracts/tooling/AlertingHarness.sol
tamtamchik: Added in 0947187

TheDZhon (approved):
LGTM

PR #1557, opened 2025-11-11
[VAULTS]: Audit fixes 11

failingtwice:
Users (especially critical roles like default admin or node operator manager) can accidentally renounce their role.

## Solution

Override the method and always revert


TheDZhon (approved):
LGTM

PR #1580, opened 2025-12-01
feat(orc-537): Update hash consesnsus contract to prevent attack on _…

chasingrainbows:
Reworked HashConsensus reporting to store one hash per member (_memberReports) plus per-hash support (_hashSupport), capping per-frame state and preventing unbounded variantsLength.

## Problem

HashConsensus previously kept every distinct report in _reportVariants, so a malicious oracle could spam unique hashes, forcing unbounded iteration and potential OOG. We replaced that structure with per-member/per-hash mappings while preserving the historical interface (getReportVariants(), consensus events, fast-lane rules).

## Solution

Replace the unbounded variant list with:
1. _memberReports to track each member’s current hash;
2. _hashSupport to count supporters per hash;
3. _currentFrameReporters for bounded cleanup between frames.


F4ever:
Possible solution
https://github.com/lidofinance/lido-improvement-proposals/pull/121#discussion_r3389408902

PR #1582, opened 2025-12-02
Correct Mainnet deploy FRT for default tier to 49.5%

failingtwice:


TheDZhon:
Seems we are not doing this anymore *close_gates*

PR #1589, opened 2025-12-05
fix: update voting impl address

failingtwice:
Update the address of Voting implementation contract.

https://docs.lido.fi/deployed-contracts/#dao-contracts


tamtamchik (approved):
🫶

PR #1602, opened 2025-12-09
feat: redeploy V3VoteScript with PDG pause on start

failingtwice:
Redeploy V3VoteScript to include PDG pause


tamtamchik (approved):
LGTM

PR #1612, opened 2025-12-14
Attempt to add integration tests which simulate graph calculations

arwer13:


test/integration/core/total-reward.integration.ts
tamtamchik, line 1: Must be moved to integration tests folder?

PR #1615, opened 2025-12-15
Fix smoothing trigger calculation

avsetsin:
The PR improves the calculation logic of required bad debt to internalize to trigger smoothen token rebase in integration tests


folkyatina (approved):
Looks good

PR #1600, opened 2025-12-09
fix: bls bytes verification

Jeday:
## Context
magical-magic

## Problem

Restores invariant where on-chain BLS check must be complete to CL BLS check

## Solution

Verify compression flag bits in pubkey and signature againts user provided Y components

PR #1620, opened 2025-12-23
Fix/bls bytes verification hardening

TheDZhon:
## Context

#1600

## Problem

The `fix/bls-bytes-verification` branch introduced BLS signature verification but lacked:
- Length validation for signatures (only pubkey length was checked)
- Comprehensive test coverage to ensure parity with Consensus Layer (CL) verification
- Some comments had incorrect precompile addresses and typos

## Solution

1. **Added missing input validation**:
   - Added `InvalidSignatureLength` error
   - Added signature length check (96 bytes) in `validateCompressedSignatureFlags()`
   - Added pubkey length check (48 bytes) in `validateCompressedPubkeyFlags()`

2. **Fixed comment typos**:
   - Corrected precompile address comments (`0x0a` → `0x11` for MAP_FP2_TO_G2, `0x0f` → `0x0d` for G2ADD)
   - Fixed signature G2 X component comments (`X_c1`/`X_c2` → `X_c0`/`X_c1`)
   - Fixed typo `staticall` → `staticcall`

3. **Added comprehensive test coverage**:
   - `BLSFuzzHarness.sol` - Harness contract for E2E testing
   - `bls.blst.e2e.fuzz.test.ts` - E2E fuzz tests using `@chainsafe/blst` as oracle to ensure accept/reject parity between Solidity and the reference BLS library
   - Extended `BLS.t.sol` with Foundry fuzz tests covering:
     - Compression flag edge cases (invalid compression/infinity/sign bits)
     - Y-coordinate boundaries (zero, P/2, P-1, just above P/2)
     - Sign bit parity invariants
     - Invalid length handling
     - Subgroup validation (non-subgroup G1/G2 points)
     - Mutation-based fuzzing


test/common/contracts/BLS12_381__Harness.sol
tamtamchik, line 1: Move to tests as we not going to deploy it on mainnet?

contracts/common/lib/BLS.sol
tamtamchik:
```suggestion
            // Load signatureG2.X_c1 skipping 16 bytes of zero padding
```

tamtamchik:
```suggestion
            // Load signatureG2.X_c0 first 32 bytes and apply sign mask
```

tamtamchik: And so on?

contracts/common/lib/BLS.sol
```diff
@@ -246,7 +249,7 @@ library BLS12_381 {
             mapToG2(b, result) // result at offset 0
             mapToG2(add(0x80, b), add(0x100, result)) // second point at result + 0x100
 
-            // Add the two G2 points together with BLS12_G2ADD precompile (0x0f)
+            // Add the two G2 points together with BLS12_G2ADD precompile (0x0d)
```
tamtamchik: Magic number?

contracts/common/lib/BLS.sol
```diff
@@ -182,7 +185,7 @@ library BLS12_381 {
                 }
             }
 
-            /// @dev Map an fp2 field element to a point in G2 curve using BLS12 precompile (0x0a)
+            /// @dev Map an fp2 field element to a point in G2 curve using BLS12 precompile (0x11)
```
tamtamchik: Magic number?

contracts/common/lib/BLS.sol
TheDZhon: Precompile addr https://eips.ethereum.org/EIPS/eip-2537

image

contracts/common/lib/BLS.sol
TheDZhon: Precompile addr https://eips.ethereum.org/EIPS/eip-2537

image

test/common/contracts/BLS12_381__Harness.sol
TheDZhon: fair, will do :pray:

contracts/common/lib/BLS.sol
tamtamchik: Yeah, my point was to add the link as comment )

contracts/common/lib/BLS.sol
tamtamchik: Yeah, my point was to add the link as comment )

test/common/bls.blst.e2e.fuzz.test.ts
```diff
@@ -0,0 +1,700 @@
...
+  const master = SecretKey.deriveMasterEip2333(getBytes(keccak256(Buffer.from(IKM_SEED, "utf-8"))));
+
+  let harness: BLS12_381__Harness;
+
+  before(async () => {
+    harness = (await ethers.deployContract("BLS12_381__Harness")) as unknown as BLS12_381__Harness;
+  });
+
+  it("matches consensus spec roots and accepts exactly the signatures that blst accepts (valid deposits)", async () => {
+    for (let i = 0; i < RUNS; i++) {
```
dry914, line 194: Currently, the tests always use the same private keys (indices 1–50, 10,000+, etc.) when `BLS_BLST_FUZZ_SEED` is not set, which limits fuzzing coverage.

I think it would be better to generate a deterministic random starting index for each test (e.g., from `keccak256(random number + testName)`) and then derive the next N keys sequentially. This would ensure different keys on each run while maintaining reproducibility via logged starting indices.

test/common/bls.blst.e2e.fuzz.test.ts
```diff
@@ -0,0 +1,700 @@
...
+      // Deterministic forkVersion and withdrawal credentials from i.
+      const salt = keccak256(toBeHex(i, 32));
+      const forkVersion = `0x${salt.slice(2, 10)}`; // bytes4
+      const withdrawalCredentials = keccak256(`0x${salt.slice(2, 2 + 64)}`); // bytes32
+
+      // Amount in wei, must be a multiple of 1 gwei.
+      const gweiAmount = 1_000_000_000n + (BigInt(i) % (32_000_000_000n - 1_000_000_000n + 1n)); // [1 ETH..32 ETH] in gwei
```
dry914: The value of this parameter in the current version of the code is very close to 1 ETH in all runs, so the index `i` always starts from 0.

I recommend using random starting index for each test (as described above).

test/common/bls.blst.e2e.fuzz.test.ts
```diff
@@ -0,0 +1,700 @@
...
+          expectError: { name: "InvalidCompressedComponent", args: [1] },
+        },
+        {
+          name: "signature: flip sign bit",
+          mutate: () => ({ signature: flipFirstByte(signature, 0x20) }),
+          expectError: { name: "InvalidCompressedComponentSignBit", args: [1] },
+        },
+        {
+          name: "pubkey: flip a non-header bit (forces point validation / pairing failure)",
+          mutate: () => ({ pubkey: flipByte(pubkey, 1, 0x01) }),
```
dry914, line 310: I think coverage would increase if we used a random bit here instead of the hardcoded `0x01` - the same applies to all cases where `0x01` is currently hardcoded.

test/common/bls.blst.e2e.fuzz.test.ts
```diff
@@ -0,0 +1,700 @@
...
+    const sigSk = master.deriveChildEip2333(777_777);
+    const sig = sigSk.sign(getBytes(keccak256(Buffer.from("non-subgroup-test", "utf-8"))));
+    const signature = sig.toHex(true);
+
+    const depositY = buildDepositY(pk, sig);
+
+    // Arbitrary message parameters; signature is not expected to match.
+    const forkVersion = "0x00000000";
+    const withdrawalCredentials = keccak256(toBeHex(424242, 32));
+    const amount = 1_000_000_000n * ONE_GWEI; // 1 ETH in gwei, in wei.
```
dry914:
```suggestion
    const amount = 1_000_000_000n * ONE_GWEI; // 1 ETH in wei.
```

test/common/bls.blst.e2e.fuzz.test.ts
dry914: The same applies to all tests where index `i` starts from `0`.

test/common/bls.blst.e2e.fuzz.test.ts
TheDZhon: Thanks, updated!

test/common/bls.blst.e2e.fuzz.test.ts
TheDZhon: Good point, fixed!

test/common/bls.blst.e2e.fuzz.test.ts
TheDZhon: fixed, thanks!

PR #1634, opened 2026-01-16
Test soft-launch validator on new PDG

failingtwice:
This integration test verifies that the specific validator deposited during Lido V3 soft launch phase can be prove and topped up with new PDG.

Steps:
- use mainnet fork
- load current soft launch state
- upgrade PDG and unpause
- prove, activate and top up the specific validator with specific WC


tamtamchik (approved):
✅

PR #1643, opened 2026-01-26
VEBO(srv3): Rework sanity checks number of requests -> balance

hweawer:
Rework sanity checks to use balance of the validators instead of number of keys in VEBO.


contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -201,6 +219,21 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
...
+    uint256 public constant DATA_FORMAT_LIST_WITH_KEY_INDEX = 2;
+
+    /// @notice Module ID for the curated staking module (NodeOperatorsRegistry)
+    /// @dev This module was deployed before the MaxEB (EIP-7251) fork and uses 32 ETH validators.
+    ///      This is a protocol constant that cannot change for existing validators.
+    uint256 public constant CURATED_MODULE_ID = 1;
+
+    /// @notice Max effective balance for curated module validators (in Gwei)
+    /// @dev 32 ETH = 32_000_000_000 Gwei (pre-MaxEB validators)
+    uint256 public constant CURATED_MODULE_MAX_BALANCE_GWEI = 32_000_000_000;
```
mkurayan: It is preferable not to hardcode such a constant, as the protocol may need to be deployed on custom networks with different configuration parameters.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -201,6 +219,21 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
...
+
+    /// @notice Max effective balance for MaxEB-enabled module validators (in Gwei)
+    /// @dev 2048 ETH = 2_048_000_000_000 Gwei (post-MaxEB validators: CSM, SimpleDVT, etc.)
+    ///      All modules deployed after EIP-7251 activation use this max effective balance.
+    uint256 public constant MAXEB_MODULE_MAX_BALANCE_GWEI = 2_048_000_000_000;
```
mkurayan: It is preferable not to hardcode such a constant, as the protocol may need to be deployed on custom networks with different configuration parameters.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -201,6 +219,21 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
     /// key: (moduleId, nodeOpId, validatorIndex).
     ///
     uint256 public constant DATA_FORMAT_LIST = 1;
+    uint256 public constant DATA_FORMAT_LIST_WITH_KEY_INDEX = 2;
+
+    /// @notice Module ID for the curated staking module (NodeOperatorsRegistry)
+    /// @dev This module was deployed before the MaxEB (EIP-7251) fork and uses 32 ETH validators.
+    ///      This is a protocol constant that cannot change for existing validators.
+    uint256 public constant CURATED_MODULE_ID = 1;
```
mkurayan: Instead of hardcoding the module ID to determine the MaxEB, query the module type via the Staking Router. Specifically, call `getStakingModule` and inspect the `withdrawalCredentialsType` field (0x01 or 0x02) for the keys associated with the given module.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -281,7 +317,7 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
 
         _consumeLimit(requestsCount);
```
mkurayan: Lets switch to ETH limit as we discussed.

Here we can trust to sources which brings exit request, and just assume that that keys belong to the modules.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: Done

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: Done

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: I moved constants to the `WithdrawableCredentials.sol` for now. If we want to go further and configure this values on different networks what contract do you think should store it?

contracts/common/lib/WithdrawalCredentials.sol
```diff
@@ -10,6 +10,20 @@ pragma solidity >=0.8.9 <0.9.0;
...
 library WithdrawalCredentials {
+    /// @notice Withdrawal credentials type for legacy modules (0x01 - execution layer)
+    uint8 public constant WC_TYPE_01 = 0x01;
+
+    /// @notice Withdrawal credentials type for compounding modules (0x02 - compounding)
+    uint8 public constant WC_TYPE_02 = 0x02;
+
+    /// @notice Max effective balance for legacy validators with 0x01 withdrawal credentials
+    /// @dev 32 ETH expressed in Gwei (pre-MaxEB/EIP-7251)
+    uint256 public constant MAX_EFFECTIVE_BALANCE_WC_TYPE_01_GWEI = 32_000_000_000;
```
mkurayan: Please do not hardcode MaxEB values, it will not be possible to run protocol on custom network with different MaxEB values.

contracts/0.8.9/lib/ExitLimitUtils.sol
```diff
@@ -10,6 +10,16 @@ struct ExitRequestLimitData {
     uint32 exitsPerFrame; // Restored exits per frame
 }
 
+/// @notice Gwei-based limit data structure for exit requests
+/// @dev Uses uint64 fields to support large Gwei values (max 18.4 million ETH per field)
+struct ExitBalanceLimitData {
```
mkurayan: You probably do not need a new `ExitBalanceLimit`. Instead, you can express the limit in ETH rather than Gwei and reuse `contracts/common/lib/RateLimit.sol`, which fits semantically.

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -145,7 +147,7 @@ struct LimitsListPacked {
     uint16 appearedValidatorsPerDayLimit;
     uint16 annualBalanceIncreaseBPLimit;
     uint16 simulatedShareRateDeviationBPLimit;
-    uint16 maxValidatorExitRequestsPerReport;
+    uint64 maxBalanceExitRequestedPerReportInGwei;
```
mkurayan: Let’s express the limit in ETH instead of Gwei. A `uint16` allows storing up to 65,535 ETH, which should be sufficient. VEBO can submit up to 3 reports per day, which is well above the current Ethereum daily exit cap (256 ETH per epoch ≈ 57,600 ETH per day).

Keeping `uint16` also avoids changes to the storage layout and eliminates the need for a complex migration.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -22,10 +23,42 @@ interface ITriggerableWithdrawalsGateway {
     ) external payable;
 }
 
+interface INodeOperatorsRegistry {
+    function getNodeOperatorsCount() external view returns (uint256);
```
mkurayan: Looks like you do not use getNodeOperatorsCount method. Can be safely removed.

hweawer: Removed

contracts/common/lib/WithdrawalCredentials.sol
hweawer: Moved to VEBO

contracts/0.8.9/lib/ExitLimitUtils.sol
hweawer: Done

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
hweawer: Done

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -955,7 +962,7 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
     event AnnualBalanceIncreaseBPLimitSet(uint256 annualBalanceIncreaseBPLimit);
     event SimulatedShareRateDeviationBPLimitSet(uint256 simulatedShareRateDeviationBPLimit);
     event MaxPositiveTokenRebaseSet(uint256 maxPositiveTokenRebase);
-    event MaxValidatorExitRequestsPerReportSet(uint256 maxValidatorExitRequestsPerReport);
+    event MaxBalanceExitRequestedPerReportInEthSet(uint16 maxBalanceExitRequestedPerReportInEth);
```
mkurayan: I suggest using `uint256` for events, consistent with the other limits.

In my opinion, from a logging perspective, narrowing the type to `uint16` provides no benefits.

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -976,7 +983,7 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
     error IncorrectSharesRequestedToBurn(uint256 actualSharesToBurn);
     error IncorrectCLBalanceIncrease(uint256 annualBalanceDiff);
     error IncorrectAppearedValidators(uint256 appearedValidatorsLimit);
-    error IncorrectNumberOfExitRequestsPerReport(uint256 maxRequestsCount);
+    error IncorrectSumOfExitBalancePerReport(uint256 maxRequestsCount);
```
mkurayan: I suggest to rename the argument `maxRequestsCount` to balance related name.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -225,8 +275,19 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
         assert(address(this).balance == balanceBeforeCall);
     }
 
-    constructor(address lidoLocator) {
+    INodeOperatorsRegistry public immutable NODE_OPERATORS_REGISTRY;
```
mkurayan: All 0x01 modules should be checked, not only NodeOperatorRegistry.

Instead of passing the NodeOperatorRegistry address to the ValidatorExitBus constructor, a list of modules that support the legacy getSigningKey interface might be provided. I suggest to use simple bit mask for that.

All other modules are treated as new-type modules supporting the updated getSigningKey interface.

When processing a signing key, the contract must determine the module it belongs to and call either the legacy or the new interface accordingly.

hweawer: Done

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
hweawer: Done

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
hweawer: Done

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -201,14 +263,33 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
...
+    /// @notice Max effective balance for legacy validators (withdrawal credentials type 0x01) in ETH
+    /// @dev Typically 32 ETH
+    uint16 public immutable MAX_BALANCE_WC_TYPE_01_ETH;
+
+    /// @notice Max effective balance for MaxEB validators (withdrawal credentials type 0x02) in ETH
+    /// @dev Typically 2048 ETH (post-EIP-7251)
+    uint16 public immutable MAX_BALANCE_WC_TYPE_02_ETH;
+
+    /// @dev Conversion factor: 1 ETH = 1_000_000_000 Gwei (1e9)
+    uint256 private constant GWEI_PER_ETH = 1_000_000_000;
```
mkurayan: Looks like unused constant, can we remove it?

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -449,26 +540,29 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
      * @return nodeOpId ID of the node operator.
      * @return moduleId ID of the staking module.
      * @return valIndex Index of the validator.
+     * @return keyIndex Index of the signing key (0 for format 1, actual value for format 2).
      */
     function unpackExitRequest(
```
mkurayan, line 521: Do we need to change unpackExitRequest method signature (return `keyIndex`)? 

Currently looks like only `ValidatorExitDelayVerifier` contract use this  method.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -134,11 +194,11 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
...
      * @notice Emitted when limits configs are set.
-     * @param maxExitRequestsLimit The maximum number of exit requests.
-     * @param exitsPerFrame The number of exits that can be restored per frame.
-     * @param frameDurationInSec The duration of each frame, in seconds, after which `exitsPerFrame` exits can be restored.
+     * @param maxExitBalanceEth The maximum exit balance limit in ETH.
+     * @param balancePerFrameEth The exit balance in ETH that can be restored per frame.
+     * @param frameDurationInSec The duration of each frame, in seconds, after which `balancePerFrameEth` can be restored.
      */
-    event ExitRequestsLimitSet(uint256 maxExitRequestsLimit, uint256 exitsPerFrame, uint256 frameDurationInSec);
+    event ExitRequestsLimitSet(uint256 maxExitBalanceEth, uint256 balancePerFrameEth, uint256 frameDurationInSec);
```
mkurayan: Maybe it is better to rename this event in order to avoid confusion with existed events which contains number of validator exits instead of balances.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -663,6 +790,7 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
         validatorData.valIndex = uint64(dataWithoutPubkey);
         validatorData.nodeOpId = uint40(dataWithoutPubkey >> 64);
         validatorData.moduleId = uint24(dataWithoutPubkey >> (64 + 40));
+        validatorData.keyIndex = 0; // Format 1 always uses keyIndex 0
```
mkurayan: Could you please double check, maybe it is better to use uint256 max value, in order to avoid confusion if `0` is a valid key index?

scripts/scratch/steps/0083-deploy-core.ts
```diff
@@ -287,7 +288,7 @@ export async function main() {
     "ValidatorsExitBusOracle",
     proxyContractsOwner,
     deployer,
-    [chainSpec.secondsPerSlot, chainSpec.genesisTime, locator.address],
+    [chainSpec.secondsPerSlot, chainSpec.genesisTime, locator.address, nodeOperatorsRegistryAddress],
```
mkurayan: Please update scratch deploy

contracts/0.8.9/oracle/ValidatorsExitBusOracle.sol
```diff
@@ -77,22 +80,22 @@ contract ValidatorsExitBusOracle is BaseOracle, ValidatorsExitBus {
      */
     function finalizeUpgrade_v2(
```
mkurayan: You can remove current upgrade and create a new one v3, please update the contract versions during the finalizeUpgarde_v3

contracts/0.8.9/oracle/ValidatorsExitBusOracle.sol
```diff
@@ -77,22 +80,22 @@ contract ValidatorsExitBusOracle is BaseOracle, ValidatorsExitBus {
...
-        uint256 exitsPerFrame,
+        uint256 maxExitBalanceEth,
+        uint256 balancePerFrameEth,
         uint256 frameDurationInSec
     ) external {
-        _initialize_v2(maxValidatorsPerReport, maxExitRequestsLimit, exitsPerFrame, frameDurationInSec);
+        _initialize_v2(maxValidatorsPerReport, maxExitBalanceEth, balancePerFrameEth, frameDurationInSec);
     }
 
     function _initialize_v2(
```
mkurayan: Should be v3

contracts/0.8.9/oracle/ValidatorsExitBusOracle.sol
```diff
@@ -77,22 +80,22 @@ contract ValidatorsExitBusOracle is BaseOracle, ValidatorsExitBus {
...
         uint256 maxValidatorsPerReport,
-        uint256 maxExitRequestsLimit,
-        uint256 exitsPerFrame,
+        uint256 maxExitBalanceEth,
+        uint256 balancePerFrameEth,
         uint256 frameDurationInSec
     ) internal {
         _updateContractVersion(2);
```
mkurayan: Lets upgrade contract version

hweawer: Done

contracts/0.8.9/oracle/ValidatorsExitBusOracle.sol
hweawer: Done

contracts/0.8.9/oracle/ValidatorsExitBusOracle.sol
hweawer: Done

scripts/scratch/steps/0083-deploy-core.ts
hweawer: Done

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: Done

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: Reverted changes in this function

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: Removed

contracts/0.8.9/oracle/ValidatorsExitBusOracle.sol
```diff
@@ -67,32 +70,32 @@ contract ValidatorsExitBusOracle is BaseOracle, ValidatorsExitBus {
         _pauseFor(PAUSE_INFINITELY);
         _initialize(consensusContract, consensusVersion, lastProcessingRefSlot);
 
-        _initialize_v2(maxValidatorsPerRequest, maxExitRequestsLimit, exitsPerFrame, frameDurationInSec);
+        _initialize_v3(maxValidatorsPerRequest, maxExitBalanceEth, balancePerFrameEth, frameDurationInSec);
     }
 
     /**
      * @notice A function to finalize upgrade to v2 (from v1). Can be called only once
```
mkurayan: Please update the comment as well.
`upgrade to v3 (from v1)`

hweawer: Done

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: Indeed 0 is a valid kapin index, changed to max value

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -676,10 +799,214 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
...
+                    itemOffset := add(data.offset, mul(PACKED_REQUEST_LENGTH_V2, i))
+                    let dataWithoutPubkey := shr(64, calldataload(itemOffset))
+                    moduleId := shr(168, dataWithoutPubkey) // Extract top 24 bits
+                }
+            } else {
+                revert UnsupportedRequestsDataFormat(dataFormat);
+            }
+
+            // Check cache before making external call
+            bool isLegacy;
```
mkurayan: I suggest to rename this variable, because CSM with MAX_BALANCE_WC_TYPE_01_ETH not legacy module. 

this boolean flag not about legacy modules (NOR,SDVT) but about module with 0x01 validators (NOR,SDVT, CSM)

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -676,10 +799,214 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
...
+            if (isLegacy) {
+                totalBalanceEth32 += MAX_BALANCE_WC_TYPE_01_ETH;
+            } else {
+                totalBalanceEth32 += MAX_BALANCE_WC_TYPE_02_ETH;
+            }
+        }
+
+        // Cast to uint16 for return - will truncate if > 65535
+        // The sanity checker will catch if this exceeds the configured limit
+        totalBalanceEth = uint16(totalBalanceEth32);
```
mkurayan: @hweawer I suggest not casting to `uint16` and instead operating on `uint256`. This guarantees that even if the exit request ETH amount exceeds `uint16` max and the limit is set to `uint16` max, the check will still work correctly.

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -552,11 +555,15 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
 
     /// @notice Applies sanity checks to the number of validator exit requests supplied to ValidatorExitBusOracle
-    /// @param _exitRequestsCount Number of validator exit requests supplied per oracle report
-    function checkExitBusOracleReport(uint256 _exitRequestsCount) external view {
-        uint256 limit = _limits.unpack().maxValidatorExitRequestsPerReport;
-        if (_exitRequestsCount > limit) {
-            revert IncorrectNumberOfExitRequestsPerReport(limit);
+    /// @notice Checks the total balance of validator exit requests supplied per oracle report
+    /// @param _maxBalanceExitRequestedPerReportInEth Total balance in ETH of all validators requested to exit in the oracle report
+    function checkExitBusOracleReport(uint16 _maxBalanceExitRequestedPerReportInEth)
```
mkurayan: @hweawer Let’s use `uint256` instead of `uint16`. The return type of `_calculateTotalExitBalanceEth` should also be updated to `uint256`.

This ensures that even if the total ETH amount of exit requests exceeds the `uint16` maximum and the limit is set to `uint16` max, the check will still function correctly.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -634,13 +734,35 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
...
+        } else {
+            revert UnsupportedRequestsDataFormat(dataFormat);
+        }
+    }
+
+    /**
+    * @notice Extracts validator data from format 1 (64 bytes per request, no keyIndex)
+    * @param exitRequestData Validator exit requests data
+    * @param index index of request in array
+    * @return validatorData Validator data with keyIndex = 0
```
mkurayan: NatSpec mismatch, the NatSpec says keyIndex = 0 but in code it 
`type(uint256).max`.

contracts/common/lib/WithdrawalCredentials.sol
```diff
@@ -10,6 +10,12 @@ pragma solidity >=0.8.9 <0.9.0;
  * @dev WC bytes layout: [0] = prefix (0x00/0x01/0x02), [1..11] = zero, [12..31] = execution address (20b)
  */
 library WithdrawalCredentials {
+    /// @notice Withdrawal credentials type for legacy modules (0x01 - execution layer)
+    uint8 public constant WC_TYPE_01 = 0x01;
```
mkurayan, line 13: @hweawer might be a conflict with staking-router-3.0 brunch, (probably GItHub sync issue) but @krogla also added the WC_TYPE_01 & WC_TYPE_01 in this lib. Please merge latest code to your brunch.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -201,14 +263,30 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
...
+    uint256 public constant DATA_FORMAT_LIST_WITH_KEY_INDEX = 2;
 
     ILidoLocator internal immutable LOCATOR;
+    IStakingRouter internal immutable STAKING_ROUTER;
+
+    /// @notice Bitmask indicating which staking modules use legacy interface
+    /// @dev Each bit represents a module ID: bit N set to 1 means module N uses legacy interface
+    ///      Legacy modules (NOR, SDVT) use old getSigningKeys interface
+    ///      New modules (CSM, CuratedV2) use new getSigningKeys interface
+    uint256 public immutable LEGACY_MODULES_BITMASK;
```
krogla: see pr #1669  I removed the requirement to use a bitmask there

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -676,10 +799,214 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
...
+    * @param data Packed exit requests data
+    * @param dataFormat Format of the data (1 or 2)
+    * @return totalBalanceEth Total balance of all validators being exited in ETH
+    */
+    function _calculateTotalExitBalanceEth(bytes calldata data, uint256 dataFormat) internal view returns (uint16 totalBalanceEth) {
+        uint256 requestsCount = data.length / _getPackedRequestLength(dataFormat);
+
+        // Cache to avoid repeated external calls for the same module
+        // Format: (moduleId << 1) | isLegacy
+        uint256 cachedModuleId = type(uint256).max; // Initialize to invalid value
```
krogla: moduleId has 1-based numbering, so moduleId=0 is also invalid

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -676,10 +799,214 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
...
+
+        for (uint256 i = 0; i < requestsCount; ++i) {
+            uint256 moduleId;
+            uint256 itemOffset;
+
+            if (dataFormat == DATA_FORMAT_LIST) {
+                // For format 1: extract moduleId from the first 24 bits
+                assembly {
+                    itemOffset := add(data.offset, mul(PACKED_REQUEST_LENGTH, i))
+                    let dataWithoutPubkey := shr(128, calldataload(itemOffset))
```
krogla: if we define constants for offsets as well as for PACKED_REQUEST_LENGTH/PACKED_REQUEST_LENGTH_V2, the code can be simplified by avoiding duplication

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -676,10 +799,214 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
...
+            if (moduleId == cachedModuleId) {
+                isLegacy = cachedIsLegacy;
+            } else {
+                isLegacy = _isLegacyModule(moduleId);
+                cachedModuleId = moduleId;
+                cachedIsLegacy = isLegacy;
+            }
+
+            // Add balance based on module's withdrawal credentials type
+            if (isLegacy) {
```
krogla: it is better get module maxEb by request it from SR, see PR #1669

contracts/0.8.9/oracle/ValidatorsExitBus.sol
krogla: why we need cast to uint16? where is the check to ensure that data is not truncated?
what if MEB would be 10kETH, and you needed to withdraw 7 keys?

test/0.8.9/contracts/StakingRouter__MockForValidatorsExitBus.sol
```diff
@@ -0,0 +1,93 @@
...
+
+    /// @notice Mock function to set staking module address
+    /// @param moduleId The module ID
+    /// @param moduleAddress The module address
+    function setStakingModuleAddress(uint256 moduleId, address moduleAddress) external {
+        _modules[moduleId].stakingModuleAddress = moduleAddress;
+    }
+
+    /// @notice Implementation of IStakingRouter.getStakingModule
+    function getStakingModule(uint256 _stakingModuleId) external view returns (StakingModule memory) {
```
krogla: getStakingModule  should revert when module is not registered

test/0.8.9/oracle/validator-exit-bus-oracle.balanceCalculation.test.ts
```diff
@@ -0,0 +1,301 @@
...
+
+      it("should default to MaxEB balance for unconfigured modules", async () => {
+        // Module 777 is not explicitly configured in the setup
+        const requests: ExitRequest[] = [{ moduleId: 777, nodeOpId: 1, valIndex: 10, valPubkey: PUBKEYS[0] }];
+        const data = encodeExitRequestsDataList(requests, DATA_FORMAT_LIST);
+
+        const totalBalance = await oracle.calculateTotalExitBalanceEth(data, DATA_FORMAT_LIST);
+
+        // Unconfigured modules have withdrawalCredentialsType = 0, which is treated as MaxEB
+        expect(totalBalance).to.equal(MAXEB_MODULE_MAX_BALANCE_ETH);
```
krogla: incorrect behavior, should be reverted because the module is not registered, and not consume limit

also check other tests for this behavior

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -22,10 +23,60 @@ interface ITriggerableWithdrawalsGateway {
...
+interface INewStakingModule {
+    function getSigningKeys(
+        uint256 nodeOperatorId,
+        uint256 startIndex,
+        uint256 keysCount
+    ) external view returns (bytes memory);
+}
+
+interface INodeOperatorsRegistry {
+    function getSigningKey(
```
krogla, line 40: see pr #1669  for a workaround to get rid of multiple interfaces

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -22,10 +23,60 @@ interface ITriggerableWithdrawalsGateway {
...
+        uint64 lastDepositAt;
+        uint256 lastDepositBlock;
+        uint256 exitedValidatorsCount;
+        uint16 priorityExitShareThreshold;
+        uint64 maxDepositsPerBlock;
+        uint64 minDepositBlockDistance;
+        uint8 withdrawalCredentialsType;
+    }
+
+    function getStakingModule(uint256 _stakingModuleId) external view returns (StakingModule memory);
```
krogla: method is deprecated

krogla (changes requested):
llm code quality is unnecessarily complicated in places )
pls:
- avoid using deprecated methods
- use forge as .sol formatter
- cleanup obvious llm comments
- learn core contracts for best practice & code style

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: Done

contracts/common/lib/WithdrawalCredentials.sol
hweawer: Ack, merged @krogla branch

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: Done

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
hweawer: Done

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: Thanks for this!

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: Thanks for this!

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: Changed calculations to unit256

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: Thanks for this!

test/0.8.9/contracts/StakingRouter__MockForValidatorsExitBus.sol
hweawer: Removed, since deprecated

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: Changed

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: put 0 as default

contracts/0.8.9/oracle/ValidatorsExitBus.sol
hweawer: simplified here

test/0.8.9/oracle/validator-exit-bus-oracle.balanceCalculation.test.ts
hweawer: fixed

mkurayan (commented):
@hweawer please fix Scratch deploy&

contracts/0.8.9/oracle/ValidatorsExitBus.sol
```diff
@@ -218,15 +275,16 @@ abstract contract ValidatorsExitBus is AccessControlEnumerable, PausableUntil, V
...
     modifier preservesEthBalance() {
         uint256 balanceBeforeCall = address(this).balance - msg.value;
         _;
         assert(address(this).balance == balanceBeforeCall);
     }
 
-    constructor(address lidoLocator) {
+    constructor(address lidoLocator, address stakingRouter) {
         LOCATOR = ILidoLocator(lidoLocator);
+        STAKING_ROUTER = IStakingRouter(stakingRouter);
```
mkurayan: Lets use LOCATOR, in the same way how we use it for other contracts addresses.

contracts/0.8.9/oracle/ValidatorsExitBusOracle.sol
```diff
@@ -226,27 +227,33 @@ contract ValidatorsExitBusOracle is BaseOracle, ValidatorsExitBus {
     }
 
     function _handleConsensusReportData(ReportData calldata data) internal {
-        if (data.dataFormat != DATA_FORMAT_LIST) {
+        if (data.dataFormat != DATA_FORMAT_LIST && data.dataFormat != DATA_FORMAT_LIST_WITH_KEY_INDEX) {
```
mkurayan: @hweawer VEBO should now support only the new extended `DATA_FORMAT_LIST_WITH_KEY_INDEX`. Otherwise, this work does not make sense if the oracle can still submit reports in the old format.

Please fix this.

contracts/0.8.9/oracle/ValidatorsExitBus.sol
mkurayan: lets keep in SR as it done right now. Thanks

mkurayan (approved):

PR #1670, opened 2026-02-24
Deposits flow review fixes

Amuhar:


contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -504,7 +504,6 @@ contract DepositSecurityModule {
 
         if (quorum == 0 || sortedGuardianSignatures.length < quorum) revert DepositNoQuorum();
         if (!STAKING_ROUTER.canDeposit(stakingModuleId)) revert DepositInactiveModule();
-        if (!LIDO.canDeposit()) revert LidoCanNotDeposit();
```
mkurayan, line 507: @Amuhar  error `LidoCanNotDeposit` not used anywhere after this check removal, should we delete it as well?

mkurayan (approved):

PR #1681, opened 2026-03-01
wip(srv3): AO sanity checks

eddort:


contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -128,44 +131,61 @@ struct LimitsList {
...
     uint16 maxItemsPerExtraDataTransaction;
     uint16 maxNodeOperatorsPerExtraDataItem;
     uint32 requestTimestampMargin;
     uint64 maxPositiveTokenRebase;
-    uint16 initialSlashingAmountPWei;
-    uint16 inactivityPenaltiesAmountPWei;
+    uint16 maxCLBalanceDecreaseBP;
     uint16 clBalanceOraclesErrorUpperBPLimit;
+    uint32 consolidationEthAmountPerDayLimit;
+    uint128 exitedValidatorEthAmountLimit;
```
mkurayan: If it is nominated in ETH, you probably do not need such a huge number as uint128. 

Maybe it is better to use uint16 or uint32?

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -128,44 +131,61 @@ struct LimitsList {
...
+    /// @dev Must fit into uint32 (<= 4_294_967_295)
+    uint256 consolidationEthAmountPerDayLimit;
+    /// @notice Effective ETH amount attributed to a single exited validator
+    ///     in the exited ETH amount per day check.
+    /// @dev Stored in Wei. Must fit into uint128.
+    uint256 exitedValidatorEthAmountLimit;
 }
 
 /// @dev The packed version of the LimitsList struct to be effectively persisted in storage
 struct LimitsListPacked {
```
mkurayan: Consider splitting or otherwise optimizing this struct.

The original idea was to fit it into 256 bits so that it would occupy a single storage slot. If the data no longer fits into one storage slot, it may be worth splitting the limits.

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
eddort: fixed

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
krogla: There is no advantage in storing two separate configurations, since in any case it still occupies 2 storage slots.
And since in almost all cases all parameters are read/written at once, why complicate the code? The compiler will process 2 slots itself.

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -419,19 +462,46 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+    /// @notice One-time migration: seeds initial snapshots into reportData
+    ///     so that the sliding-window CL decrease check has a valid starting point.
+    function migrateBaselineSnapshot() external onlyRole(MIGRATION_MANAGER_ROLE) {
+        if (reportData.length != 0) revert MigrationAlreadyDone();
+
+        address lidoAddr = LIDO_LOCATOR.lido();
+        uint256 lidoVersion = IVersioned(lidoAddr).getContractVersion();
+        if (lidoVersion != 4) revert UnexpectedLidoVersion(lidoVersion, 4);
+
+        (uint256 migrationCLActiveBalance, uint256 migrationCLPendingBalance, uint256 migrationDeposits) = ILido(lidoAddr)
```
krogla: migrationCLActiveBalance -> migrationCLValidatorsBalance

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -128,44 +131,61 @@ struct LimitsList {
...
     uint16 clBalanceOraclesErrorUpperBPLimit;
+    uint32 consolidationEthAmountPerDayLimit;
+    uint16 exitedValidatorEthAmountLimit;
 }
 
 struct ReportData {
-    uint64 timestamp;
-    uint64 totalExitedValidators;
-    uint128 negativeCLRebaseWei;
+    uint256 timestamp;      // Logical report timestamp in seconds
```
krogla: Previously, elements in the array occupied 1 storage slot, now - 4.
Time -> uint64, balances can be stored in uint128, which is more than enough, so total 2 slots per element


```suggestion
struct ReportData {
    uint64 timestamp;      // Logical report timestamp in seconds
    uint128 clBalance;      // CL balance in Wei
    uint128 deposits;       // Deposits for the period since the last report in Wei
    uint128 clWithdrawals;  // Actual ETH moved from CL to withdrawal vault this period
}
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -660,106 +916,132 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
-                break;
-            }
+    function _checkCLBalanceDecrease(CLBalanceDecreaseCheckParams memory _checkParams) internal {
+        if (_checkParams.withdrawalVaultBalance < _lastVaultBalanceAfterTransfer) {
+            revert IncorrectCLWithdrawalsVaultBalance(
+                _checkParams.withdrawalVaultBalance,
+                _lastVaultBalanceAfterTransfer
+            );
+        }
+        if (_checkParams.withdrawalsVaultTransfer > _checkParams.withdrawalVaultBalance) {
```
krogla, line 935: this check seems redundant, as `withdrawalsVaultTransfer` is calculated inside `smoothenTokenRebase` and is always `<= withdrawalVaultBalance`

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -660,106 +916,132 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+
+        // Update vault state after this report's transfer
+        _lastVaultBalanceAfterTransfer =
+            _checkParams.withdrawalVaultBalance -
+            _checkParams.withdrawalsVaultTransfer;
+
+        // If the CL balance didn't decrease accounting for withdrawals, skip the window check
+        if (_checkParams.preCLBalance <= _checkParams.postCLBalance) return;
+        if (_checkParams.preCLBalance - _checkParams.postCLBalance <= clWithdrawals) return;
+
```
krogla, line 957: replace with 1 if, as it more clear
```suggestion
        if (_checkParams.preCLBalance <= _checkParams.postCLBalance + clWithdrawals) return;
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -660,106 +916,132 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+        // maxAllowed   = adjustedBase * limitBP / 10_000
+        uint256 lastIndex = _reportCount - 1;
+        uint256 lastTimestamp = reportData[lastIndex].timestamp;
+        uint256 windowStart = lastTimestamp > CL_BALANCE_WINDOW ? lastTimestamp - CL_BALANCE_WINDOW : 0;
+        uint256 baselineIndex = _findWindowStartIndex(lastIndex, windowStart);
+
+        uint256 baselineBalance = reportData[baselineIndex].clBalance;
+        actualCLBalanceDiff = baselineBalance > _postCLBalance
+            ? baselineBalance - _postCLBalance
+            : _postCLBalance - baselineBalance;
```
krogla, line 1012: should we return `0` instead of `_postCLBalance - baselineBalance`? because this actually means that the current balance has increased relative to the balance at the beginning of the window and therefore the subsequent check `if (actualCLBalanceDiff > maxAllowedCLBalanceDiff) ...` is incorrect

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -660,106 +916,132 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
-    // TODO: MaxEB - validator count-based data lookup not relevant for balance-based accounting
-    function _exitedValidatorsAtTimestamp(uint256 _timestamp) internal view returns (uint256) {
-        for (int256 index = int256(reportData.length) - 1; index >= 0; index--) {
-            if (reportData[uint256(index)].timestamp <= SafeCast.toUint64(_timestamp)) {
-                return reportData[uint256(index)].totalExitedValidators;
+        // Compute actual CL withdrawals for this period:
+        // clWithdrawals = current vault balance - vault balance after last report's transfer
+        uint256 clWithdrawals = _checkParams.withdrawalVaultBalance - _lastVaultBalanceAfterTransfer;
+        uint256 reportTimestamp = _lastReportTimestamp + _checkParams.timeElapsed;
+        _addReportData(reportTimestamp, _checkParams.postCLBalance, _checkParams.deposits, clWithdrawals);
```
krogla, line 946: The values postCLBalance and deposits recorded in the checkpoint are processed inconsistently:
deposits refer to the previous frame and are already accounted for in preCLBalance = (preValidatorsBal+prePendingBal+deposited), while postCLBalance already includes deposits.
Therefore, in _calcWindowDiff, when calculating adjustedBase = baseClBalance + ∑ deposits, we take into account deposits from the beginning of the window twice.
It seems that if we just omit deposits from 1 element, it should be ok, correct me if I'm wrong?

krogla (changes requested):

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
eddort: done

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
eddort: done, replaced all *ActiveBalance references

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
eddort: let’s leave this as is. the current implementation handles underflow, and keeping it as two separate lines improves readability, if you're fine with that.

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
eddort: `baseline` is the CL balance snapshot at `X-36`.
  Then `_calcWindowDiff()` adds deposits and subtracts withdrawals only for reports after that baseline, i.e.
  from `baseline + 1` up to the current report.

  So we take a point in the past and account for the external flows that formed the point in the future.
  The baseline report’s own flows are not included again, therefore its snapshot is not double-counted.

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
eddort: This behavior is intentional and follows the formula requested by the analytics team
Per that formula, we use the actual difference between the baseline and the current CL balance

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
eddort: Discussed this with the team; we decided to keep the logic as is. I’ve added a comment in the code to clarify.

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
krogla: In fact, the 1st check is already included in the 2nd one.

just moving `_checkParams.postCLBalance` to the right side of the inequality in the 2nd check eliminates the `sub` operation and potential overflow error, i.e.:

```
if (_checkParams.preCLBalance <= _checkParams.postCLBalance) return;
if (_checkParams.preCLBalance <= _checkParams.postCLBalance + clWithdrawals) return;

```
and then it becomes obvious that if somehow `preCLBalance <= postCLBalance`, then the result of the 2nd check is identical to the 1st for any value of `clWithdrawals` (which cannot be negative)

and since the `if`s logic is the same, it is logical to leave only the 2nd one

but in any case its minor issue

krogla (approved):
at first glance, it seems like it should work

scripts/scratch/deploy-params-testnet.toml
```diff
@@ -104,19 +102,22 @@ pauseIntentValidityPeriodBlocks = 6646  # Pause intent validity period in blocks
...
+maxEffectiveBalanceWeightWCType02 = 2048               # maxEB equivalent weight for WC type 2
 maxItemsPerExtraDataTransaction = 8                    # Maximum items per extra data transaction
 maxNodeOperatorsPerExtraDataItem = 24                  # Maximum node operators per extra data item
 requestTimestampMargin = 128                           # Request timestamp margin
 maxPositiveTokenRebase = 5000000                       # Maximum positive token rebase
-initialSlashingAmountPWei = 1000                       # Initial slashing amount (pWei)
-inactivityPenaltiesAmountPWei = 101                    # Inactivity penalties amount (pWei)
+maxCLBalanceDecreaseBP = 360                           # Max CL balance decrease over sliding window (BP, 360 = 3.6%)
 clBalanceOraclesErrorUpperBPLimit = 50                 # CL balance oracles error upper limit (BP)
+consolidationEthAmountPerDayLimit = 57600              # Consolidation ETH amount per day limit
```
VasilyevAA, line 119: Should be 91800

mkurayan (approved):
@eddort great work, thanks!

PR #1682, opened 2026-03-02
Feat/staking router 3.0 rename balance

mkurayan:
Rename "active balance" to "validator balance" to prevent ambiguity with the Ethereum consensus specification, in which `activeBalance` refers to a distinct concept.

Ethereum spec definition:
> `get_total_active_balance(state)` returns the sum of effective balances of all validators active in the specified epoch.


krogla (approved):
lgtm

PR #1683, opened 2026-03-03
feat: fix console warnings

mkurayan:
Fix console warnings

❌ ABI mismatch between:
   Interface: contracts/0.8.9/oracle/AccountingOracle.sol:IStakingRouter
   Contract:  contracts/0.8.25/sr/StakingRouter.sol:StakingRouter
   Match type: Sub-interface

📋 Parameter name mismatches (1 functions):
   Interface: function reportStakingModuleOperatorBalances(uint256 _stakingModuleId, bytes _operatorIds, bytes _totalBalancesGwei) returns ()
   Contract:  function reportStakingModuleOperatorBalances(uint256 _stakingModuleId, bytes _nodeOperatorIds, bytes _totalBalancesGwei) returns ()
   
**Fix:** Update parameter name.   
   

❌ ABI mismatch between:
Interface: contracts/0.4.24/Lido.sol:IAccountingOracle
Contract: contracts/0.8.9/oracle/AccountingOracle.sol:AccountingOracle
Match type: Sub-interface

**Fix:** The mismatch is because AccountingOracle.getProcessingState() returns a ProcessingState struct (ABI: tuple), but the interface in Lido.sol (Solidity 0.4.24, which can't use struct returns) declares individual return values. These are wire-compatible but differ in ABI JSON representation. Since Lido version can't be changed right now, the fix is to add this pair to the skip list in the check-interfaces task.

### DIFF
--- a/contracts/0.8.9/oracle/AccountingOracle.sol
+++ b/contracts/0.8.9/oracle/AccountingOracle.sol
@@ -45,7 +45,7 @@ interface IStakingRouter {
 
     function reportStakingModuleOperatorBalances(
         uint256 _stakingModuleId,
-        bytes calldata _operatorIds,
+        bytes calldata _nodeOperatorIds,
         bytes calldata _totalBalancesGwei
     ) external;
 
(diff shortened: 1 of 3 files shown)


krogla (approved):
lgtm

PR #1713, opened 2026-03-13
Feat/staking router 3.0 wv refactoring

mkurayan:
extract system contract addresses to constructor


contracts/0.8.9/WithdrawalVault.sol
```diff
@@ -97,6 +106,14 @@ contract WithdrawalVault is Versioned, WithdrawalVaultEIP7685 {
         _updateContractVersion(3);
     }
```
krogla: remove?

PR #1719, opened 2026-03-16
Feat/srv3 lido prevent deposits

krogla:
- no deposits until report
- allocation explain


mkurayan (approved):
Please add unit tests for cowering new logic in `canDeposit`

PR #1727, opened 2026-03-18
fix: forbid partial withdrawals for unhealthy vaults

folkyatina:
## Problem

## Solution

Your proposed solution

### DIFF
--- a/contracts/0.8.25/vaults/VaultHub.sol
+++ b/contracts/0.8.25/vaults/VaultHub.sol
@@ -907,7 +907,7 @@ contract VaultHub is PausableUntilWithRoles {
             ///      forceful validator exits required for rebalancing the vault. Partial withdrawals only allowed if
             ///      the requested amount of withdrawals is enough to cover the uncovered obligations.
             uint256 obligationsShortfallAmount = _obligationsShortfallValue(_vault, connection, record);
-            if (obligationsShortfallAmount > 0 && minPartialAmountInGwei * 1e9 < obligationsShortfallAmount) {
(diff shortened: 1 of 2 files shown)


contracts/0.8.25/vaults/VaultHub.sol
```diff
@@ -907,7 +907,7 @@ contract VaultHub is PausableUntilWithRoles {
             ///      forceful validator exits required for rebalancing the vault. Partial withdrawals only allowed if
             ///      the requested amount of withdrawals is enough to cover the uncovered obligations.
             uint256 obligationsShortfallAmount = _obligationsShortfallValue(_vault, connection, record);
```
TheDZhon: comment above should be updated though

TheDZhon (changes requested):

TheDZhon (approved):

contracts/0.8.25/vaults/VaultHub.sol
tamtamchik, line 902: 🧠 We no longer require `minPartialAmountInGwei`, actually, to be `uint256`, but it'll trigger too many changes, I guess, so we can skip it for now

Was thinking on something like this
```
bool hasPartialWithdrawal = false;
for (uint256 i = 0; i < _amountsInGwei.length; i++) {
    if (_amountsInGwei[i] > 0) {
        hasPartialWithdrawal = true;
        break;
    }
}
```

tamtamchik (approved):
Just one highlight on `minPartialAmountInGwei` – not a blocker.

PR #1746, opened 2026-03-26
Introduce Redeems Reserve

rkolpakov:


test/integration/core/redeems-reserve-scenarios.integration.ts
```diff
@@ -0,0 +1,1129 @@
...
+    });
+  });
+
+  // ═══════════════════════════════════════════════════════════════════════
+  //  Smoothing tightening after refSlot redemption
+  // ═══════════════════════════════════════════════════════════════════════
+
+  context("Smoothing tightening after refSlot redemption", () => {
+    /**
+     * @notice Failure: redemption between refSlot and report reduces smoothing base.
```
Psirex: Maybe worth using "Edge Case" instead of "Failure". As scenario is not failure

test/integration/core/redeems-reserve-scenarios.integration.ts
```diff
@@ -0,0 +1,1129 @@
...
+      const elRewards = ether("100"); // large enough to hit smoothing limit
+      await updateBalance(elVaultAddr, elRewards);
+
+      const totalPooledBeforeA = await lido.getTotalPooledEther();
+      await doReport({ excludeVaultsBalances: false, reportElVault: true });
+      const totalPooledAfterA = await lido.getTotalPooledEther();
+      const distributedA = totalPooledAfterA - totalPooledBeforeA;
+
+      // Restore to same state
+      await Snapshot.restore(testSnapshot);
```
Psirex: Do not modify test's snapshot. Instead took the new one local for the test

test/integration/redeems-reserve/helpers.ts
```diff
@@ -0,0 +1,550 @@
...
+}
+
+/** Deposits ETH, sets reserve ratio, and runs a report to fill the reserve */
+export async function seedReserve(
+  ctx: ProtocolContext,
+  holder: HardhatEthersSigner,
+  reserveManager: HardhatEthersSigner,
+  opts: {
+    deposit: bigint;
+    ratioBP: bigint;
```
Psirex: Consider naming it `redeemsReserveTarget`, similar to `depositsReserveTarget`

test/integration/redeems-reserve/helpers.ts
```diff
@@ -0,0 +1,550 @@
...
+
+/** Computes expected reserve target from internal ether and ratio */
+export function expectedReserveTarget(internalEther: bigint, ratioBP: bigint): bigint {
+  return ratioBP * internalEther / 10_000n;
+}
+
+/** Validates reserve-related state derived from total pooled ether */
+export function assertReserveState(state: ProtocolState, ratioBP: bigint) {
+  expect(state.reserveTarget).to.equal(expectedReserveTarget(state.internalEther, ratioBP));
+  expect(state.reserve).to.equal(state.reserveTarget);
```
Psirex, line 295: Seems like it works only when the reserve wasn't used and buffer has enough eth

test/integration/redeems-reserve/buffer-allocation.integration.ts
```diff
@@ -0,0 +1,207 @@
...
+    });
+    await lido.connect(secondHolder).submit(ZeroAddress, { value: SECOND_DEPOSIT });
+    await doReport(ctx, { excludeVaultsBalances: true, skipWithdrawals: true });
+
+    const protocol0 = await captureState(lido);
+
+    await assertReserveAllocationInvariant(lido);
+    assertReserveState(protocol0, DEFAULT_RATIO_BP);
+
+    // --- Request withdrawals exceeding unreserved buffer ---
```
Psirex, line 125: Consider adding an assert which checks that withdrawable ether actually less than the total withdrawals amount

test/integration/redeems-reserve/buffer-allocation.integration.ts
```diff
@@ -0,0 +1,207 @@
...
+
+    const requestAmount = await lido.balanceOf(holder.address);
+    await requestWithdrawals({ ctx, from: holder, amounts: [requestAmount] });
+
+    const beforeReport = await captureBufferState(ctx);
+    const restorePoint = await Snapshot.take();
+    const sharedAllocationBefore = beforeReport.withdrawalsReserve + getUnreserved(beforeReport);
+
+    // --- Path A: report with growthShare = 0 ---
+    await doReport(ctx, { skipWithdrawals: true, excludeVaultsBalances: true });
```
Psirex, line 180: Maybe it would make sense to simulate rewards in this test, to verify that `withdrawalsReserve` grows while the redeem reserve stays unchanged after the report.

test/integration/redeems-reserve/helpers.ts
```diff
@@ -0,0 +1,550 @@
...
+}
+
+/** Funds the EL rewards vault with the given amount */
+export async function fundElRewards(ctx: ProtocolContext, amount: bigint) {
+  const elVaultAddr = await ctx.contracts.locator.elRewardsVault();
+  await updateBalance(elVaultAddr, amount);
+}
+
+/** Advances chain time past the oracle requestTimestampMargin */
+export async function advanceToReportableTime(ctx: ProtocolContext) {
```
Psirex: Nitpick: the naming is a bit misleading. It suggests that this helper advances time to the point where the oracle can report, while it actually advances time to the point where all withdrawal requests satisfy the creation-time sanity check.

test/integration/redeems-reserve/negative-rebase.integration.ts
```diff
@@ -0,0 +1,244 @@
...
+    await doReport(ctx, { clDiff: CL_LOSS });
+    const state2: ProtocolState = await captureState(lido);
+    await assertReserveAllocationInvariant(lido);
+
+    // --- Compare paths: reserve auto-shrinks to new target in both ---
+    assertReserveState(state2, RATIO_BP);
+
+    // Verify: reserve target matches internalEther-based calculation in both paths
+    expect(state2.reserveTarget).to.equal(expectedReserveTarget(state2.internalEther, RATIO_BP));
+    expect(state1.reserveTarget).to.equal(expectedReserveTarget(state1.internalEther, RATIO_BP));
```
Psirex, line 115: Should `state2.reserveTarget` be slightly lower than `state1.reserveTarget`, given that the ratio is now applied to a smaller base? If so, it may be worth adding an assertion to make that behavior explicit.

test/integration/redeems-reserve/negative-rebase.integration.ts
```diff
@@ -0,0 +1,244 @@
...
+    await Snapshot.restore(simSnapshot);
+
+    // --- Path B: small redeem, then report with same CL loss ---
+    const REDEEM_AMOUNT = await getRedeemAmount(lido, "small");
+    const redeemShares = await lido.getSharesByPooledEth(REDEEM_AMOUNT);
+    const redeemEther = await lido.getPooledEthByShares(redeemShares);
+
+    await redeemExact(lido, holder, fix, REDEEM_AMOUNT);
+
+    await doReport(ctx, { clDiff: CL_LOSS });
```
Psirex, line 106: Consider adding an assertion to verify that the share rate decreased, confirming that the report was indeed negative

test/integration/redeems-reserve/negative-rebase.integration.ts
```diff
@@ -0,0 +1,244 @@
...
+    const externalShares0 = await lido.getExternalShares();
+    const internalShares0 = state0.totalShares - externalShares0;
+    const quote0 = await captureRedeemQuote(lido, QUOTE_STETH_AMOUNT);
+    await assertReserveAllocationInvariant(lido);
+    assertReserveState(state0, RATIO_BP);
+
+    // --- Internalize bad debt, process report ---
+    const badDebtShares =
+      (await dashboard.liabilityShares()) - (await lido.getSharesByPooledEth(await dashboard.totalValue()));
+    const internalizedBadDebtShares = await vaultHub
```
Psirex, line 209: Consider adding check, `internalizedBadDebtShares` is not zero

test/integration/redeems-reserve/low-tvl.integration.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  afterEach(async () => {
+    await Snapshot.restore(testSnapshot);
+  });
+
+  after(async () => {
+    await Snapshot.restore(snapshot);
+  });
+
+  it("small TVL: redeem + WQ + rebase with two holders", async () => {
```
Psirex: Consider adding an additional edge case where withdrawal finalization gets effectively clogged by the redeem reserve.

Example scenario:
Users A, B, and C hold most of the protocol stake. User A withdraws their full position. User B redeems part of their position, consuming the entire redeem reserve, and withdraws the remainder via the Withdrawal Queue. User C then withdraws.

Result: User C cannot be finalized. Even though the reserve has decreased, it is still not enough to cover the full withdrawal request.

Resolution: submit an oracle report that allocates all buffered ether to finalization, which resolves the issue.

test/integration/redeems-reserve/redeem-between-ref-and-report.integration.ts
```diff
@@ -0,0 +1,363 @@
...
+    const redeemShares = await lido.getSharesByPooledEth(redeemAmount);
+    const redeemEther = await lido.getPooledEthByShares(redeemShares);
+    await redeemExact(lido, holder, fix, redeemAmount);
+
+    const headroomDrained = await getHeadroom(lido);
+
+    await doReport(ctx, { excludeVaultsBalances: false, reportElVault: true });
+
+    const state2: ProtocolState = await captureState(lido);
+    const deferredRewards = await ethers.provider.getBalance(elVaultAddr);
```
Psirex, line 164: Needed a check, that deferred rewards are not zero

test/integration/redeems-reserve/redeem-between-ref-and-report.integration.ts
```diff
@@ -0,0 +1,363 @@
...
+    await assertReserveAllocationInvariant(lido);
+
+    // --- Fund EL rewards exactly at headroom ---
+    const headroomFull = await getHeadroom(lido);
+    await fundElRewards(ctx, headroomFull);
+
+    // --- Path A: report without redeem (via snapshot) ---
+    const simSnapshot = await Snapshot.take();
+
+    await doReport(ctx, { excludeVaultsBalances: false, reportElVault: true });
```
Psirex, line 136: It seems the current implementation of `accounting.report()` always calls `simulateReport()` internally. That may violate the requirement that redeem happens after refSlot. Worth double-checking that, in this test, `simulatedShareRate` is actually calculated from the pre-redeem state.

test/integration/redeems-reserve/full-circle.integration.ts
```diff
@@ -0,0 +1,262 @@
...
+    const batchesState = await withdrawalQueue.calculateFinalizationBatches(simulatedShareRate, maxTimestamp, 1000n, {
+      remainingEthBudget: buffered,
+      finished: false,
+      batches: Array(36).fill(0n),
+      batchesLength: 0n,
+    });
+    const overrideBatches = [...batchesState.batches].filter((x) => x > 0n);
+
+    // --- Process report with override batches ---
+    await report(ctx, {
```
Psirex, line 239: Consider adding an assertion for the redeem reserve: before the report, it should match the corresponding share of the protocol, and after the report, it should be zero.

test/integration/redeems-reserve/helpers.ts
rkolpakov: Yes, in cases where reserve drained/not replenished it uses explicit check

PR #1755, opened 2026-03-31
fix(VaultHub): add safe cast for reported values

folkyatina:
In `_applyVaultReport()`, the oracle-reported `_reportSlashingReserve` value (uint256) is downcast
to uint128 without using SafeCast, despite the contract importing it. Solidity 0.8.25 does not
check explicit narrowing type conversions -- higher-order bits are silently truncated. If
`_reportSlashingReserve` exceeds `type(uint128).max`, the `minimalReserve `would be set to an
incorrect (lower) value, undermining slashing reserve enforcement.

```
_record.minimalReserve = uint128(Math256.max(CONNECT_DEPOSIT,
_reportSlashingReserve));
```
While `type(uint128).max` (~3.4 * 10^38 wei, ~3.4 * 10^20 ETH) far exceeds any realistic value, the
inconsistency with the contract's own SafeCast import suggests the check was intended but
overlooked. Similar unchecked downcasts exist for `_reportCumulativeLidoFees` and other fields in the Report struct.

**Recommendations:**

 Use `SafeCast.toUint128()` for the downcast to ensure a clean revert with a
descriptive error rather than silent truncation in edge cases. Apply the same treatment to other
unchecked downcasts in `_applyVaultReport()`.

### DIFF
--- a/contracts/0.8.25/vaults/VaultHub.sol
+++ b/contracts/0.8.25/vaults/VaultHub.sol
@@ -1107,8 +1107,8 @@ contract VaultHub is PausableUntilWithRoles {
         uint256 _reportMaxLiabilityShares,
         uint256 _reportSlashingReserve
     ) internal {
-        _record.cumulativeLidoFees = uint128(_reportCumulativeLidoFees);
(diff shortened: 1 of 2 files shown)


dry914 (approved):
LGTM

PR #1767, opened 2026-04-16
Introduce Redeems Buffer

rkolpakov:


contracts/0.4.24/Lido.sol
```diff
@@ -573,6 +594,7 @@ contract Lido is Versioned, StETHPermit, AragonApp {
      */
     struct BufferedEtherAllocation {
         uint256 total;
+        uint256 redeemsReserve;
```
Psirex, line 597: Consider adding it as last property of the struct

contracts/0.4.24/Lido.sol
```diff
@@ -1063,13 +1123,24 @@ contract Lido is Versioned, StETHPermit, AragonApp {
...
     ) external {
         _whenNotStopped();
 
         ILidoLocator locator = _getLidoLocator();
         _auth(_accounting(locator));
 
+        // --- 1. Buffer round-trip: withdraw unredeemed ETH, reconcile bufferedEther ---
+        address buffer = locator.redeemsBuffer();
+        if (buffer != address(0)) {
+            IRedeemsBuffer(buffer).withdrawUnredeemed(_redeemedEther);
```
Psirex, line 1137: Nitpick: the naming is a bit misleading here, since withdrawUnredeemed() is called with redeemedEther

contracts/0.8.25/RedeemsBuffer.sol
```diff
@@ -0,0 +1,263 @@
...
+
+/**
+ * @title RedeemsBuffer
+ * @author Lido
+ * @notice Holds reserve ETH for instant stETH-to-ETH redemptions
+ */
+contract RedeemsBuffer is PausableUntil, AccessControlEnumerableUpgradeable {
+    using SafeERC20 for IERC20;
+    using SafeCast for uint256;
+    bytes32 public constant PAUSE_ROLE = keccak256("RedeemsBuffer.PauseRole");
```
Psirex, line 43: I’d recommend keeping the previous role naming convention unless there is a specific need to change it. Having the constant name match the keccak256 input that produces the role hash is a convenient convention and worth preserving if possible

contracts/0.8.25/RefSlotStore.sol
```diff
@@ -0,0 +1,76 @@
...
+import {IRefSlotStore} from "contracts/common/interfaces/IRefSlotStore.sol";
+
+import {_getCurrentRefSlot, RefSlotCache} from "./vaults/lib/RefSlotCache.sol";
+
+/**
+ * @title RefSlotStore
+ * @author Lido
+ * @notice Generic key-value store that auto-snapshots uint104 values at oracle frame boundaries
+ */
+contract RefSlotStore is IRefSlotStore, AccessControlEnumerable {
```
Psirex, line 18: Consider making this contract expose only the beacon for the current oracle reporting phase. It may be better to move away from the refSlot terminology and use naming aligned with oracle frames instead — something like `AccountingFrameClock.getCurrentFrame()`. A shared global storage contract that can be used by different callers, where one caller may overwrite values written by another, feels a bit scary

contracts/0.8.9/Accounting.sol
```diff
@@ -188,23 +207,24 @@ contract Accounting {
             update.sharesToBurnForWithdrawals,
             update.totalSharesToBurn // shares to burn from Burner balance
         ) = _contracts.oracleReportSanityChecker.smoothenTokenRebase(
-            _pre.totalPooledEther - _pre.externalEther, // we need to change the base as shareRate is now calculated on
-            _pre.totalShares - _pre.externalShares,     // internal ether and shares, but inside it's still total
+            _pre.totalPooledEther - _pre.externalEther - _pre.redeemedEther, // we need to change the base as shareRate is now calculated on
+            _pre.totalShares - _pre.externalShares - update.redeemedShares,  // internal ether and shares, but inside it's still total
```
Psirex, line 211: Could this line underflow in the case of a negative report and deferred redeemed shares from the previous report?

contracts/0.4.24/Lido.sol
```diff
@@ -1103,17 +1220,44 @@ contract Lido is Versioned, StETHPermit, AragonApp {
...
+    function _growRedeemsReserve() internal {
+        uint256 target = _getRedeemsReserveTarget();
+        BufferedEtherAllocation memory allocation = _getBufferedEtherAllocation();
+
+        uint256 growthShareBP = REDEEMS_RESERVE_GROWTH_SHARE_POSITION.getStorageUint256();
+        uint256 availableEther = allocation.withdrawalsReserve.add(allocation.unreserved);
+        uint256 minGrowth = availableEther.mul(growthShareBP) / TOTAL_BASIS_POINTS;
+        uint256 growth = Math256.max(allocation.unreserved, minGrowth);
+
+        uint256 newRedeemsReserve = Math256.min(allocation.redeemsReserve.add(growth), target);
```
Psirex, line 1257: This can zero the Lido-side reserve while a post-refSlot redeem is still carried to the next report.

Example:
1. `REDEEMS_RESERVE_POSITION = 100`, and `RedeemsBuffer` is funded with 100 ETH.
2. The report `refSlot` passes.
3. After `refSlot`, a user redeems 10 ETH.
   - `RedeemsBuffer.getRedeemed()` is 10 ETH.
   - `RedeemsBuffer.getRedeemedForLastRefSlot()` for this report is still 0.
   - `RedeemsBuffer` physically holds 90 ETH.
4. Governance sets `redeemsReserveTargetRatio = 0`.
5. The current report executes.
   - `Accounting` consumes 0 redeemed ETH.
   - `RedeemsBuffer.reconcile(0, 0)` returns the 90 unredeemed ETH to Lido.
   - This line computes `newRedeemsReserve = 0`.
   - `fundReserve()` sends 0 ETH back to `RedeemsBuffer`.
6. After the report, `RedeemsBuffer.getRedeemed()` still carries 10 ETH for the next report, but the Lido-side reserve is already 0.
7. On the next report, that carried 10 ETH becomes the refSlot snapshot. `RedeemsBuffer.reconcile(10, ...)` then reverts when it tries to compute `unredeemed = _reserveBalance - _redeemedEther.value`, because `_reserveBalance` is 0 and `_redeemedEther.value` is 10.

The report should not clear the Lido-side reserve while there are carried post-refSlot redeems.

PR #1793, opened 2026-04-28
fix: integration tests that depends on burner shares

dry914:
Forked protocol state (non-empty Burner, non-zero EL/WV vault balances) leaks into oracle-report math and breaks assertions written for scratch mode:
  - `bad-debt.integration.ts` — subtract Burner-burnt shares in the total-shares assertion.                                                                                                           
  - `sanity-checker-bad-debt.integration.ts` — reportBurner: false in the small-CL-decrease case so positive rebase from Burner doesn't mask the expected share-rate drop.
  - `withdrawal-edge-cases.integration.ts` — reportBurner: false on all reports so clDiff actually drives share-rate movement and bunker-mode detection.                                              
  - `withdrawals-bad-debt.integration.ts` — in finalizeWithdrawals set reportBurner: false + excludeVaultsBalances: true; drop burnerShares from totalSharesToBurn; use batchShareRate from the WithdrawalsFinalized event for the prefinalize branch condition; bump maxPositiveTokenRebase 1000 → 5000 so the threshold binary search finds a non-zero value.                                   
  - `accounting.integration.ts` — same pattern (Burner/vault isolation in the affected reports).


test/integration/vaults/withdrawals-bad-debt.integration.ts
```diff
@@ -221,7 +231,7 @@ describe("Integration: Withdrawals finalization with bad debt internalization",
     await upDefaultTierShareLimit(ctx, increaseBy);
 
     // Make the sanity checker more sensitive to the activation of smoothen token rebase
```
tamtamchik: Comment seems outdated...

tamtamchik (dismissed):

tamtamchik (dismissed):
GG

folkyatina (dismissed):

tamtamchik (approved):

PR #1820, opened 2026-05-20
Top up max value and pause + statemind audit fixes

Amuhar:

## Context

TopUpGateway pause and maxTopUpPerBlockGwei value  in SR to limit sum of top in block

Audit fixes:
 - TopUpGateway updates top-up rate-limit state even when no deposit is performed
 - StakingRouter.topUp() does not emit an event after depositing ETH
 - Module active check is duplicated in router deposit path
 - Activation epoch check rejects validators active in the proven epoch
 - Do not update last update data in topUp method in TopUpGateway if got 0 limits amount
 - update canDeposit commit in DSM


contracts/0.8.25/sr/StakingRouter.sol
```diff
@@ -686,8 +693,10 @@ contract StakingRouter is ISRBase, AccessControlEnumerableUpgradeable {
         /// @dev This method is only supported for new modules (0x02 withdrawal credentials)
         SRUtils._requireWCType2(stateConfig.withdrawalCredentialsType);
 
+        uint256 maxTopUpPerBlockWei = uint256(SRStorage.getRouterState().maxTopUpPerBlockGwei) * 1 gwei;
+        uint256 depositableEther = Math.min(LIDO.getDepositableEther(), maxTopUpPerBlockWei);
```
mkurayan: I think it should behave similarly to initial deposits: compute allocation against the entire buffer first, and only then cap the resulting value by maxTopUpPerBlock.

contracts/0.8.25/TopUpGateway.sol
mkurayan: Minor suggestion: rename `_minBlockDistance` to `_minTopUpBlockDistance` so that the naming is consistent with the other parameter names, such as `_minTopUpGwei`, `_maxValidatorsPerTopUp`.

PR #1821, opened 2026-05-21
srv3 update 1

krogla:
Cumulative update for SRv3 release
* fixes for audit findings
* optimiztions and refactor
* some new freatures


contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -502,15 +466,14 @@ contract DepositSecurityModule {
         if (nonce != onchainNonce) revert ModuleNonceChanged();
 
         if (quorum == 0 || sortedGuardianSignatures.length < quorum) revert DepositNoQuorum();
```
tamtamchik: I'm a bit confused, why not check `quorum != 0` on setter?

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -561,7 +532,7 @@ contract DepositSecurityModule {
      *
      * The signature, if present, must be produced for the keccak256 hash of the following message:
      *
-     * | UNVET_MESSAGE_PREFIX | blockNumber | blockHash | stakingModuleId | nonce | nodeOperatorIds | vettedSigningKeysCounts |
+     * | UNVET_MESSAGE_PREFIX | contractVersion | blockNumber | blockHash | stakingModuleId | nonce | nodeOperatorIds | vettedSigningKeysCounts |
```
tamtamchik: contractVersion or VERSION? Looks like some kind of drift, this happens in many places across the contract natspec

contracts/0.8.9/DepositSecurityModule.sol
tamtamchik: DEPOSIT_CONTRACT is already IDepositContract why cast?

contracts/0.8.9/DepositSecurityModule.sol
tamtamchik: Also double SLOAD? or we can pay 100gas? )

contracts/0.8.9/DepositSecurityModule.sol
tamtamchik, line 1: Bump year? it's 2023

tamtamchik: 8 and 16 are magiсal, maybe (if we are changing this file already)

 ```solidity
 /// @dev Byte length of one packed node operator id (uint64) in `nodeOperatorIds`.
///      Must match the packing expected by StakingRouter.decreaseStakingModuleVettedKeysCountByNodeOperator.
uint256 internal constant NODE_OPERATOR_ID_LENGTH = 8;

/// @dev Byte length of one packed vetted signing keys count (uint128) in `vettedSigningKeysCounts`.
///      Must match the packing expected by StakingRouter.decreaseStakingModuleVettedKeysCountByNodeOperator.
uint256 internal constant VETTED_KEYS_COUNT_LENGTH = 16;
```

contracts/0.8.9/DepositSecurityModule.sol
Amuhar: As I understand, all guardians can be removed and than quorum = 0, it was designed that way

tamtamchik (approved):
GG

contracts/0.8.9/DepositSecurityModule.sol
mkurayan: fixed

contracts/0.8.9/DepositSecurityModule.sol
mkurayan: Reverted, thanks

contracts/0.8.9/DepositSecurityModule.sol
mkurayan: fixed

PR #1824, opened 2026-05-22
Delete canTopUp & delete unnecessary check of depositable keys

Amuhar:

- Delete canTopUp method: off-chain will make separate checks instead of use canTopUp method
- Removed unnecessary depositableValidatorsCount cap as it is already checked in SR._getModuleDepositAllocation


krogla (approved):
lgtm

PR #1837, opened 2026-05-26
srv3: add dsm version to sig

eddort:


contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -378,7 +382,7 @@ contract DepositSecurityModule {
         int256 guardianIndex = _getGuardianIndex(msg.sender);
 
         if (guardianIndex == -1) {
-            bytes32 msgHash = keccak256(abi.encodePacked(PAUSE_MESSAGE_PREFIX, blockNumber));
+            bytes32 msgHash = keccak256(abi.encodePacked(PAUSE_MESSAGE_PREFIX, sig.contractVersion, blockNumber));
```
krogla: why not just 
`bytes32 msgHash = keccak256(abi.encodePacked(PAUSE_MESSAGE_PREFIX, VERSION, blockNumber));` ?

In that case, the first `if` statement with an explicit check isn't needed

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -524,7 +538,15 @@ contract DepositSecurityModule {
         Signature[] memory sigs
     ) internal view {
         bytes32 msgHash = keccak256(
-            abi.encodePacked(ATTEST_MESSAGE_PREFIX, blockNumber, blockHash, depositRoot, stakingModuleId, nonce)
+            abi.encodePacked(
+                ATTEST_MESSAGE_PREFIX,
+                sigs[0].contractVersion,
```
krogla: same approach, much easier 

`  bytes32 msgHash = keccak256(
            abi.encodePacked(
                ATTEST_MESSAGE_PREFIX,
                **VERSION**,` ?

and there is no need to perform an unnecessary and costly check (lines 500-510)

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -596,6 +620,7 @@ contract DepositSecurityModule {
                 // values with a dynamic type checked earlier
                 abi.encodePacked(
                     UNVET_MESSAGE_PREFIX,
+                    sig.contractVersion,
```
krogla: same

krogla (changes requested):
I think it can be simplified

krogla (approved):
lgtm

PR #1843, opened 2026-06-02
feat: add token rate pusher with args

dry914:
Add `ITokenRatePusherWithArgs` interface and it's support in `TokenRateNotifier` contract.

see stonks PR - https://github.com/lidofinance/stonks/pull/29


contracts/0.8.9/TokenRateNotifier.sol
```diff
@@ -42,97 +61,134 @@ contract TokenRateNotifier is Ownable, IPostTokenRebaseReceiver {
...
 
-    /// @notice `observer_` index in `observers` array.
+    /// @dev Handles a failed observer notification: bubbles up empty-data reverts (assumed
+    ///      "out of gas") to keep gas estimation correct, otherwise emits `PushTokenRateFailed`.
+    ///      This check is required to prevent incorrect gas estimation of the method. Without it,
+    ///      Ethereum nodes that use binary search for gas estimation may return an invalid value
+    ///      when the pushTokenRate() reverts because of the "out of gas" error. Here we assume that
+    ///      the pushTokenRate() method doesn't have reverts with empty error data except "out of gas".
+    function _handlePushFailure(address observer_, bytes memory lowLevelRevertData_) internal {
+        if (lowLevelRevertData_.length == 0) revert ErrorTokenRateNotifierRevertedWithNoData();
```
dry914, line 173: **Problem.** In the current implementation, any observer that reverts with empty data brings down the entire Accounting.handleOracleReport transaction (a guard introduced for gas-estimation correctness: empty revert is treated as OOG and bubbled up). Empty revert data occurs not only on OOG, but also on:
  - missing target selector with no fallback;
  - explicit `revert()` / `assembly { revert(0, 0) }`.
  
  **Result**: a single broken or malicious observer blocks Lido's daily rebase until removed via a DAO vote (~3-day Aragon vote). The behavior is inherited from the original contract - same on mainnet today.
  
  **Option A** — leave as is
  No code changes; document the risk more loudly. Rely on the trusted-observer model and governance vigilance.
  - ✅ Zero diff; matches production.
  - ❌ Brick risk remains; only recovery path is a DAO vote.
  
  **Option B** — drop the empty-data check
  Treat all reverts (including empty-data) as soft failures, log `PushTokenRateFailed`, continue the loop.
  - ✅ Local fix; a bad observer can no longer brick the rebase; minimal diff.
  - ❌ Loses the OOG guard. If the caller ever starts using binary-search gas estimation, under-estimation could land on prod and txs would fail.
  
  **Option C** — try/catch inside Accounting._notifyRebaseObserver
  Wrap the notifier call itself on the Accounting side; whatever happens inside the notifier/observers, the oracle report does not revert.
  - ✅ Architecturally the right place for a liveness fix; protects the protocol as a whole, not just this channel.
  - ❌ Out of scope for the current PR; requires a separate Accounting audit.

failingtwice: The assumptions that the subscribers are trusted and are always guaranteed to have non-zero revert data is still true. So the only possible 0-data revert is Out Of Gas error. So I propose to leave as-is.

contracts/0.8.9/TokenRateNotifier.sol
```diff
@@ -42,97 +61,134 @@ contract TokenRateNotifier is Ownable, IPostTokenRebaseReceiver {
...
             }
         }
         return INDEX_NOT_FOUND;
     }
 
     event PushTokenRateFailed(address indexed observer, bytes lowLevelRevertData);
-    event ObserverAdded(address indexed observer);
-    event ObserverRemoved(address indexed observer);
+    event ObserverAdded(address indexed observer, ObserverKind kind);
+    event ObserverRemoved(address indexed observer, ObserverKind kind);
```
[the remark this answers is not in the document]
dry914: kind param is removed

contracts/0.8.9/TokenRateNotifier.sol
```diff
@@ -1,33 +1,52 @@
...
+    /// @notice An interface that each legacy observer should support.
     bytes4 public constant REQUIRED_INTERFACE = type(ITokenRatePusher).interfaceId;
 
-    /// @notice All observers.
-    address[] public observers;
+    /// @notice An interface that each args-bearing observer should support.
+    bytes4 public constant REQUIRED_INTERFACE_WITH_ARGS = type(ITokenRatePusherWithArgs).interfaceId;
+
+    /// @notice All observers, in insertion order. Mixed kinds; an address may appear at most once.
+    Observer[] public observers;
```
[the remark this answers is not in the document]
dry914: intended

contracts/0.8.9/interfaces/ITokenRatePusher.sol
```diff
@@ -5,6 +5,8 @@ pragma solidity 0.8.9;
 
 /// @author kovalgek
 /// @notice An interface for entity that pushes token rate.
+/// @dev Legacy, no-arg flavor. Observers that need the full per-rebase payload (timestamps,
```
tamtamchik: Not sure if this is legacy? It's just another flavor.

contracts/0.8.9/TokenRateNotifier.sol
```diff
@@ -1,33 +1,52 @@
...
+/// @dev Observers are kept in a single registry tagged with their kind:
+///      - Legacy observers implement `ITokenRatePusher` and are notified via `pushTokenRate()`;
+///      - WithArgs observers implement `ITokenRatePusherWithArgs` and are notified via
+///        `pushTokenRate(...)` with the full per-rebase payload forwarded from
+///        `handlePostTokenRebase` (mirrors `IPostTokenRebaseReceiver` 1:1).
 contract TokenRateNotifier is Ownable, IPostTokenRebaseReceiver {
     using ERC165Checker for address;
 
+    /// @notice Distinguishes the two notification flavors an observer can subscribe to.
+    enum ObserverKind { Legacy, WithArgs }
```
tamtamchik:
```suggestion
    enum ObserverKind { NoArgs, WithArgs }
```

contracts/0.8.9/TokenRateNotifier.sol
```diff
@@ -1,33 +1,52 @@
...
 import {ERC165Checker} from "@openzeppelin/contracts-v4.4/utils/introspection/ERC165Checker.sol";
 import {ITokenRatePusher} from "./interfaces/ITokenRatePusher.sol";
+import {ITokenRatePusherWithArgs} from "./interfaces/ITokenRatePusherWithArgs.sol";
 import {IPostTokenRebaseReceiver} from "./interfaces/IPostTokenRebaseReceiver.sol";
 
 
 /// @author kovalgek
 /// @notice Notifies all `observers` when rebase event occurs.
+/// @dev Observers are kept in a single registry tagged with their kind:
+///      - Legacy observers implement `ITokenRatePusher` and are notified via `pushTokenRate()`;
```
tamtamchik:
```suggestion
///      - NoArgs observers implement `ITokenRatePusher` and are notified via `pushTokenRate()`;
```

contracts/0.8.9/TokenRateNotifier.sol
```diff
@@ -1,33 +1,52 @@
...
-    /// @notice Maximum amount of observers to be supported.
+    /// @notice Maximum total amount of observers (across both kinds) to be supported.
     uint256 public constant MAX_OBSERVERS_COUNT = 32;
 
     /// @notice A value that indicates that value was not found.
     uint256 public constant INDEX_NOT_FOUND = type(uint256).max;
 
-    /// @notice An interface that each observer should support.
+    /// @notice An interface that each legacy observer should support.
     bytes4 public constant REQUIRED_INTERFACE = type(ITokenRatePusher).interfaceId;
```
tamtamchik: Why not LEGACY_REQUIRED_INTERFACE? :) Would suggest using `REQUIRED_INTERFACE_NO_ARGS`

contracts/0.8.9/interfaces/ITokenRatePusherWithArgs.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-FileCopyrightText: 2026 Lido <info@lido.fi>
+// SPDX-License-Identifier: GPL-3.0
+
+pragma solidity 0.8.9;
+
+/// @notice An interface for an entity that pushes token rate and consumes the full per-rebase
+///         payload forwarded from `Lido.handlePostTokenRebase` via the `TokenRateNotifier`.
```
tamtamchik:
```suggestion
///         payload forwarded from `Accounting.handleOracleReport` via the `TokenRateNotifier`.
```

contracts/0.8.9/TokenRateNotifier.sol
```diff
@@ -1,33 +1,52 @@
...
 
-    /// @notice All observers.
-    address[] public observers;
+    /// @notice An interface that each args-bearing observer should support.
+    bytes4 public constant REQUIRED_INTERFACE_WITH_ARGS = type(ITokenRatePusherWithArgs).interfaceId;
+
+    /// @notice All observers, in insertion order. Mixed kinds; an address may appear at most once.
```
tamtamchik: > in insertion order

Nope, as removal shifts indexes.

tamtamchik (commented):
Overall great commitment! ❤️ 

ℹ️ Would avoid Legacy wording, as this is just another type.

failingtwice (dismissed):

tamtamchik (dismissed):
LGTM!

contracts/0.8.9/interfaces/ITokenRatePusherWithArgs.sol
```diff
@@ -0,0 +1,36 @@
...
+///      the full rebase payload instead of the no-arg `pushTokenRate()`. Field semantics mirror
+///      `IPostTokenRebaseReceiver` 1:1.
+///
+///      Implementers MUST return `true` from `supportsInterface` for this interface's id and MUST
+///      NOT also return `true` for `ITokenRatePusher`'s id unless dual-flavor behavior is
+///      intentional — when both ids are claimed, `TokenRateNotifier` registers the observer as
+///      `WithArgs` (richer payload wins).
+interface ITokenRatePusherWithArgs {
+    /// @notice Pushes token rate and receives the rebase payload that triggered the notification.
+    /// @param _reportTimestamp    Timestamp of the oracle report that triggered the rebase.
```
TheDZhon:
```suggestion
    /// @param _reportTimestamp    Timestamp of the oracle report data snapshot moment.
```

gathering != triggering rebase, bc latter is delivery

contracts/0.8.9/interfaces/ITokenRatePusherWithArgs.sol
```diff
@@ -0,0 +1,36 @@
...
+    /// @param _timeElapsed        Time elapsed since the previous report, in seconds.
```
TheDZhon:
```suggestion
    /// @param _timeElapsed        Time elapsed since the previous applied report, in seconds.
```

contracts/0.8.9/interfaces/ITokenRatePusherWithArgs.sol
```diff
@@ -0,0 +1,36 @@
...
+    /// @param _sharesMintedAsFees Amount of stETH shares minted as Lido protocol fees during this
+    ///        rebase. PER-REBASE value, NOT cumulative; may be `0` for a non-profitable report.
+    function pushTokenRate(
+        uint256 _reportTimestamp,
+        uint256 _timeElapsed,
+        uint256 _preTotalShares,
+        uint256 _preTotalEther,
+        uint256 _postTotalShares,
+        uint256 _postTotalEther,
+        uint256 _sharesMintedAsFees
```
TheDZhon, line 32: hmm, this one is treasury + node operator fee both, thus here we assume that revenue is the whole protocol fee, and node operator fee is cost of revenue;

is it by-design and intended? @tamtamchik @avsetsin

contracts/0.8.9/TokenRateNotifier.sol
```diff
@@ -1,129 +1,184 @@
...
         TOKEN_RATE_PROVIDER = tokenRateProvider_;
+        _transferOwnership(initialOwner_);
     }
 
-    /// @notice Add a `observer_` to the back of array
+    /// @notice Register an observer. The notification flavor is auto-detected from the observer's
+    ///         `supportsInterface` declaration: if it claims `REQUIRED_INTERFACE_WITH_ARGS` it is
+    ///         registered as `WithArgs` (and notified with the full rebase payload); otherwise it
+    ///         must claim `REQUIRED_INTERFACE_NO_ARGS` and is registered as `NoArgs`.
+    /// @dev If the observer claims BOTH interfaces, `WithArgs` wins (richer payload). Observers
```
TheDZhon: idk, I would probably just specified a kind with arg provided for the fn

contracts/0.8.9/TokenRateNotifier.sol
```diff
@@ -1,129 +1,184 @@
...
             revert ErrorAddExistedObserver();
         }
 
-        observers.push(observer_);
+        observers.push(Observer({addr: observer_, kind: kind_}));
         emit ObserverAdded(observer_);
     }
 
-    /// @notice Remove an observer from the array
+    /// @notice Remove an observer (of any kind) by address.
```
TheDZhon:
```suggestion
    /// @notice Remove a single found-first observer (of any kind) by address.
```

contracts/0.8.9/TokenRateNotifier.sol
```diff
@@ -1,129 +1,184 @@
...
-                ///      have reverts with empty error data except "out of gas".
-                if (lowLevelRevertData.length == 0) revert ErrorTokenRateNotifierRevertedWithNoData();
-                emit PushTokenRateFailed(
-                    observers[obIndex],
-                    lowLevelRevertData
-                );
+
+        uint256 observersLength_ = observers.length;
+        for (uint256 obIndex = 0; obIndex < observersLength_; obIndex++) {
+            Observer storage entry = observers[obIndex];
```
TheDZhon:
```suggestion
            Observer memory entry = observers[obIndex];
```

should be cheaper to load the whole structure in a single read (addr + byte-long enum)

contracts/0.8.9/interfaces/ITokenRatePusherWithArgs.sol
```diff
@@ -0,0 +1,36 @@
...
+///         payload forwarded from `Accounting.handleOracleReport` via the `TokenRateNotifier`.
+/// @dev This is the args-bearing variant of `ITokenRatePusher`. Observers implementing this
+///      interface are auto-detected by `TokenRateNotifier.addObserver` via ERC165 and called with
+///      the full rebase payload instead of the no-arg `pushTokenRate()`. Field semantics mirror
+///      `IPostTokenRebaseReceiver` 1:1.
+///
+///      Implementers MUST return `true` from `supportsInterface` for this interface's id and MUST
+///      NOT also return `true` for `ITokenRatePusher`'s id unless dual-flavor behavior is
+///      intentional — when both ids are claimed, `TokenRateNotifier` registers the observer as
+///      `WithArgs` (richer payload wins).
```
TheDZhon: it's a bit confusing tbh, maybe better just provide a choice of flavor upon registration

contracts/0.8.9/TokenRateNotifier.sol
```diff
@@ -1,129 +1,184 @@
...
 
-        observers.push(observer_);
+        observers.push(Observer({addr: observer_, kind: kind_}));
         emit ObserverAdded(observer_);
     }
 
-    /// @notice Remove an observer from the array
+    /// @notice Remove an observer (of any kind) by address.
     /// @param observer_ observer address to remove
     function removeObserver(address observer_) external onlyOwner {
```
TheDZhon, line 98: maybe it would be better to remove observer by index, not by addr, because the addr is ambiguous now

TheDZhon (changes requested):
👏  A few suggestions and small questions

contracts/0.8.9/interfaces/ITokenRatePusherWithArgs.sol
TheDZhon: hmm, I see, we try to re-calc https://github.com/lidofinance/stonks/pull/28/changes#diff-1d3b2b892608d1411815c7f6ce884e288497c35876819b3b0492924c63798740R157

contracts/0.8.9/TokenRateNotifier.sol
dry914: Since `addObserver` dedups across both kinds, an address maps to at most one observer here, so `removeObserver(address)` is unambiguous. What ambiguity did you have in mind?

contracts/0.8.9/interfaces/ITokenRatePusherWithArgs.sol
dry914: The notifier forwards `Lido.handlePostTokenRebase` verbatim (1:1), so this is the whole protocol fee mint (treasury + node-operator).
The split is done [downstream](https://github.com/lidofinance/stonks/blob/feature/dao-898-nest-v2-develop-the-protocol/contracts/automated-buybacks/revenue/StakingRevenueSource.sol#L157-L170): `StakingRevenueSource.pushTokenRate` takes the full `sharesMintedAsFees` and derives the treasury part via `StakingRouter.getStakingFeeAggregateDistribution()` (`* treasuryFee / (modulesFee + treasuryFee)`).

contracts/0.8.9/TokenRateNotifier.sol
TheDZhon: basically that we remove only first occurrence and it's not clear what kind it would be unless you know upfront, and imagine if you need to remove second occurrence ⇒ the sequence would be remove first, remove second, add first back

TheDZhon (dismissed):
GG

contracts/0.8.9/TokenRateNotifier.sol
dry914: An address is registered at most once (`addObserver` rejects address duplicates), so the match is unique and unambiguous.

dry914: (Added corresponded comment for clarity)

tamtamchik (approved):

PR #1848, opened 2026-06-06
Duplicated indicies check moved in common cycle & removed extra withdrawable_epoch check

Amuhar:
Duplicates check of validator indicies moved in common cycle
Removed extra withdrawable_epoch check


krogla (approved):
lgtm

PR #1856, opened 2026-06-11
Removed can deposit from dsm

Amuhar:
## Problem

  DSM.canDeposit is no longer used by off-chain tooling, and it didn't fully reflect whether a deposit into a module is actually
  possible (it ignored getDepositableEther and the module's allocation) — so the method is removed. As a result
  StakingRouter.canDeposit is no longer needed either, and neither is the Lido dependency in DSM.

## Solution

  Remove unused canDeposit from DSM and Staking Router

  What was done:
  - Removed DepositSecurityModule.canDeposit(uint256)
  - Removed StakingRouter.canDeposit(uint256)
  - Removed the ILido interface, the LIDO immutable, and the _lido constructor argument from DepositSecurityModule
  - Updated deploy scripts (scratch & upgrade) to the new constructor signature
  - Removed Lido__MockForDepositSecurityModule mock


krogla (approved):
appreciate it when the code gets shorter

mkurayan (approved):

PR #1921, opened 2026-07-17
feat: execution delegation framework

eddort:
1. **Added EDF support to DSM v5.** Guardians are now contracts, and signatures are checked through ERC-1271. Pause and unvet also work through EDF `execute()`. The guardian address is now included in the signed message.

2. **Integrated the official EDF repository.** Added deployment through its official command using a pinned commit. Added checks for the artifact, network, and Factory code, and saved the result in deployment state.

3. **Prepared EDF parameters for Hoodi and mainnet.** Added manifests with guardians, oracle committees, quorums, and the depositor. Added parameter validation and configuration checks for reused contracts.

4. **Archived SRv3 parameters.** The original Hoodi and mainnet deployment parameters are saved unchanged in `scripts/archive/`. The active files now contain EDF parameters.

5. **Separated protocol deployment from voting.** Added a separate path to prepare DSM v5 and the new Locator implementation. Voting contract deployment and test upgrade execution use separate steps.

6. **Added a test EDF vote for devnets and local forks.** Implemented the Locator switch, DSM permission transfer, oracle member replacement, and depositor permission grant. Added state checks before and after execution. Fixed mainnet holder selection and voting-time handling in mock voting.

7. **Updated and extended DSM and EDF tests.** Changed guardian fixtures to use contracts. Added checks for ERC-1271, signatures bound to guardians, deposits with a quorum, pause/unvet through EDF, delegate rotation and revocation, and termination. Test signing helpers now use 65-byte ECDSA signatures for EOA delegates instead of the compact 64-byte format.

8. **Updated the integration runner.** Set `RUN_NETWORK=local` and connected to an external node through `RPC_URL`. Added support for selected test files. Changed the existing `test:integration:upgrade` command to run the EDF scenario.

9. **Fixed fixtures and checks on forks.** Reward-transfer counts now include all registered CSM and CMv2 modules. Vault fixtures update stale oracle reports. The consensus helper waits for transaction confirmations before reading the result.

10. **Fixed Foundry installation in CI.** Updated the installation action and pinned Foundry to `v1.7.1` in unit and integration workflows.

11. **Added automatic deployment fees.** `AUTO_FEE=true` gets current estimates from Hardhat/ethers before each deployment. Kept the manual mode. Added tests for both modes, updated estimates, and missing EIP-1559 fee data.


contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -27,12 +29,9 @@ interface IStakingRouter {
...
 contract DepositSecurityModule {
-    /**
-     * @dev Short ECDSA signature as defined in https://eips.ethereum.org/EIPS/eip-2098.
-     */
-    struct Signature {
-        bytes32 r;
-        bytes32 vs;
+    struct GuardianSignature {
+        address guardian;
+        bytes signature;
```
F4ever, line 34: As I understand it, we're switching from (r, vs) to (r, v, s), right?
So the signature will be 65 bytes here?

F4ever:
DSM.sol LGTM

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -455,15 +455,15 @@ contract DepositSecurityModule {
      * Signatures must be sorted in ascending order by address of the guardian. Each signature must
      * be produced for the keccak256 hash of the following message (each component taking 32 bytes):
```
dry914, line 456: Nitpick/info:

> each component taking 32 bytes

But guardian is 20 bytes.

dry914: (the same in :367)

test/0.8.9/depositSecurityModule.test.ts
```diff
@@ -665,24 +709,68 @@ describe("DepositSecurityModule.sol", () => {
...
+      const sig = new DSMPauseMessage(guardian1.address, blockNumber).sign(guardian1.privateKey);
+      await guardian1.contract.setSignatureResponseMode(3);
+
+      await expect(dsm.connect(stranger).pauseDeposits(blockNumber, sig)).to.be.revertedWithCustomError(
+        dsm,
+        "InvalidSignature",
       );
     });
 
+    it("Reverts if a signature bound to one guardian is submitted for another guardian", async () => {
```
dry914, line 771: Can we add the same check for `depositBufferedEther`? (one guardian's signature is submitted for another one)

test/0.8.9/depositSecurityModule.test.ts
dry914, line 1181: Name says `(0,x,x)` but it's `(x,x) `now - `guardian1` is gone, so we lost the mixed case. Bring it back or rename?

lib/dsm.ts
```diff
@@ -1,12 +1,14 @@
-import { solidityPackedKeccak256 } from "ethers";
+import { BigNumberish, concat, solidityPackedKeccak256, toBeHex } from "ethers";
 
 import { DepositSecurityModule } from "typechain-types";
 
-import { sign, toEip2098 } from "./ec";
+import { sign } from "./ec";
```
dry914, line 5: `toEip2098` (lib/ec.ts:23) has no callers left after this PR. Worth dropping - it's still exported from lib, and a 64-byte compact sig now fails inside the delegation contract (OZ v5 only accepts 65 bytes), which would be a confusing thing to debug.

lib/deploy.ts
```diff
@@ -84,13 +81,28 @@ async function getDeploySigner(deployer: string): Promise<Signer> {
   return deployerSigner;
 }
 
-function getDeployTxParams(): DeployTxParams {
-  if (GAS_PRIORITY_FEE !== null && GAS_MAX_FEE !== null) {
+async function getDeployTxParams(): Promise<DeployTxParams> {
+  const gasLimit = process.env.GAS_LIMIT || null;
+  if (toBool(process.env.AUTO_FEE)) {
+    const { maxPriorityFeePerGas, maxFeePerGas } = await ethers.provider.getFeeData();
```
dry914, line 87: With `AUTO_FEE=true` there's no upper bound - if the network is busy `getFeeData()` can come back with a very high `maxFeePerGas`. Could we keep `GAS_MAX_FEE` as a cap here instead of ignoring it?

scripts/run-test-integration.sh
```diff
@@ -5,6 +5,19 @@ set -o pipefail
...
+  echo "Error: NETWORK must be set for integration tests"
+  exit 1
+}
+
+if [[ -n ${RUN_NETWORK:-} && $RUN_NETWORK != "local" ]]; then
+  echo "Error: integration tests require an external node and only support RUN_NETWORK=local"
+  exit 1
+fi
+
+export RUN_NETWORK="local"
```
dry914, line 18: This means integration tests now always need a node running beforehand - `yarn test:integration` used to spin up the in-process fork itself. Can we add a line to the README?

scripts/run-test-integration.sh
```diff
@@ -15,4 +28,10 @@ export SKIP_LINT_SOLIDITY=true
 prepare_migration_env "test"
 prepare_trace_args
 
-yarn hardhat --network "$RUN_NETWORK" test test/integration/**/*.ts "${TRACE_ARGS[@]}"
+if (($#)); then
+  TEST_FILES=("$@")
+else
+  TEST_FILES=(test/integration/**/*.ts)
```
dry914, line 34: `**` doesn't recurse without globstar, so this is really `test/integration/*/*.ts` - one level only. 17 of 74 files never run, including all of `vaults/roles/` and `vaults/scenario/` plus the 4 at the top level. Not from this PR, but this is the line.

dry914 (dismissed):
Looks solid in general 💪, please check info comments.

lib/deploy.ts
eddort: Thanks for the comment! The idea was to offer automatic fee estimation using ethers’ default behavior. Other deployment tools already support this, but we didn’t. Adding an implicit cap would go against that intent. If manual fee control is needed, `AUTO_FEE` can be left disabled and `GAS_MAX_FEE` and `GAS_PRIORITY_FEE` configured explicitly.

Ziars (approved):

mkurayan (approved):

dry914 (approved):
LGTM

PR #1940, opened 2026-08-07
Gloas: verifiers

madlabman:
Add Gloas support:
- primitives for verifiers


contracts/0.8.25/ValidatorExitDelayVerifier.sol
```diff
@@ -195,43 +211,21 @@ contract ValidatorExitDelayVerifier {
...
+     * @dev Ensures that `exitEpoch` is equal to `FAR_FUTURE_EPOCH` at the target block.
+     * @param recentBlock The recent block header anchored through EIP-4788.
+     * @param targetBlock The target block header and its proof against recentBlock.header.stateRoot.
+     * @param validatorWitnesses Array of validator proofs against targetBlock.header.stateRoot.
      * @param exitRequests The concatenated VEBO exit requests, each 64 bytes in length.
      */
     function verifyValidatorExitDelay(
-        ProvableBeaconBlockHeader calldata beaconBlock,
+        ProvableBeaconBlockHeader calldata recentBlock,
+        BlockRootsHeaderWitness calldata targetBlock,
```
dry914: Why is the `block_roots` hop mandatory now? 

Looks like old direct paths already covered essentially the same range.

contracts/0.8.25/vaults/predeposit_guarantee/CLProofVerifier.sol
```diff
@@ -124,14 +124,13 @@ abstract contract CLProofVerifier {
...
-     * @param _pivotSlot slot of the fork that alters first validator GIndex
-     * @dev if no fork changes are known,  _gIFirstValidatorPrev = _gIFirstValidatorCurr and _pivotSlot = 0
+     * @param _gIFirstValidatorPreGloas packed(general index | depth in Merkle tree, see GIndex.sol) GIndex of first validator in a pre-Gloas CL state tree
+     * @param _gIValidators packed GIndex of the Gloas validators field
+     * @param _pivotSlot first slot of the Gloas fork
      */
-    constructor(GIndex _gIFirstValidatorPrev, GIndex _gIFirstValidatorCurr, uint64 _pivotSlot) {
-        GI_FIRST_VALIDATOR_PREV = _gIFirstValidatorPrev;
-        GI_FIRST_VALIDATOR_CURR = _gIFirstValidatorCurr;
+    constructor(GIndex _gIFirstValidatorPreGloas, GIndex _gIValidators, uint64 _pivotSlot) {
```
dry914, line 131: `_pivotSlot = 0` used to be the documented "no fork" config here. 

Now it silently routes every slot through the Gloas branch and makes `GI_FIRST_VALIDATOR_PRE_GLOAS` dead. Add `require(_pivotSlot != 0)`? Tests already had to move to `MAX_UINT64` for this reason.

test/0.8.25/validatorExitDelayVerifier.test.ts
```diff
@@ -35,10 +38,9 @@ describe("ValidatorExitDelayVerifier.sol", () => {
   });
 
   // Mainnet values
-  // Pectra hardfork slot
-  // https://github.com/ethereum/consensus-specs/blob/365320e778965631cbef11fd93328e82a746b1f6/specs/electra/fork.md#configuration
+  // Keep the fixed-depth validator fixture on the pre-Gloas path.
   const FIRST_SUPPORTED_SLOT = 11649024;
-  const PIVOT_SLOT = 11649024;
+  const PIVOT_SLOT = 30_000_000;
```
dry914, line 40: `PIVOT_SLOT = 30_000_000` keeps the whole suite on the pre-Gloas path - the Gloas branch
is only covered by gindex assertions, never by an actual proof. `clProofVerifyer.test.ts`
already does this properly via `prepareGloasCLState`; worth mirroring here.

test/0.8.25/validatorExitDelayVerifier.test.ts
```diff
@@ -50,25 +52,29 @@ describe("ValidatorExitDelayVerifier.sol", () => {
   const SLOTS_PER_HISTORICAL_ROOT = 8192;
 
   describe("ValidatorExitDelayVerifier Constructor", () => {
-    const GI_FIRST_VALIDATOR_PREV = "0x0000000000000000000000000000000000000000000000000096000000000028";
-    const GI_FIRST_VALIDATOR_CURR = "0x0000000000000000000000000000000000000000000000000096000000000028";
+    const GI_FIRST_VALIDATOR_PRE_GLOAS = "0x0000000000000000000000000000000000000000000000000096000000000028";
+    const GI_VALIDATORS = "0x0000000000000000000000000000000000000000000000000000000000016600";
     const GI_FIRST_HISTORICAL_SUMMARY_PREV = "0x000000000000000000000000000000000000000000000000000000b600000018";
     const GI_FIRST_HISTORICAL_SUMMARY_CURR = "0x000000000000000000000000000000000000000000000000000000b600000018";
```
dry914, line 55: Pre-Gloas value, while `deploy-params-testnet.toml:162` uses `0x...170c00000018`. 

PS also lines 258, 881

contracts/common/lib/GIndex.sol
```diff
@@ -94,16 +95,46 @@ function concat(GIndex lhs, GIndex rhs) pure returns (GIndex) {
...
+function progressiveListNodeGIndex(uint256 i) pure returns (GIndex gI) {
+    if (i > (type(uint256).max - 1) / 3) {
+        revert IndexOutOfRange();
+    }
+
+    // Progressive-list chunk sizes are powers of four. The geometric series
+    // identifies the chunk containing i without walking every preceding chunk.
+    uint256 k = fls(i * 3 + 1) >> 1;
+
+    unchecked {
```
dry914, line 121: `unchecked` is a no-op — `k ≤ 255 >> 1 = 127`, so `3 * k + 3` can't overflow.

contracts/0.8.25/ValidatorExitDelayVerifier.sol
dry914, line 386: Per EIP-7688 `HistoricalSummary` stays a plain Container, so `GI_FIRST_BLOCK_ROOT_IN_SUMMARY_PREV/CURR` are always equal (and are in every config).

Dead branch - keep for future or remove?

scripts/upgrade/steps/0100-deploy-base-contracts.ts
dry914, line 379: `ValidatorExitDelayVerifier` has no proxy, so carrying the old address keeps the old implementation.

dry914, line 383: PDG has proxy, but no new implementation is deployed

dry914 (commented):
Contract changes look right, but upgrade flow doesn't ship new `ValidatorExitDelayVerifier` and PDG implementation.

tamtamchik (changes requested):
Agree with @dry914 comments, need to extend tests a bit

contracts/common/lib/GIndex.sol
madlabman: That's the point of using unchecked here,since the 3 * k + 3 can't overflow.

contracts/0.8.25/vaults/predeposit_guarantee/CLProofVerifier.sol
madlabman: The semantic didn't change, the old version simply used CURR version of the gindices which was superseded by GLOAS set.

contracts/0.8.25/ValidatorExitDelayVerifier.sol
madlabman: Already removed, thank you.

contracts/0.8.25/ValidatorExitDelayVerifier.sol
madlabman: Was removed before, thanks.

contracts/0.8.25/vaults/predeposit_guarantee/CLProofVerifier.sol
dry914: Re-reading this after the constructor rewrite — the `require(_pivotSlot != 0)` I asked for no longer makes sense, so dropping that.

With the GIndices gone and only `_gloasSlot` left, both extremes are meaningful configurations now:

- `type(uint64).max` — Gloas fork slot not yet known (what `upgrade-params-mainnet.toml` and `-hoodi.toml` use);
- `0` — Gloas active from genesis (valid for devnets);
- anything in between — the actual fork slot.

So a non-zero check would break the devnet case. And an accidental `0` on mainnet is already ruled out by the config schema: `gloasSlot` is a required key, so a missing one is a parse error rather than a silent zero.

One small ask before closing this: the convention currently only lives in TOML comments. Could you spell it out in the natspec on `GLOAS_SLOT`, here and in `CLValidatorVerifier`, so the sentinel values are documented where the immutable is declared? Worth noting that `ValidatorExitDelayVerifier` at least constrains it through `firstSupportedSlot > gloasSlot`, while `CLProofVerifier` and `CLValidatorVerifier` accept any value.

contracts/upgrade/UpgradeVoteScript.sol
```diff
@@ -103,806 +43,147 @@
...
-        permissions = bytes.concat(bytes20(factory), bytes4(ISetMerkleGateTree.validateInputData.selector));
-        for (uint256 i = 0; i < gates.length; ++i) {
-            if (gates[i] == address(0)) revert InvalidMerkleGateAddress();
-            permissions = bytes.concat(permissions, bytes20(gates[i]), bytes4(IMerkleGate.setTreeParams.selector));
-        }
-    }
+    function _getVoteItems() internal view returns (VoteItem[] memory items) {
+        UpgradeConfig config = UpgradeConfig(CONFIG);
+        items = new VoteItem[](RAW_ACTIONS_COUNT);
+        uint256 i;
```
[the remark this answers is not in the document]
dry914: fixed

PR #1971, opened 2026-09-17
fix: require GENESIS_FORK_VERSION for PDG deploy on non-local networks

tamtamchik:
Scratch deploy defaulted `GENESIS_FORK_VERSION` to mainnet `0x00000000`, so PDG on devnets got a wrong `DEPOSIT_DOMAIN` and rejects devnet deposit signatures. Tests missed it because they read the domain from the contract.

- Scratch step 0000: default `0x00000000` only on `hardhat`/`localhost`/`local`, other networks require the env.
- `deploy-gloas-verifiers.ts`: no fallback to the old PDG constructor args, env required.

Existing devnets need a PDG redeploy with the correct version.


scripts/scratch/steps/0000-populate-deploy-artifact-from-env.ts
```diff
@@ -13,14 +13,17 @@ function getEnvVariable(name: string, defaultValue?: string): string {
 }
 
 export async function main() {
+  const isLocalNode = ["hardhat", "localhost", "local"].includes(network.name);
```
dry914, line 16: `network.name` is the Hardhat runtime, which `migration-env.sh` fills from `RUN_NETWORK`, not `NETWORK` (`scripts/utils/migration-env.sh:5-6`, `:90-96`). Pointing `RUN_NETWORK=local` at an external node is a supported pattern here (`migration-env.sh:45`), and `tests-integration-hoodi.yml:44-46` is wired that way.

`NETWORK=hoodi RUN_NETWORK=local yarn deploy:scratch` therefore matches the allowlist and silently takes `0x00000000`, while Hoodi is `0x10000910` (`upgrade-params-hoodi.toml:66`): the wrong `DEPOSIT_DOMAIN` this PR fixes. The Hoodi workflow runs `yarn test:integration`, not the scratch deploy, so CI stays
  green.

Keying on the target network closes it. The fallback covers `scratch:deploy:localhost`, which sets no `NETWORK`:
```
      const targetNetwork = process.env.NETWORK ?? network.name;
      const isLocalNode = ["hardhat", "localhost", "local"].includes(targetNetwork);
```

